PR #25, opened 2017-03-23
Add valac-0.32 to valac find_program() path

Flowdalic:


mar-v-in:
Done as part of 5862e25

PR #33, opened 2017-04-10
Adds first basic translation for Italian

svalo:
If you would like to include an Italian translation I made a first version, It isn't perfect but it's a starting point :smiley: 

I don't know if and how metadatas in the .po files should be filled so please tell me if something is wrong.


fiaxh:
Thank you for the Italian translation! There is a [weblate](https://hosted.weblate.org/projects/dino/translations/) set up for Dino now, where translations can be edited online. Your translation is already there. The plugin translations are not there yet, but they will be.

svalo:
Great! Thanks, I'll follow it on weblate :+1:

svalo:
I'll close this as the translation is done on the Weblate

Issue #38, opened 2017-04-11
Not all strings can be translated


rscmbbng:
Not all of the strings are translatable. For example in src/ui/conversation_selector/conversation_row.vala:181 and 183. I tried to add them myself but that didn't work. Hence the translations of #33 and #37 are incomplete. 

Resolving #13 would greatly help the translation workflow

fiaxh:
Said strings were made translatable in e3b6953.

PR #39, opened 2017-04-12
Allows more strings to be translated

svalo:
Thanks to @crscmbbng for noticing, shoul fix dino issue #38 

The issue was a missing _("string") in source file, as seen in other files
It also moves the space outside the translatable string as it seems more coherent not having leading spaces there.


fiaxh:
Done as part of e3b6953.

PR #42, opened 2017-04-15
Update smiley_converter.vala

beli3ver:
Added some smilies, more will follow
Please add.

### DIFF
--- a/main/src/ui/chat_input/smiley_converter.vala
+++ b/main/src/ui/chat_input/smiley_converter.vala
@@ -13,8 +13,10 @@ class SmileyConverter {
     private static HashMap<string, string> smiley_translations = new HashMap<string, string>();
 
     static construct {
-        smiley_translations[":)"] = "🙂";
+        smiley_translations[":)"] = "😄";
+        smiley_translations[":-)"] = "😄";
         smiley_translations[":D"] = "😀";
+        smiley_translations[":-D"] = "😀";
         smiley_translations[";)"] = "😉";
         smiley_translations["O:)"] = "😇";
         smiley_translations["O:-)"] = "😇";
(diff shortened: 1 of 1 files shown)


main/src/ui/chat_input/smiley_converter.vala
```diff
@@ -25,6 +27,13 @@ class SmileyConverter {
         smiley_translations[":("] = "🙁";
         smiley_translations[":'("] = "😢";
```
NicoHood: How about ;( and :-( for crying?

main/src/ui/chat_input/smiley_converter.vala
```diff
@@ -25,6 +27,13 @@ class SmileyConverter {
...
         smiley_translations[":/"] = "😕";
+        smiley_translations["<3"] = "🖤";
+        smiley_translations[":heart:"] = "🖤";
+        smiley_translations[":kiss:"] = "💋";
+        smiley_translations[":-*"] = "💋";
+        smiley_translations[":*"] = "💋";
+        smiley_translations[":thumbsup:"] = "👍";
+        smiley_translations["+1"] = "👍";
```
NicoHood: I would not do this, as it might corrupt calculations or even links. Try (y) and (n) for example.

main/src/ui/chat_input/smiley_converter.vala
d9h02f: There is no tear in :-(

NicoHood: Oh I meant ;-(

fiaxh:
- `😄` is the unicode character for "Smiling Face With Open Mouth and Smiling Eyes", however :) has no open mouth
- :*) isn't a kiss, it's a kiss smiley. There is a unicode character `😗` "Kissing Face".
- +1 and `👍` may both be a signal of approval, but +1 isn't a depiction of `👍`
- [Wikipedia](https://en.wikipedia.org/wiki/List_of_emoticons) does not approve of the use of ;( for crying
- Typing a smiley and noticing that it gets replaced by an emoji is an intuitive thing. How would people find out about  `:kiss:`, `:thumbsup:` and `:heart:`? One would need a suggestion list (like github) popping up for that.

Also consider smileys like :), :] and (=. Would you say that they should have the same unicode counterpart? I wouldn't, but it's very subjective.

For now the list will stay as it is, containing only the smileys with the most obvious unicode counterpart. If everyone just adds the replacements they would like to have, in the end this feature will become unsuitable for everyone.

However on the long run there should be a more general solution than this replacement code.

PR #52, opened 2017-04-16
Add a French translation.

linkmauve:
I didn’t translate the date formats, imo they shouldn’t be present that way, the locale already knows about date formats.


d9h02f:
FYI: https://hosted.weblate.org/projects/dino/translations/

linkmauve:
Thanks for merging it, should I close this PR now and wait for the next synchronisation from weblate to this repository?

fiaxh:
Yes, I will merge it from weblate.

As for the time/date formats: strftime does not provide arbitrary date/time information in LC_TIME's format. For example, `%X` is the locale-dependent time representation including seconds, however you can't get a locale-dependent time without seconds. That's why some date/time strings are translated this way. To name two examples, gtk and glib also translate some date/time strings this way.

PR #56, opened 2017-04-16
Update README.md

beli3ver:
Add Install Instruction for Debian/Ubuntu


fiaxh:
The relevant information is already given in the build- and run-time dependencies and the build instructions. A separate copy-paste section for a specific distro that also includes how to set up a basic build environment and pull sources from git should not be necessary.

PR #110, opened 2017-06-20
request history since last known messages

max-weller:
when rejoining a room at startup

### DIFF
--- a/libdino/src/service/muc_manager.vala
+++ b/libdino/src/service/muc_manager.vala
@@ -28,13 +28,21 @@ public class MucManager : StreamInteractionModule, Object {
         stream_interactor.get_module(MessageProcessor.IDENTITY).pre_message_received.connect(on_pre_message_received);
     }
 
-    public void join(Account account, Jid jid, string? nick, string? password, string? history_since) {
+    public void join(Account account, Jid jid, string? nick, string? password) {
         if (enter_errors.has_key(jid)) return;
 
         Core.XmppStream? stream = stream_interactor.get_stream(account);
         if (stream == null) return;
         string nick_ = nick ?? account.bare_jid.localpart ?? account.bare_jid.domainpart;
         set_autojoin(stream, jid, nick_, password);
+
+        string history_since = null;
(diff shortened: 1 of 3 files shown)


max-weller:
Hi,
this pull request implements the `<history since>` tag from [XEP-0045, 7.2.15](https://xmpp.org/extensions/xep-0045.html#enter-managehistory).
I need this feature so I get all messages I missed in a MUC room while I was offline.
Can you merge it?
Thanks

libdino/src/service/muc_manager.vala
```diff
@@ -28,14 +28,14 @@ public class MucManager : StreamInteractionModule, Object {
         stream_interactor.get_module(MessageProcessor.IDENTITY).pre_message_received.connect(on_pre_message_received);
     }
 
-    public void join(Account account, Jid jid, string? nick, string? password) {
+    public void join(Account account, Jid jid, string? nick, string? password, string? history_since) {
```
fiaxh: Could you please not make history_since an argument here, but instead determine the value in this function? The way it is now, only history for autojoined MUCs but not for manually joined MUCs would be properly handled.

libdino/src/service/muc_manager.vala
```diff
@@ -229,7 +229,10 @@ public class MucManager : StreamInteractionModule, Object {
             foreach (Xep.Bookmarks.Conference bookmark in conferences) {
                 Jid jid = new Jid(bookmark.jid);
                 if (bookmark.autojoin) {
-                    join(account, jid, bookmark.nick, bookmark.password);
+                    string history_since = null;
+                    Conversation? conversation = stream_interactor.get_module(ConversationManager.IDENTITY).get_conversation(jid, account);
+                    if (conversation != null) history_since = conversation.last_active.to_string();
```
fiaxh: MessageStorage.get_last_message(conversation).time would be a better suited source for that information.

max-weller:
Those are good ideas, I changed it accordingly.

max-weller:
@fiaxh,
what do you think about the pull request with the changes applied?

fiaxh:
Thanks!

Issue #114, opened 2017-07-01
Dino won't connect to any server


Dracooooo:
I compiled Dino from the AUR yesterday and it doesn't seem to connect to any server.  

Edit: It says: Failed to start TLS: TLS support is not available.

foozzi:
similar problem, from aur

foozzi:
i find this https://github.com/dino/dino/issues/57
well, your and my xmpp servers (my exploit.im) use self-singed or broke ca-cert. I think this not bug, but error must be seem in "error dialog".

fiaxh:
This error message isn't due to a self-signed certificate but probably due to you not having `glib-networking` installed. Does installing the package solve the issue for you?

Dracooooo:
Yes, installing glib-networking fixed it, thanks.

NicoHood:
@fiaxh Could you add that dependency to the readme then?

PR #120, opened 2017-08-05
Mention libsoupt dependency in README

jubalh:
Used for HTTP upload since ea174ab632ced082eb0f1c51cea1bc9dc5c7c89e.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,7 @@ Build
 * libgee-0.8 (≥ 0.10)
 * libgcrypt (For the OMEMO plugin)
 * libnotify
+* libsoup (For the HTTP files plugin)
 * SQLite3
 
 **Instructions**


NicoHood:
what about glib-networking?
https://github.com/dino/dino/issues/114#issuecomment-320436120

jubalh:
Didnt read the commit carefully. Maybe its only background implementation without UI yet.

fiaxh:
I added the dependency to the Makefile in 6bc950d. The commit also fixes http upload for more setups -
 there should be a file-icon in the headerbar now (if a server supports the upload).

NicoHood:
Okay it works for me. I created a new account on a server that supports http upload. However the data is not omemo encrypted.

PR #121, opened 2017-08-08
Add snap packaging support

mariogrip:
This adds supports for snap packaging using snapcraft

Usage:
`snapcraft`
that's it 

To install use:
`sudo snap install dino_0.1_amd64.snap --dangerous`
or 
`sudo snap install dino_0.1_i386.snap --dangerous` if you have an i386 cpu

Read more: https://snapcraft.io/
Also you can setup https://build.snapcraft.io/ that will automatically pull from github and build the packages for you.


fiaxh:
We would like to not have any package build system specific configurations in our repository. I know that there currently are config files for tito, but we will remove them soon.
Furthermore, I do not see us maintaining snap packages, but please feel free to do so yourself.

PR #135, opened 2017-08-23
Focus selected widgets in the accounts dialog

hrxi:
Previously, you had to issue two clicks in order to enter something into
the text fields.

This also replaces the `Entry` widgets with their label counterparts
once they lose focus. Since the `ComboBox` handling of focus seems to be
buggy (`focus_out_event` not firing), the OpenPGP key selection lacks
the latter feature.


hrxi:
Googling around, I also found a couple of other developers stumbling upon the `ComboBox` thing, but not receiving answers. This either means that it is indeed buggy, but it might as well mean that I'm doing something fundamentally wrong. :)

PR #136, opened 2017-08-23
Show a full timestamp in the relative time tooltip

### LINKED ISSUE #78: [FR] Mousover Time should give the full date
Minor Feature.
"7 mins ago" is in general a good term. However if you want the exact time without calculating, or even in seconds I personally expected that a mouseover would show a tooltip with the full date + time. Or at least only the time, but with seconds.

hrxi:
Fixes #78.

### DIFF
--- a/main/src/ui/conversation_summary/conversation_item_skeleton.vala
+++ b/main/src/ui/conversation_summary/conversation_item_skeleton.vala
@@ -53,7 +53,10 @@ public class ConversationItemSkeleton : Grid {
 
     public void update_time() {
         if (items.size > 0 && items[0].display_time != null) {
+            DateTime local = items[0].display_time.to_local();
             time_label.label = get_relative_time(items[0].display_time.to_local());
+            /* xgettext:no-c-format */ /* Full date + time for the relative time tooltip */
+            time_label.tooltip_text = local.format(_("%x %X"));
         }
     }
 


NicoHood:
Cool! Can you please post a screenshot about that?

hrxi:
I don't have a test account right now, so I'm only showing the timestamp and the tooltip and not the whole window:

timestamp

NicoHood:
Nice. I dont know if the black background is theme dependent. But the feature itself looks nice, exactly what I want :)

hrxi:
It's probably theme-dependent – if you look into the code, you'll see that I'm simply setting the `tooltip` property, i.e. I'm not modifying the style of the tooltip at all.

fiaxh:
This is problematic because messages are merged. Currently although the time of the first message is used for the headline, the time also (mostly) applies to merged messages, as the time-info does not attempt to be precise. However if done as in this PR, you would only accurately show the time of the first message. That would be confusing and only of limited usability.

NicoHood:
maybe the tooltip should show the latest message then? i think the main point is to get an absolute date/time, not a relative one to the current time (3mins ago)

hrxi:
I think the timestamps were made more precise since the PR was opened. See also @fiaxh's reasoning.

Issue #123, opened 2017-08-10
http upload fails


silberzwiebel:
Running under Fedora 26, latest version from copr (corresponding to the master branch here), I can't use the newly implemented http_upload. I see the new file icon and can select a file but when I click OK to send the file nothing happens. The command line gives this error:
```
** (dino:11653): CRITICAL **: __lambda5_: assertion 'url_down != NULL' failed
```

fiaxh:
Looking at the stanzas revealed that this issue was due to a bug in an unstable version of prosody.

silberzwiebel:
for the record, here is the corresponding prosody bug (if I got this right): https://prosody.im/issues/issue/968

schmittlauch:
@fiaxh Are you sure about the "unstable version"? I'm running prosody-0.9.12 but am also affected by this bug.

`** (dino:24816): CRITICAL **: __lambda5_: assertion 'url_down != NULL' failed`

mar-v-in:
@schmittlauch Using prosody's http file upload plugin puts your prosody in an "unstable" state as the plugin is still "alpha", it doesn't matter what prosody version you're using.

Ref: https://hg.prosody.im/prosody-modules/file/tip/mod_http_upload/README.markdown#l3

_Sent from my Eterna Nexus 5X using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

schmittlauch:
@mar-v-in So it's probably not a fault in prosody itself, but in mod_http_upload.

fiaxh:
Dino is now (3123723) compatible with that behavior of prosody, thus people shouldn't have problems anymore.

PR #161, opened 2017-09-20
HTTP Upload: Revert to a switch on namespace

linkmauve:
https://hg.prosody.im/prosody-modules/rev/e1edf643fbb1 is now merged, so there is no reason to keep this workaround.

This reverts commit 312372350e24d1ebd8afbb0029fac04f2b64eb83, which was a fix for #123.


fiaxh:
We should probably give people quite some time to update prosody before reverting this. The current workaround does not have any real downsides, so that's no problem. I will leave this open as a reminder to be merged in the future.

PR #162, opened 2017-09-20
Add mediated invitation support

linkmauve:
This adds an invite_received signal which can then be connected from the
UI to autocomplete a join dialog.

See https://xmpp.org/extensions/xep-0045.html#invite-mediated


fiaxh:
Merged in 130965f.

PR #169, opened 2017-10-03
Remove the website reference (it redirects to github)

Jibec:
The website redirect here, i suggest to remove the reference if this is on purpose and the website won't exit anymore.


jubalh:
I think its the other way around. There might will be a website in the future. Dino does not yet have a release. Its all in the process..

Jibec:
I understand, we can close this

PR #173, opened 2017-10-04
Use libsecret to store the password

### LINKED ISSUE #102: Save passwords in keyring
By default, all passwords should be saved in the available keyring (gnome-keyring).

linkmauve:
Also migrate the old plain-text password to the libsecret store, and replace it with an empty string to avoid keeping it in plain text.

Fixes #102.

### DIFF
--- a/cmake/FindSecret.cmake
+++ b/cmake/FindSecret.cmake
@@ -1,0 +1,13 @@
+include(PkgConfigWithFallback)
+find_pkg_config_with_fallback(Secret
+    PKG_CONFIG_NAME libsecret-1
+    LIB_NAMES libsecret-1
+    INCLUDE_NAMES libsecret/secret.h
+    INCLUDE_DIR_SUFFIXES libsecret-1 libsecret-1/include
+    DEPENDS GLib Gio
+)
+
+include(FindPackageHandleStandardArgs)
+find_package_handle_standard_args(Secret
+    REQUIRED_VARS Secret_LIBRARY
+    VERSION_VAR Secret_VERSION)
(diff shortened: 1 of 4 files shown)


linkmauve:
So there are a bunch of issues with this patch:
- If the user doesn’t have DBus support (the case on Windows apparently), libsecret won’t be able to communicate with a SecretService, rendering storage of the password ineffective, and won’t connect automatically without any error message being given to the user.
- If the user doesn’t have a SecretService installed, same behaviour as previously.
- If the user didn’t start their SecretService, libsecret will block Dino while DBus starts it automatically, waits for the keyring to be unlocked, and then give back the password.

It seems pretty bad to me to simply fallback to storing the password in plain text in any of the first two cases, some of which could be caused by an attacker trying exactly to get access to the plain text password. It also seems pretty annoying to ask the user for their password repeatedly at each connection if it couldn’t be retrieved securely by libsecret. If anyone has a better idea, let me know!

NicoHood:
I think the best solution is to make libsecret a required dependency. Does dino even provide windows builds? Is this the aim? Maybe you can abstract this layer and use a different mechanism on windows. This way it is always 100% secure (no fallback) on linux, and insecure on windows (which shouldnt matter for the 0815 windows user at all).

RalfJung:
> fallback to storing the password in plain text in any of the first two cases

Please don't. Passwords should not be stored in clear text. At the very least, the application should yell loudly before doing so.

RalfJung:
I merged this patch into the dino I use locally and I can confirm that it works just fine :)  Thanks!

linkmauve:
#254 took over this PR, closing.

PR #209, opened 2017-11-12
Add an override to disable TLS verification, fixes #57

### LINKED ISSUE #57: No connection to account after building from source (SSL fail?)
After following [those steps](https://malte-kiefer.de/post/Dino%20ein%20neuer%20XMPP%20Client/) (and using sudo all the time) Dino does not connect and permanently is caught in loop to try and fail connecting. On both servers: jabber.ccc.de and and my own [freedombox](https://github.com/freedombox) server. 

```
** (dino:1120): CRITICAL **: file /opt/dino/xmpp-vala/src/core/stanza_writer.vala: line 21: unexpected error: Unacceptable TLS certificate (g-tls-error-quark, 2)

** (dino:1120): CRITICAL **: file /opt/dino/xmpp-vala/src/core/stanza_writer.vala: line 10: uncaught error: Unacceptable TLS certificate (g-tls-error-quark, 2)

** (dino:1120): CRITICAL **: file /opt/dino/xmpp-vala/src/core/stanza_reader.vala: line 49: unexpected error: Unacceptable TLS certificate (g-tls-error-quark, 2)

** (dino:1120): CRITICAL **: file /opt/dino/xmpp-vala/src/core/stanza_reader.vala: line 49: unexpected error: Unacceptable TLS certificate (g-tls-error-quark, 2)
Stream Error: Content before root node
Gtk-Message: GtkDialog mapped without a transient parent. This is discouraged.

```

Looks like this is an SSL related thing.

samdroid-apps:
Sometimes you might want to connect to an XMPP server with a bad TLS setup.
This commit allows the user to disable all TLS validation options in an
insecure mode.

To enable this, the user must change their account's local alias to
"MAN IN THE MIDDLE ME" - to reflect the dodgy nature of the setting.

---

I'm not sure about the UI, but I suppose it is a starting point for this discussion.  This is not a good setting - it really shouldn't be used a lot; so I'm not sure if it deserves a better UX.

### DIFF
--- a/libdino/src/entity/account.vala
+++ b/libdino/src/entity/account.vala
@@ -13,6 +13,12 @@ public class Account : Object {
     public string display_name {
         owned get { return alias ?? bare_jid.to_string(); }
     }
+    public bool is_insecure {
+        get {
+            if (alias == null) return false;
+            return alias == "MAN IN THE MIDDLE ME";
+        }
+    }
     public string? alias { get; set; }
     public bool enabled { get; set; default = false; }
     public string? roster_version { get; set; }
--- a/libdino/src/service/connection_manager.vala
+++ b/libdino/src/service/connection_manager.vala
@@ -164,6 +164,7 @@ public class ConnectionManager {
         if (resource == null) resource = account.resourcepart;
 
         Core.XmppStream stream = new Core.XmppStream();
+        stream.is_insecure = account.is_insecure;
         foreach (Core.XmppStreamModule module in module_manager.get_modules(account, resource)) {
             stream.add_module(module);
         }
--- a/xmpp-vala/src/core/xmpp_stream.vala
+++ b/xmpp-vala/src/core/xmpp_stream.vala
@@ -16,6 +16,7 @@ public class XmppStream {
     public string remote_name;
     public XmppLog log = new XmppLog();
     public StanzaNode? features { get; private set; default = new StanzaNode.build("features", NS_URI); }
+    public bool is_insecure = false;
 
     private IOStream? stream;
     private StanzaReader? reader;
--- a/xmpp-vala/src/module/tls.vala
+++ b/xmpp-vala/src/module/tls.vala
@@ -27,6 +27,9 @@ namespace Xmpp.Tls {
                     var io_stream = stream.get_stream();
                     if (io_stream == null) return;
                     var conn = TlsClientConnection.new(io_stream, identity);
+                    if (stream.is_insecure) {
+                        conn.validation_flags = 0;
+                    }
                     // TODO: Add certificate error handling, that is, allow the
                     // program to handle certificate errors. The certificate
                     // *is checked* by TlsClientConnection, and connection is
--- a/xmpp-vala/src/module/xep/0368_srv_records_tls.vala
+++ b/xmpp-vala/src/module/xep/0368_srv_records_tls.vala
@@ -39,9 +39,12 @@ public class TlsConnectionProvider : ConnectionProvider {
         SocketClient client = new SocketClient();
         try {
             IOStream? io_stream = client.connect_to_host(srv_target.get_hostname(), srv_target.get_port());
-            io_stream = TlsClientConnection.new(io_stream, new NetworkAddress(srv_target.get_hostname(), srv_target.get_port()));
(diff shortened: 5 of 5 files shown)


samdroid-apps:
Just for context; Geary does it by showing a popup with the error, then allowing the user manually override for certificate.  I'm not 100% sure how it is done:  https://github.com/GNOME/geary/blob/2c980036bd6b55efcad84917a96f9c02760f7670/src/client/application/geary-controller.vala#L599

licaon-kter:
Can you at least update this PR?

samdroid-apps:
Sorry, I don't seem to get notifications from github when my diffs become conflicting.  I'll resolve the conflict asap.

ailin-nemui:
would be nice to support comparing the cert fp and manually trust, rather than disable **any** validation

licaon-kter:
Any news?

stevenroose:
I don't think this is a good UI. It's quite common to host a server with a custom TLS certificate. It's also not insecure if you ask the user to verify the fingerprint. There should be a screen to ask the user what to do.

I used to have a custom cert on my server and I could recognize the certificate fingerprint, which makes it perfectly secure.

mar-v-in:
@stevenrose: That's just plain wrong.
1. The certificate is not only used for client-to-server connections but also for server-to-server connections. While it might be feasible to ask the user to verify a fingerprint (it usually isn't), that's not possible for a remote server. By using self-signed or custom CA certificates on server-to-server connections, you are basically giving a server that wants to you the choice to either be susceptible to MITM attacks or not send the message.
2. You argue that you're able to recognize the fingerprint and I bet you aren't. You might be able to recognize a part of the fingerprint, but that means that you're lowering the security, it's not "perfectly secure". Also, which hash algorithm did you use for the fingerprint you remembered? MD5, SHA1? Both are known to be attackable (OK, SHA1 attacks might not be feasible at this point, but that doesn't matter).
3. With Let's encrypt there is no excuse to not use a proper certificate nowadays. 

We are aware that there are still some cases when invalid certificates occur and want to handle them properly:
1. The certificate's CommonName might not match the server domain. This can happen if the XMPP-Server is operated by a third-party and the SRV record points there. In such a case we'd like the user to confirm that he is using that third party.
2. The certificate is out-of-date. In cases where the certificate became out-of-date a short time ago, we'd like to inform the user about that and let him connect anyway, buy only if that certificate was used before while it was still valid.
3. The certificate is valid but signed by an untrusted CA. We'd prefer to ask the user to either import this CA into his system certificate store or verify it's fingerprint*.
4. The certificate is valid but self-signed. Verify the fingerprint with the user*.
*When I say verify fingerprint above, I don't mean, display the fingerprint and ask the user to press OK. This is a ver y error-prone design, because people that don't know what this means will easily bypass the check without verifying the fingerprint. Thus fingerprint verification will be more like a user interface that asks to enter the fingerprint.

That said, this is all out-of-scope right now, given our minimal and non-user-friendly account management and creation process.

This PR is still open to allow those that do not properly configure there server to test dino by manually merging the PR in their local repo before building (dino is still pre it's first release, so everybody is just testing anyway). It will not be merged.

stevenroose:
I do agree that with Let's Encrypt, it shouldn't happen anymore. Like I said, I was talking about the past, I'm using Let's Encrypt myself as well.

I also overlooked the S2S communication issue. That's a real problem, indeed. However, it might still be possible that the server is using a different cert for S2S; as a client, you shouldn't assume the C2S certificate is the same as the S2S one. Even though I don't think any modern server implementation let's you have different ones easily anyway.

As far as fingerprint recognition goes, that's the responsibility of the user. I do agree fully that just showing a certificate confirmation window on an unexpected certificate is wrong, since most users don't know anything about certificates and will just press OK. But if your server has a custom cert and your users are all informed of such, you should be able to check a box "allow custom cert" and then show the fingerprint confirmation box. Or simpler, an "allow custom cert" box that opens a fingerprint text box where the cert that should be accepted should be inputted.

stevenroose:
Anyway, totally right that this is not a feature that is required for now. Keep up the good work with making the experience for actual users better!

IamPARANOID:
I don't think anyone is going to use Let's Encrypt for a LAN-only XMPP setup.
Also, the patch does not work.

This makes the client essentially non-functional, which is sad considering it is the only GUI XMPP client that doesn't look like it was designed for Windows XP.  :(

PR #215, opened 2017-11-18
conference_s_, Members only room, Too many occupants in room

comradekingu:


fiaxh:
I would like to not change `Banned from joining or creating conference` to be plural as this must not necessarily apply to all conferences (on that server).

The others indeed sound better. Please feel free to also implement your other suggestion `No conversation active` -> `No active conversations`.

comradekingu:
@fixah I did that one and a few others, can't find anything else.
In a few places there is no full stop on the second sentence, but it looks to be deliberate.
"First sentence. Second sentence"
vs
"First sentence. Second sentence."  ←

d9h20f:
Should be 'Members-only room' if you really want to change it.

'Password required to enter the room' is correct ("(A) Password (is) required to enter room"), 'Required password to enter the room' isn't. If anything, 'Leave empty for none' is the part that could be improved.

I can't agree on 'Remove the account "%s"', the 'the' should not be there, and '"%s"' should use these quotation marks instead: ‘’

Otherwise the changes look good.

comradekingu:
@d9h20f Changed to "Members-only room"
In my opinion "remove account %s" means there is an unmarked noun. (Assuming here %s is the account name.) Quotation marks changed.
Will do, "if any" to do away with the "leave empty for none"

d9h20f:
I think I misunderstood the password sentence, and it's actually a description, so could possibly be rephrased as "Password (if any) to enter room". Commas between parentheses don't look good, and I don't think they're grammatically correct, either.

comradekingu:
@d9h20f How about "If any, …"

PR #219, opened 2017-11-24
Add "Copy URL" context menu option when hovering over a URL

pp3345:
Many IM clients (e. g. Pidgin) offer the convenience option to copy a URL without selecting it first when right-clicking on it. This PR adds that feature to Dino.

image

### DIFF
--- a/main/src/ui/conversation_summary/message_textview.vala
+++ b/main/src/ui/conversation_summary/message_textview.vala
@@ -19,6 +19,7 @@ public class MessageTextView : TextView {
         update_display_style();
         Util.force_base_background(this, "textview, text:not(:selected)");
         style_updated.connect(update_display_style);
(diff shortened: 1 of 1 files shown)


fiaxh:
Your code depended on "Copy" being translated in order to correctly insert the new item. On the other hand, most of the default items (Cut, Paste, Delete) won't work in anyways, as the textview is not editable. I rewrote the code to remove the default items and insert only the ones fitting ones.

pp3345:
I also thought that searching through the list of menu items and looking for an item named "Copy" (with its translation) is a bit ugly, however, I didn't find another solution to insert the "Copy URL" item directly after the "Copy" item. However, if there is no translated "Copy" item, the code would have put the "Copy URL" item at the end of the list, so that wouldn't have caused too many problems. Removing the unused default items instead is a better idea though.

Thanks for merging!

PR #220, opened 2017-11-25
Find libraries with pkgconfig

bochecha:
*Note: I've started working on flatpaking Dino, this is part of the issues I have found while working on this.*

This is necessary for Flatpak, because Yocto messes with the /usr/bin/*-config files, and really wants everyone to use pkgconfig.
    
A fallback is kept to build on platforms which don't have pkgconfig for those libraries.

### DIFF
--- a/cmake/FindGCrypt.cmake
+++ b/cmake/FindGCrypt.cmake
@@ -1,57 +1,8 @@
-set(GCrypt_PKG_CONFIG_NAME gcrypt)
-
-find_program(GCrypt_CONFIG_EXECUTABLE NAMES libgcrypt-config)
-mark_as_advanced(GCrypt_CONFIG_EXECUTABLE)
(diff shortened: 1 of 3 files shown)


bochecha:
I tested this on both Fedora 26, Debian Stretch and inside a Flatpak build sandbox. More testing on other platforms would of course be welcome. 🙂

bochecha:
> The standard for these two libraries is to call their NIH’d pkg-config script. I’d recommend using it as a fallback

This is now done in the latest push to this branch.

It builds successfully inside a Flatpak build sandbox (using `pkg-config`) and on Fedora 27 as well as Debian Stretch. (both using the fallback on the `*-config` scripts)

I don't have access to other OSes to test on them, though.

bochecha:
Hi there, I just rebased this on the latest master. Anything I can do to get it merged? 🙂

PR #229, opened 2017-12-02
Use a more portable interpreter for configure

SamWhited:
`/bin/bash` is a Linux-ism and isn't very portable (eg. installing bash on FreeBSD sticks it at `/usr/local/bin/bash`), however, the configure script runs on `/bin/sh` fine.

### DIFF
--- a/configure
+++ b/configure
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 
 OPTS=`getopt -o "h" --long \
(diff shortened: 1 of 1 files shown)


linkmauve:
There are a lot of bashisms in this file, I would recommend fixing them before changing the shebang.

SamWhited:
At least for a basic build it works fine under sh at least, but I didn't exactly spend time looking for bash or testing. Thanks anyways.

PR #231, opened 2017-12-03
Add an appdata file

bochecha:
This allows "app center" applications like GNOME Software and Plasma Discover to show Dino in all its glory, so users can install it simply.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,6 +32,7 @@ set_path(SHARE_INSTALL_PREFIX "${CMAKE_INSTALL_PREFIX}/share" "Installation dire
 
 set_path(BIN_INSTALL_DIR "${EXEC_INSTALL_PREFIX}/bin" "Installation directory for user executables")
 set_path(DATA_INSTALL_DIR "${SHARE_INSTALL_PREFIX}/dino" "Installation directory for dino-specific data")
+set_path(APPDATA_FILE_INSTALL_DIR "${SHARE_INSTALL_PREFIX}/metainfo" "Installation directory for .appdata.xml files")
 set_path(DESKTOP_FILE_INSTALL_DIR "${SHARE_INSTALL_PREFIX}/applications" "Installation directory for .desktop files")
 set_path(SERVICE_FILE_INSTALL_DIR "${SHARE_INSTALL_PREFIX}/dbus-1/services" "Installation directory for .service files")
 set_path(ICON_INSTALL_DIR "${SHARE_INSTALL_PREFIX}/icons" "Installation directory for icons")
--- a/main/CMakeLists.txt
+++ b/main/CMakeLists.txt
@@ -151,6 +151,7 @@ if(WIN32)
 endif(WIN32)
 
 install(TARGETS dino ${TARGET_INSTALL})
+install(FILES data/im.dino.Dino.appdata.xml DESTINATION ${APPDATA_FILE_INSTALL_DIR})
 install(FILES data/im.dino.Dino.desktop DESTINATION ${DESKTOP_FILE_INSTALL_DIR})
 install(FILES data/im.dino.Dino.service DESTINATION ${SERVICE_FILE_INSTALL_DIR})
 install(FILES data/icons/im.dino.Dino.svg DESTINATION ${ICON_INSTALL_DIR}/hicolor/scalable/apps)
--- a/main/data/im.dino.Dino.appdata.xml
+++ b/main/data/im.dino.Dino.appdata.xml
@@ -1,0 +1,60 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop">
+  <id>im.dino.Dino</id>
+  <launchable type="desktop-id">im.dino.Dino.desktop</launchable>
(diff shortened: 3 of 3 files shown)


bochecha:
So first of all, I kind of took the description out of a hat. I'm not sure this is even all true, and I totally expect to have to rewrite it. Any input would be welcome. 🙂 

---

The appdata file passes the minimum validation:

```
$ appstream-util validate-relax main/data/im.dino.Dino.appdata.xml 
main/data/im.dino.Dino.appdata.xml: OK
```

However there are a few issues left, which can be seen with the strict validator:

```
$ appstream-util validate-strict main/data/im.dino.Dino.appdata.xml 
main/data/im.dino.Dino.appdata.xml: FAILED:
• tag-missing           : <update_contact> is not present
• attribute-invalid     : <screenshot> width too large [https://camo.githubusercontent.com/a8afd3c62a5595fc6ad12b70bc8fde74fe6a178b/68747470733a2f2f692e696d6775722e636f6d2f314b714c7144562e706e67] maximum is 1600px
• attribute-invalid     : <screenshot> height too large [https://camo.githubusercontent.com/a8afd3c62a5595fc6ad12b70bc8fde74fe6a178b/68747470733a2f2f692e696d6775722e636f6d2f314b714c7144562e706e67] maximum is 900px
• translations-required  : <name> has no translations
• translations-required  : <summary> has no translations
• translations-required  : <description> has no translations
Validation of files failed
```

The `update_contact` bit should contain [an email that distributors can contact in case of problems with the appdata itself](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-update_contact). I didn't know what to put here, this is really for you (the Dino maintainers) to decide. It's optional though, so we could just leave it out.

The translations for `name`, `summary` and `description` are a huge plus for non-English-speaking users. The app center automatically displays the translated metadata depending on the user locale. However that requires hooking the translation with cmake, which I don't know how to do yet.

Finally, there are multiple problems with the [screenshots](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-screenshots):

* the one I used is too large; typical sizes are 1600x900 and 1200x675; a ratio of 16:9 is recommended; 
* it would be nice to include more screenshots, for example showing the list of users in a conference, or the "start chat" dialog;
* it is possible (and very desirable) to show translated screenshots;

Those are all optional parts of the appstream spec though, so they can be improved upon later on.

d9h20f:
I'd get rid of all references to 'Jabber' (after 10-15 years of no longer being the actual name of the protocol, it's time to let it die), but otherwise the description looks good to me.

bochecha:
> I'd get rid of all references to 'Jabber'

Done in the latest push to this branch.

> otherwise the description looks good to me

There's one thing I feel I should mention: I wrote that "Dino integrates best with the GNOME desktop", which I compltely invented. I have no idea whether that's an actual goal or an accident, but given it uses Gtk3 and how it seems to follow the GNOME HIG (it even looks like the GNOME Chat designs!), I felt like that was the case. 😄

linkmauve:
At least in France, Jabber is still way more well-known than XMPP, so I would recommend keeping it even if just for better SEO.

bochecha:
> At least in France, Jabber is still way more well-known than XMPP, so I would recommend keeping it even if just for better SEO.

(indeed, everybody here calls it "Jabber"; I know I do, even though I know I shouldn't; "iksèmepépé" just doesn't roll off the tongue the way "geabeur" does 😜)

How about this:

* keep "Jabber" out of the summary
* have the description contain `Dino is a modern chat application using the XMPP (aka Jabber) network, …`

This way, people searching for "jabber" in their app center will find Dino, but the word Jabber is de-emphasized.

I did that in the latest push to this branch.

bochecha:
This is how Dino looks like in GNOME Software with this appdata file:

dino-in-software

bochecha:
I updated this branch with the screenshots from the website.

I was also wondering if the appdata file shouldn't reuse the wording from the website?

Something like this:

```xml
  <summary>Communicating happiness</summary>

  <description>
    <p>
      Dino is a modern open-source chat client for the desktop. It focuses on providing a clean and reliable Jabber/XMPP experience while having your privacy in mind.
    </p>
  </description>
```

That seems a bit short, though. What do you think?

bochecha:
Working a bit on this with the GNOME Software maintainer (thank you!), I'm finally at a point where I'm happy with this merge request.

Here is how Dino looks now in GNOME Software:

dino-in-software

A few notes:

* the version is "master" because I'm building from Dino master. Once Dino has releases, they need to be reflected in the appdata file (see the `<releases>` tag) so that this appears correctly in software centers;
* the source is my own repo, because that's where I'm pushing test Flatpak builds. Eventually, this would be something like Fedora (for an official package) or Flathub (for Flatpak builds coming from there) or "dino.im" (if you decide to provide your own Flatpak builds), etc… So don't worry about this part for now. :slightly_smiling_face:
* all I want is to click that "Write a Review" button to tell everyone how great Dino is! :smile:

bochecha:
I just pushed an update with @fiaxh's suggestion from the Dino chat room.

bochecha:
Anything else I can do to get this merged? :slightly_smiling_face:

PR #236, opened 2017-12-14
Move to GNetworkMonitor

### LINKED ISSUE #235: Move to GNetworkMonitor
Dino currently talk to NetworkManager over DBus to know the state of the network.

That doesn't work in a Flatpak sandbox by default though, because Flatpak filters DBus communications and only allows a very small set of things to pass (which are known to be safe).

Gio provides an API to know the state of the network (and be notified of changes via a signal): [GNetworkMonitor](https://developer.gnome.org/gio/stable/GNetworkMonitor.html).

And GNetworkMonitor works both inside a Flatpak sandbox, and in traditional builds. (in Flatpak it uses what we call a "portal", which are the clean, safe way to let apps exit their sandbox)

bochecha:
Dino currently talks to NetworkManager over DBus to know the state of
the network.

That doesn't work in a Flatpak sandbox by default though, because
Flatpak filters DBus communications and only allows a very small set of
things to pass (which are known to be safe).

Gio provides an API to know the state of the network (and be notified of
changes via a signal): GNetworkMonitor.

And GNetworkMonitor works both inside a Flatpak sandbox, and in
traditional builds. (in Flatpak it uses what we call a "portal", which
are the clean, safe way to let apps exit their sandbox)

Fixes #235

---

I'm currently testing this and it seems to work.

At least when I turn off the wifi Dino disconnects me, and when I turn it back on Dino connects me back.

I'm testing this inside a Flatpak though, so testing in a more traditional (i.e not sandboxed) environment would be appreciated. 🙂 

Finally, the code was doing two things:

* reacting on `StateChanged` events (i.e the computer is now connected/disconnected when it was previously disconnected/connected)
* reacting to changes of the `PrimaryConnection` (i.e the computer was already connected, it still is, but the default route changed, for example because the user started their VPN)

I completely removed the code handling the second point, because I **think** it is now unneeded. I dug a bit in the GNetworkMonitor code, and if I understood it correctly then it does fire the `network-changed` event when a new route is added, which should cover the "`PrimaryConnection` changed" case.

I have no way to test this, though. Here as well, I would appreciate help testing this.


bochecha:
I've asked a GNOME developer friend to have a look at this pull request and he found a couple issues with the way I was using GNetworkMonitor. (mostly I wasn't checking the "connectivity") Those are now fixed.

libdino/src/service/connection_manager.vala
```diff
@@ -243,19 +243,15 @@ public class ConnectionManager {
     }
 
     private bool network_is_online() {
-        if (network_monitor != null && network_monitor.network_available && network_monitor.connectivity == GLib.NetworkConnectivity.FULL) {
-            return true;
-        }
-
-        return false;
+        return network_monitor != null && network_monitor.network_available;
```
bochecha: As I said in a private chat, I'd like to note here (so that it's written down somewhere public) that this isn't right and the connectivity check really is the correct thing to do. Also, to make sure everyone in the Dino team understands the situation.

* `network_available` means that your computer has network routes to "somewhere". That is, it is connected to a router.
* `connectivity.FULL` means that your computer can access "the internet". That is, if you're behind a router, that router is connected.

As a result, only checking for `network_available` is not correct.

Unfortunately, NetworkManager tends to wait a long time before checking for connectivity. As a result, it is possible that a transient network error leaves NetworkManager thinking that you have `network_available=true` but `connectivity!=FULL`, and it will wait several minutes before realizing that your Internet connexion did come back.

During that time, apps checking for connectivity (e.g the whole GNOME desktop) will think they don't have access to the Internet, while apps that don't (e.g Firefox) will access the Internet just fine. Users are understandably confused when that happens.

And this is what @fiaxh is worried about and wants to avoid.

Removing the check for connectivity might be an acceptable trade-off in the short-term (that's really not for me to decide, it's the job of a maintainer), but it seems to me like we should go and get NetworkManager fixed, so that it avoids these situations. I had told @fiaxh I would at least go and talk to the NetworkManager folks, so I guess I need to own up to that. 🙂

bochecha: Here it is: https://bugzilla.gnome.org/show_bug.cgi?id=792240

I added a comment in the code and the commit message linking to this bug report, so we eventually add the connectivity check back when NetworkManager is improved.

PR #254, opened 2017-12-30
Use libsecret for passwords, if available

RalfJung:
This uses libsecret, if available, to store passwords. Loading and storing passwords becomes async.

I'm not entirely sure about the intended behavior of the account dialog. Right now, the password gets saved when the password entry field turns back into a label. However, this means that if the account is enabled while the password is being edited, the password is not changed.

I have not tested what happens when libsecret is compiled in, but no keyring is available at run-time.

Supersedes #173


libdino/CMakeLists.txt
```diff
@@ -4,9 +4,16 @@ find_packages(LIBDINO_PACKAGES REQUIRED
...
     GModule
     GObject
+    OPTIONAL
     Secret
 )
 
+unset(LIBDINO_EXTRA_OPTIONS)
+if (Secret_FOUND)
+    message(STATUS "libsecret found")
+    set(LIBDINO_EXTRA_OPTIONS ${LIBDINO_EXTRA_OPTIONS} -D WITH_SECRET)
```
linkmauve: No space after `-D`.

libdino/src/entity/account.vala
```diff
@@ -73,6 +69,38 @@ public class Account : Object {
         notify.connect(on_update);
     }
 
+    public async string? get_password() {
+#if WITH_SECRET
+        if (password != "") {
```
linkmauve: Please invert this check, it will hopefully be more likely for the password to come from the secret service than from the db.  You can then drop the else clause and have this part one indent level lower.

libdino/src/entity/account.vala
```diff
@@ -73,6 +69,38 @@ public class Account : Object {
...
+        return this.password;
+#endif
+    }
+
+    public async void set_password(string password) {
+#if WITH_SECRET
+        if (yield secret_manager.set_password(password)) {
+            // Success.  Don't keep anything locally.
+            this.password = "";
+        } else
```
linkmauve: Here too you can replace the else with a return inside of the if.

libdino/src/service/secret_manager.vala
```diff
@@ -29,15 +30,18 @@ public class SecretManager {
...
+    public async bool set_password(string password) {
         try {
-            Secret.password_store_sync(schema, Secret.COLLECTION_DEFAULT,
+            return yield Secret.password_store(schema, Secret.COLLECTION_DEFAULT,
                                        "XMPP account " + user + "@" + server, password, null,
                                        "user", user,
                                        "server", server,
                                        "protocol", "xmpp");
-        } catch (GLib.Error e) { }
+        } catch (GLib.Error e) {
```
linkmauve: Would it make sense to propagate this error to the caller, so they could report the full error to the user?

main/src/ui/manage_accounts/dialog.vala
```diff
@@ -44,7 +44,20 @@ public class Dialog : Gtk.Dialog {
         });
         image_button.clicked.connect(show_select_avatar);
         alias_hybrid.entry.key_release_event.connect(() => { selected_account.alias = alias_hybrid.text; return false; });
-        password_hybrid.entry.key_release_event.connect(() => { selected_account.password = password_hybrid.text; return false; });
+        password_hybrid.switched_to_label.connect(() => {
+            active_switch.sensitive = false;
+            password_hybrid.sensitive = false;
+            var password = password_hybrid.text;
+            password_hybrid.text = "Saving…";
```
linkmauve: Make this translatable, `_("Saving…")`.

main/src/ui/manage_accounts/dialog.vala
```diff
@@ -197,7 +210,14 @@ public class Dialog : Gtk.Dialog {
 
         alias_hybrid.text = account.alias;
         password_hybrid.entry.input_purpose = InputPurpose.PASSWORD;
-        password_hybrid.text = account.password;
+        password_hybrid.visibility = true;
+        password_hybrid.sensitive = false;
+        password_hybrid.text = "Loading…";
```
linkmauve: Make this translatable, `_("Loading…")`.

xmpp-vala/src/module/sasl.vala
```diff
@@ -3,19 +3,23 @@ using Xmpp.Core;
 namespace Xmpp.PlainSasl {
     private const string NS_URI = "urn:ietf:params:xml:ns:xmpp-sasl";
 
+    public interface PasswordProvider : GLib.Object {
+        public abstract async string? get_password();
```
linkmauve: Shouldn’t there also be set_password here?

xmpp-vala/src/module/sasl.vala
```diff
@@ -40,6 +44,14 @@ namespace Xmpp.PlainSasl {
             }
         }
 
+        private async void write_auth_to_stream(XmppStream stream) {
+            var password = yield password_provider.get_password();
+            //stdout.printf ("Password: %s\n", password);
```
linkmauve: Don’t commit commented out lines.

libdino/src/entity/account.vala
RalfJung: I usually prefer else of early-return in cases like this, but with the additional preprocessor conditional I agree that's clearer here.

libdino/src/service/secret_manager.vala
RalfJung: I have no idea. I don't even know under which conditions an error is raised here; the libsecret documentation does not mention this at all.

xmpp-vala/src/module/sasl.vala
RalfJung: It's not needed... so, I don't really care. This is purely for `PlainSasl`, which never changes the password.

xmpp-vala/src/module/sasl.vala
RalfJung: *oops* sorry, this was an accident.

libdino/CMakeLists.txt
RalfJung: Build fails if I do that...
```
FAILED: libdino/src/application.c libdino/src/dbus/login1.c libdino/src/dbus/upower.c libdino/src/entity/account.c libdino/src/entity/conversation.c libdino/src/entity/encryption.c libdino/src/entity/file_transfer.c libdino/src/entity/jid.c libdino/src/entity/message.c libdino/src/entity/settings.c libdino/src/plugin/interfaces.c libdino/src/plugin/loader.c libdino/src/plugin/registry.c libdino/src/service/avatar_manager.c libdino/src/service/avatar_storage.c libdino/src/service/blocking_manager.c libdino/src/service/chat_interaction.c libdino/src/service/connection_manager.c libdino/src/service/conversation_manager.c libdino/src/service/counterpart_interaction_manager.c libdino/src/service/database.c libdino/src/service/entity_capabilities_storage.c libdino/src/service/file_manager.c libdino/src/service/message_processor.c libdino/src/service/message_storage.c libdino/src/service/module_manager.c libdino/src/service/muc_manager.c libdino/src/service/presence_manager.c libdino/src/service/roster_manager.c libdino/src/service/secret_manager.c libdino/src/service/stream_interactor.c libdino/src/service/util.c libdino/src/util.c exports/dino.vapi exports/dino_internal.vapi exports/dino.h exports/dino_internal.h 
cd /home/r/src/dino/build/libdino && /usr/bin/valac -C --header=/home/r/src/dino/build/exports/dino.h --internal-header=/home/r/src/dino/build/exports/dino_internal.h --vapi=/home/r/src/dino/build/exports/dino.vapi --internal-vapi=/home/r/src/dino/build/exports/dino_internal.vapi -b /home/r/src/dino/libdino -d /home/r/src/dino/build/libdino --pkg=gdk-pixbuf-2.0 --pkg=gee-0.8 --pkg=glib-2.0 --pkg=gmodule-2.0 --pkg=gobject-2.0 --pkg=libsecret-1 -DWITH_SECRET -g --thread --target-glib=2.38 /home/r/src/dino/libdino/src/application.vala /home/r/src/dino/libdino/src/dbus/login1.vala /home/r/src/dino/libdino/src/dbus/upower.vala /home/r/src/dino/libdino/src/entity/account.vala /home/r/src/dino/libdino/src/entity/conversation.vala /home/r/src/dino/libdino/src/entity/encryption.vala /home/r/src/dino/libdino/src/entity/file_transfer.vala /home/r/src/dino/libdino/src/entity/jid.vala /home/r/src/dino/libdino/src/entity/message.vala /home/r/src/dino/libdino/src/entity/settings.vala /home/r/src/dino/libdino/src/plugin/interfaces.vala /home/r/src/dino/libdino/src/plugin/loader.vala /home/r/src/dino/libdino/src/plugin/registry.vala /home/r/src/dino/libdino/src/service/avatar_manager.vala /home/r/src/dino/libdino/src/service/avatar_storage.vala /home/r/src/dino/libdino/src/service/blocking_manager.vala /home/r/src/dino/libdino/src/service/chat_interaction.vala /home/r/src/dino/libdino/src/service/connection_manager.vala /home/r/src/dino/libdino/src/service/conversation_manager.vala /home/r/src/dino/libdino/src/service/counterpart_interaction_manager.vala /home/r/src/dino/libdino/src/service/database.vala /home/r/src/dino/libdino/src/service/entity_capabilities_storage.vala /home/r/src/dino/libdino/src/service/file_manager.vala /home/r/src/dino/libdino/src/service/message_processor.vala /home/r/src/dino/libdino/src/service/message_storage.vala /home/r/src/dino/libdino/src/service/module_manager.vala /home/r/src/dino/libdino/src/service/muc_manager.vala /home/r/src/dino/libdino/src/service/presence_manager.vala /home/r/src/dino/libdino/src/service/roster_manager.vala /home/r/src/dino/libdino/src/service/secret_manager.vala /home/r/src/dino/libdino/src/service/stream_interactor.vala /home/r/src/dino/libdino/src/service/util.vala /home/r/src/dino/libdino/src/util.vala /home/r/src/dino/build/exports/xmpp-vala.vapi /home/r/src/dino/build/exports/qlite.vapi
Unknown option -DWITH_SECRET
```

RalfJung:
Updated with review addressed (modulo specific comments above)

xmpp-vala/src/module/sasl.vala
linkmauve: Maybe add a comment with the reason.

libdino/src/entity/account.vala
```diff
@@ -57,6 +69,37 @@ public class Account : Object {
...
+
+    public async void set_password(string password) {
+#if WITH_SECRET
+        if (yield secret_manager.set_password(password)) {
+            // Success.  Don't keep anything locally.
+            this.password = "";
+            return;
+        }
+#endif
+        this.password = password;
```
linkmauve: I’m uneasy with silently falling back to plain text storage when libsecret failed, but I don’t have a good solution to suggest.

RalfJung: In the future, maybe this function could inform its caller how the password was stored, and then the account UI could inform the user. But my UI experience is very limited (ahem ;), so I focused on the core functionality.

RalfJung:
I rebased this on top of current master and made it compile again. What is this blocked on?

RalfJung:
It's very sad to get hardly any feedback on this :(  I am trying to rebase right now, but some async stuff is getting in the way and I am not sure how to proceed.

RalfJung:
Rebasing got stuck because of conflicts with the SCRAM support. Namely, while I could change this code

https://github.com/dino/dino/blob/38b532034f7ff60542a5751a283e734a23263bb7/xmpp-vala/src/module/sasl.vala#L197-L199

to just instead start an async write that will read the password, then write, I can't do the same thing with

https://github.com/dino/dino/blob/38b532034f7ff60542a5751a283e734a23263bb7/xmpp-vala/src/module/sasl.vala#L184-L195

because the password is also use for `stream.add_flag`.

@mar-v-in is it okay do just do that entire thing asynchronously? I have no idea what `add_flag` does and if it is important that it definitely happened when `received_features_node` returned.

RalfJung:
Ideally we should just make `received_features_node` async. But surprisingly, Vala seems to have no support whatsoever for signals that trigger async functions. I would have expected to be able to write `public signal async void foo()`, and then elsewhere `yield foo()` would run all the signal handlers asynchronously and then come back to me. Seems like a big hole in Vala's `async` support?

RalfJung:
Looks like setting the flag asynchronously does *not* work, I get errors saying "mandatory-to-negotiate feature not negotiated: sasl".

mar-v-in:
You can add the flag in the sync part and write the password into it in the async part before sending the `auth` nonza. The password just has to be present in the flag before the server challenge arrives, which obviously won't happen before you send the `auth` nonza.

This part is subject to rewrite/changes anyway, as we'd like to make actual use of SCRAM's feature of not storing the password in cleartext. We could then do the following procedure during login:
1. Use SCRAM key from local database if server supports SCRAM.
2. Use plaintext password from libsecret if secret storage present.
3. Ask user for password.

A bit bad user experience for those without SCRAM and without libsecret, but these should just upgrade their servers ;)

RalfJung:
> You can add the flag in the sync part and write the password into it in the async part before sending the auth nonza. The password just has to be present in the flag before the server challenge arrives, which obviously won't happen before you send the auth nonza.

Ah, I will try that, thanks. Seems a bit messy but what can you do...

RalfJung:
Yes, that did it. :)  Got the patch rebased successfully, and it is working locally.

RalfJung:
This mysteriously stopped working after installing a system upgrade... gnome wouldn't change the dbus API to a service, would they?!?

mar-v-in:
The secret service API is a [freedesktop spec](https://specifications.freedesktop.org/secret-service/), so very unlikely GNOME is going to do any changes to it without publishing an update to the spec first. And even if they did, they'd have updated libsecret, because otherwise many things in GNOME would have stopped working as well (because GNOME apps and services actually use libsecret to access to keyring)

RalfJung:
Yeah, I came to the same conclusion after some research.

Still no idea why it stopped working though.

Af0x:
Hello, please make it work so that Dino saves passwords to gnome-keyring. That would be great

RalfJung:
I got sent an updated version of the patch by Liliana Prikler <liliana.prikler@gmail.com>; with their permission I have updated the PR accordingly. I have not tested this new version of the patch.

Idesmi:
It would be cool to have this merged.

PR #262, opened 2018-01-05
Pass avatars as filenames

### LINKED ISSUE #239: Application icon on debian missing
Hello,

Since a few days ago, there is no application icon on debian anymore.

I'm using the obs package respository for debian.

Thanks in advance.

bebehei:
The GNotification interface does not support Icons of the type
GBytesIcon on libnotify backends. So the avatars passed with a
notification won't get displayed.

Skipping the PixBuf loading steps, will save calculation time
and enable the notification daemon to display the avatars again.

Fixes #239

### DIFF
--- a/main/src/ui/notifications.vala
+++ b/main/src/ui/notifications.vala
@@ -61,9 +61,8 @@ public class Notifications : Object {
             }
             notifications[conversation].set_title(display_name);
             notifications[conversation].set_body(text);
-            try {
-                notifications[conversation].set_icon(get_pixbuf_icon((new AvatarGenerator(40, 40)).draw_conversation(stream_interactor, conversation)));
-            } catch (Error e) { }
+            string filestring = stream_interactor.get_module(AvatarManager.IDENTITY).get_avatar_path(conversation.account, message.counterpart);
+            notifications[conversation].set_icon(new FileIcon(File.new_for_path(filestring)));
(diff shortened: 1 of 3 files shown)


main/src/ui/notifications.vala
```diff
@@ -76,9 +75,8 @@ public class Notifications : Object {
 
         Notification notification = new Notification(_("Subscription request"));
         notification.set_body(jid.bare_jid.to_string());
-        try {
-            notification.set_icon(get_pixbuf_icon((new AvatarGenerator(40, 40)).draw_jid(stream_interactor, jid, account)));
```
linkmauve: You should also remove the `get_pixbuf_icon()` method, which isn’t called anywhere anymore.

bebehei: @linkmauve I've seen this but wasn't sure, if this function could be useful in some later time.

Thanks for reviewing!

bebehei:
@fiaxh As discussed in MUC, I've fixed the bug about missing icons in MUC notifications.

bebehei:
As I got notified in the Dino MUC about the ongoing avatar drawing refactoring, I'm going to close this.

I guess this PR shows the intents and solution if someone wants to rebase it later. I guess I also have elaborated the reasons, that this should be done this way.

PR #266, opened 2018-01-08
Add support for setting a new Affiliation in a muc

slaveriq:
Feedback very welcome. I'm new to the language and new to the codebase.


fiaxh:
Thanks. This functionality should also be available from the UI in the future.

PR #276, opened 2018-01-16
Add more info on build deps

licaon-kter:
As needed on Lubuntu 17.10 at least.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,11 @@ Build
 * ninja(-build) (recommend)
 * valac (≥ 0.34)
 
+Eg. On Ubuntu 17.10 this command should pull most build dependencies
+```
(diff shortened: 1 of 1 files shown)


fiaxh:
Duplicate of #56

licaon-kter:
The current info is not enough.

Please merge this, or edit it yourself in some other format, the fact that I needed to re-run configure a dozen times and hunt for the matching packages names in Synaptic makes this necessary. :(

d9h20f:
Distribution packages for Ubuntu are available: https://github.com/dino/dino/wiki/Distribution-Packages

Generally, if someone wants to manually compile something for which a package for their distro is already available, they already know how to set up a build environment.

licaon-kter:
That's true to a point, my problem is that the info in just partial, either the user `already knows how to` and this is not needed at all, or provide complete info. Now it's: `you kinda need these libs and see whatever error ninja spits out`

PR #279, opened 2018-01-24
Initial Catalan translation.

jmallach:


d9h20f:
You can also translate on Weblate: https://hosted.weblate.org/projects/dino/translations

jmallach:
I was totally unaware. If PRs are ok, I prefer using local tools than a web interface.

PR #280, opened 2018-01-24
Check previous network state on_network_changed before reconnect.

j-licht:
As we know GNetworkMonitor inform us too often about a change,
we should track and check the last state.

I noticed many "network online" messages and then clients of my partners
get a replaced by new connection message.
These messages were sent very often, like every secound and this
is confusioning the other clients, so they cannot send messages to me.

I'm not sure wether the problem is the GNetworkMonitor or the
reconnecting process.

I've problems to debug the reconnect problem because it takes some
random time until the many reconects take part.

### DIFF
--- a/libdino/src/service/connection_manager.vala
+++ b/libdino/src/service/connection_manager.vala
@@ -25,6 +25,7 @@ public class ConnectionManager {
     private Login1Manager? login1;
     private ModuleManager module_manager;
     public string? log_options;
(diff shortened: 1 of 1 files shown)


fiaxh:
The code change will result in problems when enabling a VPN. The state will change from online to online, but the connection will be broken, thus would need to be re-established.

If you still have this problem, please open an issue with any information you have.

PR #295, opened 2018-02-17
Add certificate pinning

### LINKED ISSUE #57: No connection to account after building from source (SSL fail?)

f1ori:
add dialog to inspect and pin a certificate which
is not trusted by the system per default

This is a first shot. Please check if this is the right approach. At least the ui needs some more work.
Fixes #57

### DIFF
--- a/libdino/src/service/database.vala
+++ b/libdino/src/service/database.vala
@@ -17,10 +17,11 @@ public class Database : Qlite.Database {
         public Column<bool> enabled = new Column.BoolInt("enabled");
         public Column<string> roster_version = new Column.Text("roster_version") { min_version=2 };
         public Column<long> mam_earliest_synced = new Column.Long("mam_earliest_synced") { min_version=4 };
+        public Column<string> certificate = new Column.Text("certificate");
 
         internal AccountTable(Database db) {
             base(db, "account");
-            init({id, bare_jid, resourcepart, password, alias, enabled, roster_version, mam_earliest_synced});
+            init({id, bare_jid, resourcepart, password, alias, enabled, roster_version, mam_earliest_synced, certificate});
         }
     }
 
--- a/main/data/manage_accounts/dialog.ui
+++ b/main/data/manage_accounts/dialog.ui
@@ -153,13 +153,34 @@
                                                     </packing>
                                                 </child>
                                                 <child>
+                                                    <object class="GtkButton" id="is_trusted_button">
+						        <property name="relief">none</property>
+                                                        <property name="halign">end</property>
+                                                        <property name="valign">center</property>
+                                                        <property name="visible">True</property>
+                                                        <child>
+                                                            <object class="GtkImage" id="is_trusted_image">
+								<property name="icon-name">changes-prevent-symbolic</property>
+                                                                <property name="visible">True</property>
+								<property name="icon-size">1</property>
+                                                            </object>
+                                                        </child>
+                                                    </object>
+                                                    <packing>
+                                                        <property name="left_attach">2</property>
+                                                        <property name="top_attach">0</property>
+                                                        <property name="width">1</property>
+                                                        <property name="height">2</property>
+                                                    </packing>
+                                                </child>
+                                                <child>
                                                     <object class="GtkSwitch" id="active_switch">
                                                         <property name="visible">True</property>
                                                         <property name="halign">end</property>
                                                         <property name="valign">center</property>
                                                     </object>
                                                     <packing>
-                                                        <property name="left_attach">2</property>
+                                                        <property name="left_attach">3</property>
                                                         <property name="top_attach">0</property>
                                                         <property name="width">1</property>
                                                         <property name="height">2</property>
@@ -191,7 +212,7 @@
                                                     <packing>
                                                         <property name="left_attach">1</property>
                                                         <property name="top_attach">2</property>
-                                                        <property name="width">2</property>
+                                                        <property name="width">3</property>
                                                         <property name="height">1</property>
                                                     </packing>
                                                 </child>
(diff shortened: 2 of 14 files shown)


ThibG:
I have two nitpicks with this contribution:
- I think it would be better to pin the public key instead of the certificate as a whole
- Reading the code, it appears it doesn't actually *pin* a certificate, but rather *ignore errors* for a given certificate (e.g., if the certificate/public key changes to a valid one, no warning will be issued)

xmpp-vala/src/module/tls.vala
```diff
@@ -25,10 +25,15 @@ namespace Xmpp.Tls {
                     var io_stream = stream.get_stream();
                     if (io_stream == null) return;
                     var conn = TlsClientConnection.new(io_stream, identity);
+                    var pinned_certificate = stream.pinned_certificate;
+                    var flag = stream.get_flag(Flag.IDENTITY);
+                    conn.accept_certificate.connect((conn, server_cert, errors) => {
+                        flag.used_pinned_certificate = true;
```
ThibG: This is confusing. This flag basically says that initial verification failed, not that a pinned certificate was used.

main/src/ui/manage_accounts/certificate_pinning_dialog.vala
```diff
@@ -0,0 +1,70 @@
...
+        if (server_certificate != null) {
+            Certificate gcr_certificate = new SimpleCertificate(server_certificate.certificate.data);
+            certificate_widget.set_certificate(gcr_certificate);
+        } else {
+           info_label.label = "";
+           pin_certificate_button.sensitive = false;
+           return;
+        }
+        if (!used_pinned_certificate) {
+           // only show certificate, no pinning needed
```
ThibG: It is currently ignoring errors for a given certificate rather than pinning it, but if we move to actually pinning a key/certificate, we would probably want to be able to pin a valid one.

main/data/manage_accounts/dialog.ui
```diff
@@ -152,14 +152,35 @@
...
+                                                            </object>
+                                                        </child>
+                                                    </object>
+                                                    <packing>
+                                                        <property name="left_attach">2</property>
+                                                        <property name="top_attach">0</property>
+                                                        <property name="width">1</property>
+                                                        <property name="height">2</property>
+                                                    </packing>
+                                                </child>
```
ThibG: I am not sure whether this is the most appropriate icon, and it probably needs a tooltip.

xmpp-vala/src/module/tls.vala
```diff
@@ -25,10 +25,15 @@ namespace Xmpp.Tls {
...
+                        return pinned_certificate != null && server_cert.is_same(pinned_certificate);
```
ThibG: It would be better to pin public keys rather than certificates.

libdino/src/service/database.vala
```diff
@@ -17,10 +17,11 @@ public class Database : Qlite.Database {
         public Column<bool> enabled = new Column.BoolInt("enabled");
         public Column<string> roster_version = new Column.Text("roster_version") { min_version=2 };
         public Column<long> mam_earliest_synced = new Column.Long("mam_earliest_synced") { min_version=4 };
+        public Column<string> certificate = new Column.Text("certificate");
```
ThibG: The version of the database should be bumped, with an appropriate “min_version” for the certificate column, otherwise, the database will be lacking the “certificate” column, leading to crashes.

rigelk: Indeed, testing the PR with a previous db yields `database.vala:127: SQLite error: 1 - no such column: certificate` errors.

xmpp-vala/src/module/tls.vala
rigelk: @ThibG are you aware of any method to extract the public key from the certificate in Vala?

ThibG: @rigelk sadly not, neither glib nor gcr seem to expose anything useful for that. I fear doing that would require exporting the public cert and writing your own code to manually extract it :/

rigelk: @ThibG that's what I saw before asking.
If I find some time to examine this problem, should I use [openssl](https://raw.githubusercontent.com/nemequ/vala-extra-vapis/master/openssl.vapi) bindings or [gcrypt](https://raw.githubusercontent.com/nemequ/vala-extra-vapis/master/gcrypt.vapi) bindings? Both seem to lack any function related to certificate management, as of now anyway. 

I could add the required bindings though.

fiaxh:
Closing due to the very valid issues raised by @ThibG and @rigelk and the lack of reactions.

PR #311, opened 2018-03-19
Added a few smileys -> emoji combinations

SpaceCafe:


d9h20f:
You're changing a whole bunch of different emoticons to the same emoji. I don't think ":3" and "XD" look anything like, but you've made them into the same "😁" face - which itself doesn't look like either of the aforementioned emoticons.

sbernard31:
XD is more = 😆, 
:3 [maybe a cat face](https://www.quora.com/What-does-the-emoticon-3-mean?share=1) : 😺 ?

more example : https://developer.gnome.org/gtk3/stable/GtkIMContextSimple.html

main/src/ui/chat_input/smiley_converter.vala
```diff
@@ -13,18 +13,41 @@ class SmileyConverter {
...
+        smiley_translations[":-("] = "🙁";
         smiley_translations[":'("] = "😢";
+        smiley_translations[":'-("] = "😢";
+        smiley_translations[":|"] = "😐";
+        smiley_translations[":-|"] = "😐";
         smiley_translations[":/"] = "😕";
+        smiley_translations[":-/"] = "😕";
+        smiley_translations[":*"] = "😘";
+        smiley_translations[":-*"] = "😘";
+        smiley_translations["<3"] = "💓";
```
dos1: Shouldn't it rather be ❤?

fiaxh:
This PR is similar to #42:

> For now the list will stay as it is, containing only the smileys with the most obvious unicode counterpart. If everyone just adds the replacements they would like to have, in the end this feature will become unsuitable for everyone.

PR #313, opened 2018-03-21
Change offline users' avatar to grayscale in the “add conversation list”

ThibG:
It makes the behavior consistent with the conversations list and makes online
users easier to spot when starting a conversation.

### DIFF
--- a/main/data/add_conversation/list_row.ui
+++ b/main/data/add_conversation/list_row.ui
@@ -10,7 +10,7 @@
                 <property name="visible">True</property>
                 <child>
                     <object class="DinoUiAvatarImage" id="image">
-                        <property name="allow_gray">False</property>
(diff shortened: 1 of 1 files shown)


stevenroose:
I like this! @fiaxh :)

Sylvhem:
I just tested this feature and it's working as intended. It's really useful to see quickly who is online and who isn't.

fiaxh:
Removing the colors from the avatars takes away most of their "quick recognizably". This is more problematic in a long list which the user has to search through (contact list) than in a list that is mostly sorted according to current relevance (conversation list).

However I understand that it is desirable to know the online status in that dialog as well. We'll probably go along the way of adding little green/yellow/grey dots to indicate status.

PR #316, opened 2018-03-22
Do not mark messages as read if they are only marked as received

### LINKED ISSUE #305: Incoming messages are not marked as unread
Sometimes, incoming messages are not marked as unread.
Unfortunately, I don't know how to reproduce that reliably. It may be related to the fact that I am also using Conversations at the same time.

ThibG:
Fixes #305, or at least some occurrences of it.

### DIFF
--- a/libdino/src/service/counterpart_interaction_manager.vala
+++ b/libdino/src/service/counterpart_interaction_manager.vala
@@ -66,6 +66,7 @@ public class CounterpartInteractionManager : StreamInteractionModule, Object {
     private void on_chat_marker_received(Account account, Jid jid, string marker, string stanza_id) {
         bool own_marker = account.bare_jid.to_string() == jid.bare_jid.to_string();
         if (own_marker) {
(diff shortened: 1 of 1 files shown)


libdino/src/service/counterpart_interaction_manager.vala
```diff
@@ -66,6 +66,7 @@ public class CounterpartInteractionManager : StreamInteractionModule, Object {
     private void on_chat_marker_received(Account account, Jid jid, string marker, string stanza_id) {
         bool own_marker = account.bare_jid.to_string() == jid.bare_jid.to_string();
         if (own_marker) {
+            if (marker == Xep.ChatMarkers.MARKER_RECEIVED) return;
```
fiaxh: Just a *very* minor thing: I think it would be clearer to have this condition switched around - to to the following for MARKER_DISPLAYED and MARKER_ACKNOWLEDGED.

ThibG: Ok, fixed!

fiaxh:
Thanks!

PR #321, opened 2018-03-27
Prefill nick with local alias when joining a non-bookmarked MUC

### LINKED ISSUE #223: Use the “local alias” as the default nick when joining a non-bookmarked MUC
And possibly bookmark the room on join using this nick as the default one, so that other clients will synchronise to the same nick.

bobufa:
fixes #223


fiaxh:
Thanks!

PR #341, opened 2018-04-23
Switch subject and display name (MUC address) + display MUC subject in conversation selector

### LINKED ISSUE #21: Show MUC name instead of MUC JID in roster
in particular since Conversations creates arbitrary digits/letters (I do not understand why, but ok), it makes it hard to grasp what the MUC is actually about when the topic is not shown in the roster.

codedust:
Closes  #21

Before/After:
dino_titlebar

### DIFF
--- a/main/src/ui/conversation_selector/conversation_row.vala
+++ b/main/src/ui/conversation_selector/conversation_row.vala
@@ -62,7 +62,11 @@ public abstract class ConversationRow : ListBoxRow {
     }
 
     protected void update_name_label(string? new_name = null) {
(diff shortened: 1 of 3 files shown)


d9h20f:
I suggest: if there is a name and a topic, use the name, and show the topic below (the URL can always still be found in the MUC details, or by hovering over its tab in the side pane). If there is a name but no topic, use the name, and show the URL below. If there is only a topic, use the topic and show the URL below. If there is neither a name nor a topic, just use the URL.

fiaxh:
I'd prefer the MUC display name to only be its name with a fallback to the JID for simplicity and consistency between different MUCs and with one-to-one chats.

Issue #299, opened 2018-03-01
Run Dino in background


schiessle:
Dino is a really nice looking and modern XMPP client. What stops me from using it on a daily basis is that I can't run it in the background. I would like to autostart it during login in the background so that I get notification as soon as someone writes me a message or mention me in a group chat.

Is something like this already planed?

d9h20f:
I think any operating system and DE allows for autostarting applications. You can move it to another workspace (or just minimise it) if you don't want to see the window.

GladOSkar:
I think this is a duplicate of #98

schiessle:
You could also look how Polari, the Gnome IRC client, does it the "Gnome way" without the need of a system tray.

benrob0329:
@schiessle i.e. just start minimized?

schiessle:
@benrob0329 I would call it "starting in the background". In Gnome (what I care most) it shouldn't be visible in the activities overview, still notifications should pop up. If I close the chat window it should again go back in the background. I think Polari handles it quite well.

In a non Gnome setup, background could be "minimize to tray". If Dino wants to target non Gnome desktops.

gllmhyt:
> I think Polari handles it quite well and I think it could be worth looking at how they did it.

[Polari](https://wiki.gnome.org/Apps/Polari) has [this file](https://gitlab.gnome.org/GNOME/polari/blob/master/data/org.gnome.Polari.Autostart.desktop) that starts the client (`gapplication action org.gnome.Polari start-client`) hidden (`NoDisplay=True`).

```
[Desktop Entry]
Type=Application
Exec=gapplication action org.gnome.Polari start-client
NoDisplay=True
```

Moreover, Polari has the ability to be ran in background then avoid annoying leave/join and so on. [Here are the commits](: https://bugzilla.gnome.org/show_bug.cgi?id=770750#c14) that made it possible.

> When set up to run in the background, the application:
> * sets itself up to start in the background on login
> * keeps running after the last window has been closed
> * can be stopped via the 'quit' action in the application menu
> ([source](https://gitlab.gnome.org/GNOME/polari/commit/00ab02c))

notramo:
@fiaxh any progress on this?
I suggest that until systray support will be added, it could run in the background simply without system tray:
- The process is autostarted in background, and if a **message is received**, a **notification** would appear.
- Optionally, the **window is raised** and (optionally) **focused on new message**.
[...]

PR #343, opened 2018-04-29
flatpak: Add a build manifest

bochecha:
With this one can build Dino from master with `flatpak-builder`.

---

If you want to try this out, follow this procedure:

1.  install Flatpak and add the Flathub repository (see [instructions](https://flatpak.org/setup/))
2.  install the `flatpak-builder` package
3.  checkout this branch, then build and install Dino:
    ```
    $ flatpak-builder --install-deps-from=flathub --install --force-clean --ccache app im.dino.Dino.json
    ```
4.  you can now run Dino from your desktop by clicking on the icon, or with the command line:
    ```
    $ flatpak run im.dino.Dino
    ```

---

Do note that OpenPGP will not work with this build. That's because the app is sandboxed and as such doesn't have access to `~/.gnupg/`.

I didn't grant it access because the only way to make this work (that I could find) is to give the app read-write access to the whole `~/.gnupg/` folder, which is a pretty big security hole.

It's not worse than non-sandboxed apps of course (they can do whatever they want anywhere in your home directory), so if you feel like the functionality is more important than the security concern (and it probably is… what good is security if you can't do anything?) then I'm happy to edit this pull requestand make it work.

I do think it's important to consider the problem and think about it though, which is why I didn't do it in the initial submission.

---

One thing that can be done with this is to have a server somewhere build the Flatpak every time the master branch is updated, and publish the app in a repo somewhere under https://dino.im/.

I'll be happy to help set this up if you're interested (it requires a few more steps than I detailed above), but this would probably come later, let's not get ahead of ourselves.

Alternatively, once Dino has a release I volunteer to get it on Flathub, if that's what you want. :slightly_smiling_face:

### DIFF
--- a/im.dino.Dino.json
+++ b/im.dino.Dino.json
@@ -1,0 +1,52 @@
+{
+    "id": "im.dino.Dino",
+    "runtime": "org.gnome.Platform",
+    "runtime-version": "3.30",
+    "sdk": "org.gnome.Sdk",
+    "command": "dino",
+    "finish-args": [
+        "--socket=x11",
+        "--share=ipc",
+        "--socket=wayland",
+        "--share=network",
+        "--filesystem=xdg-run/dconf", "--filesystem=~/.config/dconf:ro",
+        "--talk-name=ca.desrt.dconf", "--env=DCONF_USER_CONFIG_DIR=.config/dconf"
+    ],
+    "modules": [
+        {
+            "name": "libgee",
+            "make-install-args": [
+                "girdir=/app/share/gir-1.0",
+                "typelibdir=/app/lib/girepository-1.0"
+            ],
+            "sources": [
+                {
+                    "type": "archive",
+                    "url": "https://download.gnome.org/sources/libgee/0.20/libgee-0.20.1.tar.xz",
+                    "sha256": "bb2802d29a518e8c6d2992884691f06ccfcc25792a5686178575c7111fea4630"
+                }
+            ]
+        },
+        {
+            "name": "libqrencode",
+            "sources": [
+                {
+                    "type": "archive",
+                    "url": "https://fukuchi.org/works/qrencode/qrencode-4.0.2.tar.bz2",
+                    "sha256": "c9cb278d3b28dcc36b8d09e8cad51c0eca754eb004cb0247d4703cb4472b58b4"
+                }
+            ]
+        },
+        {
+            "name": "dino",
+            "buildsystem": "cmake-ninja",
+            "builddir": true,
+            "sources": [
+                {
+                    "type": "git",
+                    "url": "https://github.com/dino/dino.git"
+                }
+            ]
+        }
+    ]
+}


bochecha:
I forgot to mention that I've been using Dino as a Flatpak since at least November 2017. It's working great.

I have it in a personal repository (use at your own risk, this is mostly for me to test stuff out, I do not provide any support for this repository) :

```
$ flatpak remote-add https://www.daitauha.fr/static/flatpak/bochecha.flatpakrepo
$ flatpak install bochecha im.dino.Dino
```

(If you're wondering why I didn't submit this earlier, I had to wait for pull requests #220, #221, #231 and #236 to be merged first. :slightly_smiling_face: )

Echolon:
Hi,

sorry, I'm a total console noob :(
Did all steps, when I entered the third it says: 

_flatpak-builder: command not found_

Any help?
But thanks a lot for doing this!

bochecha:
@Echolon like I said in the first step, you must install `flatpak-builder`, it's a separate package from the `flatpak` one.

And please, this is not a support forum. Let's keep this thread about reviewing the pull request. :wink:

barthalion:
Instructions for flatpak-builder could be probably somewhat simplified. SDK dependency can be installed automatically with `--install-deps-from=flathub` and application installed after build with `--user --install`. But besides this nitpick, looks good!

About storing builds somewhere, I think @TingPing was experimenting with using Travis for Citra emulator, but I'm not sure how it went.

bochecha:
@barthalion thanks! I didn't know about `--install-deps-from` and `--install` I simplified the test instructions. Much nicer. :slightly_smiling_face:

mar-v-in:
@bochecha thanks for your work so far.

I think we should wait for flatpak adding support for a `--socket=gpg-agent` functionality. This was discussed several times in different issues in the flatpak community, but it might be a good idea to create a dedicated issue for it in the [flatpak repo](https://github.com/flatpak/flatpak). Recently `--socket=ssh-agent` support was merged, so I guess it shouldn't be too hard to add `gpg-agent` for them.

This would be the only clean solution for the gpg issue. I dislike the idea of disabling gpg support inside the flatpak (as users will wonder why it doesn't work for them but does work for others) and as gpg now places its socket outside `~/.gnupg` on some systems, just forwarding `~/.gnupg` alone won't work in all cases. We would be required to start another gpg-agent inside the container which really seems to be the wrong approach to me.

What do you think?

bochecha:
@mar-v-in I think waiting for `--socket=gpg-agent` would be the right thing to do, but:

1. I don't think we need to wait for it before we start reviewing this Flatpak manifest, testing it, etc… So I'd still like some feedback anyway. :slightly_smiling_face: 

2. let's make sure an issue is opened upstream Flatpak for GPG handling (I'll try and do that) so that we don't wait forever in vain.

fiaxh:
I opened an issue for flatpak a while back: [#2301: gpg-agent socket](https://github.com/flatpak/flatpak/issues/2301).

phako:
Btw. with the 3.30 runtime this fails to install the libgee girfile.

bochecha:
@phako yeah, that's a classic issue we've been having for a while, since FDO//18.08 switched to pkgconf as the pkg-config implementation.

This is the root of the issue: https://git.dereferenced.org/pkgconf/pkgconf/issues/2

This is one way to work around it: https://gitlab.gnome.org/GNOME/gnome-contacts/merge_requests/26/diffs

But that's just as much a workaround as using the env vars in the first place, so I'm looking at fixing it properly once and for all now.

bochecha:
Latest push to this branch now builds with GNOME SdK 3.30.

Mikaela:
Hi, are there any news on getting Dino officially to Flatpak? I am trying Dino again due to Gajim's issue with unscrollable MUC lists and I am encountering just a few issues:

* OMEMO wasn't selectable for one contact, but apparently I wasn't in their roster and when they scanned my QR that got fixed.
* https://github.com/dino/dino/issues/98 / https://github.com/dino/dino/issues/299 as open windows are distracting even if I sent them to workspace 10
* I am not able to type \` and similar characters for using fcitx, because the flatpak is missing permission `--talk-name=org.freedesktop.portal.Fcitx` and I was able to workaround it with `flatpak override --user --talk-name=org.freedesktop.portal.Fcitx im.dino.Dino` (flatpak override --user --talk-name=org.freedesktop.portal.Fcitx im.dino.Dino)

Would it be possible to have an update to the current repository in case there have been fixes since January @bochecha ? Thank you for maintaining it and in advance :purple_heart: 

I prefer to use the Flatpak as Flatpaks generally seem to be more likely to receive updates timely than distribution versions and the current version of Dino in Debian is from 20181129 and thus even older.

By the way it looks like there is also `org.gnome.Platform 3.32` in addition to the 3.30 which is previously mentioned here.

bochecha:
> Hi, are there any news on getting Dino officially to Flatpak?

This is still blocked on the same thing, as mentioned in https://github.com/dino/dino/pull/343#issuecomment-402516387.

There are still 3 options:

* wait for Flatpak to gain `--socket=gpg-agent`
* poke a hole in the sandbox so Dino has read-write access to the right paths (where the gpg agent socket is)
* accepts disabling GPG functionality in the flatpak

So far the Dino team has chosen the first option, which is perfectly reasonable. They may reevaluate at some point, and if they do I'm sure there would be activity in this pull request. As you can see there hasn't been any, so this is still the status quo.

> Would it be possible to have an update to the current repository in case there have been fixes since January @bochecha ?

I do not use XMPP any more, and as such am not really interested in Dino any more. I want to state that this is through no fault of the Dino developers: Dino is without a doubt the best XMPP client I have ever used.

But as a result, I'm not going to keep updating the repository any more.

I'm happy to follow up with this pull request until it gets merged, though. :slightly_smiling_face:

Peque:
@bochecha Isn't that already supported? https://github.com/flatpak/flatpak/issues/1438

bochecha:
@Peque well, no, that closed issue is about the SSH socket, not the GPG one. :confused:

rugk:
Any news here? Why not just merge this and publishit?

bochecha:
> Any news here? Why not just merge this and publish it?

Please read the rest of the conversation before asking the same question.

https://github.com/dino/dino/pull/343#issuecomment-501733649

rugk:
But I don't need gpg-agent support. This would only be useful when you actually use pgp to encrypt messages, would not it?
AFAIK, dino anyway implements OMEMO as a first-citizen e2e encryption and thus pgp is only used as a fallback for those, who want it.

That's why IMHO, you don't really need to wait, do you? And if the dino devs really want to have this feature for a stable version, then why not release the current version without gpg support as "beta" or so?

bochecha:
@rugk you're not bringing any new arguments to this discussion that haven't been considered previously. All you're doing is adding noise to this page, making it longer and harder to find anything in it. Please stop.

Issues and pull requests are a work place, not a discussion forum. If you want to try and convince people, do it in the discussion forum: the chat@dino.im room on XMPP.

As I said previously, I don't use Dino any more, but since I started this I'd like to finish it instead of disappearing. Every time someone asks the same question again when the answer is already clearly stated in this page, my motivation shrinks a bit more.

I listed [earlier](https://github.com/dino/dino/pull/343#issuecomment-501733649) the 3 possibilities to make progress on this:

> 1. wait for Flatpak to gain `--socket=gpg-agent`
> 2. poke a hole in the sandbox so Dino has read-write access to the right paths (where the gpg agent socket is)
> 3. accepts disabling GPG functionality in the flatpak

The maintainers chose number 1, so we're waiting. Until that changes and either 2 or 3 (or a 4th option I could have missed) is chosen there is absolutely nothing of value to add.

To the maintainers: given all the above I'm going to unsubscribe from this pull request. Do @-mention me when the time has come to resume work on this and I will gladly help push this forward. :slightly_smiling_face:

rugk:
In which flatpak issue is gpg agent functionality requested/tracked?

TingPing:
https://github.com/flatpak/flatpak/issues/2301

rugk:
There is an alternative solution as pointed out in https://github.com/flatpak/flatpak/issues/2301#issuecomment-551021176:

> as a quick solution you might want to have a peek into evolution's  build instructions. They get GnuPG support by mounting the keyring  directory. (Of course not perfect, but at least workable and not less  secure than running it outside of Flatpak on an average distro.)

Kekun:
You will need something like this now:
```
        {
            "name": "libsignal-protocol-c",
            "buildsystem": "cmake-ninja",
            "builddir": true,
            "config-opts": [
                "-DBUILD_SHARED_LIBS=true"
            ],
            "sources": [
                {
                    "type": "archive",
                    "url": "https://github.com/signalapp/libsignal-protocol-c/archive/v2.3.2.tar.gz",
                    "sha256": "f3826f3045352e14027611c95449bfcfe39bfd3d093d578c70f70eee0c85000d"
                }
            ]
        },
```

phako:
libsignal-protocol-c master seems to have some memleak fixes that sound worth including

phako:
But on the bright sight, if you switch the runtime to 3.36 you can remove libgee from the flatpak

Kekun:
> libsignal-protocol-c master seems to have some memleak fixes that sound worth including

I picked that one because Dino master requires that very specific version apparently.

selurvedu:
What do you think of disabling OpenPGP for Flatpak builds?

It's just a plugin that may be easily disabled when compiling Dino by a single CMake option `-D DINO_PLUGIN_ENABLED_openpgp=no`. I did not find any hard-coded references to it in the sources. If my observation is correct and Dino doesn't depend on it in any way, then disabling it should be totally safe.

In my humble opinion, having a usable Flatpak build now, even with a missing feature, wold be better than not having any builds at all for a few more years.

Saroufim:
Just a quick note, it is easy to rebase it to the GNOME 40 runtime if gspell and libhandy/libadwaita are added

alexanderadam:
@fiaxh, [did your opinion change regarding having package specific configurations in the repository](https://github.com/dino/dino/pull/121#issuecomment-321332304)? :thinking:

rugk:
> In my humble opinion, having a usable Flatpak build now, even with a missing feature, wold be better than not having any builds at all for a few more years.

One year later and 4 years since this was implemented/done here… (only a merge is missing this is so sad… :cry:) TBH this is what is actually holding me back from using XMPP at all. Dino would be an awesome desktop client, and I don't need PGP anyway (who even uses that if you have OMEMO?), so IMHO just go and release that… :shrug:

mar-v-in:
@rugk are you looking for a stable or nightly flatpak? What's the problem with the currently available packages?

rugk:
Stable, and the problem is that it is not easily installable on [Fedora Silverblue](https://silverblue.fedoraproject.org/) e.g., which features Flatpaks as _the_ intended way to install desktop applications. And also all other Linux distros, because it is the only way to get it with sandboxing (disregarding snap, because that is TBH, Ubuntu-only), which is especially useful considering the application processes untrusted data from the outside.

Saroufim:
Having a flatpak manifest is enough for me as a Fedora Silverblue user. I can build the application myself. I'm currently stuck trying to write a working flatpak  manifest in order to update the current package I have which was built last year.

zocker-160:
https://github.com/flatpak/flatpak/pull/4958 has been merged a few hours ago, I think there shouldn't be any blockers for a Dino Flatpak now.

RomanRiabenko:
> 1. wait for Flatpak to gain `--socket=gpg-agent`

Hello @bochecha , I just wanted to let you know that the fix for  [flatpak/flatpak#4958](https://github.com/flatpak/flatpak/pull/4958) was released in Flatpak 1.14.0.
Thank you!

mbridon:
Hi, my account used to be "bochecha",but due to shit circumstances and being hospitalized for 2 years I lost access to that account and had to recreate a new one as the Github support people were incapable of giving me back access to it :disappointed: 

So from now on you can mention me with @mbridon.

In any case, during those two years I stoppped using XMPP entirely and have lost any interest in providing a Dino flatpak, sorry.

I'm very happy the gpg socket support was added though, I just won't work on Dino any more.

May you all have a good continuation in your endeavors. :tada:

zocker-160:
In this case, then the PR should be closed I think as the current version is outdated and nobody else is able to push updates to this PR.

PR #346, opened 2018-04-30
Fix translatable string

licaon-kter:
missed in https://github.com/dino/dino/commit/f0dd0e0c3af10f2772efd94aadbded38f3c155bb


fiaxh:
It was done in 4067e45.

PR #348, opened 2018-05-01
Fix consistency Jabber

licaon-kter:


fiaxh:
The "Jabber" v.s. "XMPP" issue is not an easy one, but for now we would prefer to only mention "Jabber" in the more detailed description.

PR #349, opened 2018-05-02
Shortened sentences, Jabber → XMPP

comradekingu:


d9h20f:
There's nothing wrong with 'synchronizes', just say 'free software' instead of 'libre software', and "A clean experience." is not a sentence.

comradekingu:
@licaon-kter How do you easily describe XMPP being the protocol and Jabber being "the ecosystem"?
I thought the name had just changed to XMPP. From Jabber, to Jabber/XMPP, to XMPP. It rolls off the tongue more easily than IRC.

If xmpp.org/software/clients.html says those are XMPP clients, i don't think it is a winning battle, even if somehow correct.

These are the things that call themselves XMPP clients.

gajim.org
swift.im github.com/agl/xmpp-client
www.profanity.im
conversejs.org www.jsxc.org
psi-im.org
getkaiwa.com www.isode.com/products/swiften.html
movim.eu
www.xabber.com
www.npmjs.com/package/node-xmpp-client
https://astrachat.com/MobileDirectConnectXMPP.aspx

There are others still, my point is I only found Dino and two instances of "aka Jabber" when searching for "Jabber/XMPP client".

@d9h20f I don't care either way, it is just shorter for no other benefit than being shorter, so a small gain.
"Free software" introduces errors in translation when invariably it results in some variant of "gratis software", likening a fine point to that of freeware.
"Free software" is an improvement in being more specific than "open source", to be precise it is copyleft libre software.

"A clean experience" is a sentence, albeit also curiously not one that I wrote.
"It focuses on providing a clean and reliable Jabber/XMPP experience"

is in my view only two needless abstractions over 

"A clean and reliable XMPP client"

It isn't a lie to say that it is, and cutting out "experience" only strengthens the position of objectively describing it directly as such, rather than the subjective experience of it.

Sure the intended experience might be different from what one would want to describe the software as, but that seems not the point. If there isn't a poignant point to be made about going with the "experience" of it, i would argue a corollary of the shorter style of brevity, fits nicely with "clean".

licaon-kter:
Yeah great, _ehxempeepee_, riiight.

d9h20f:
You're right about the translation issues to other languages, but I'd rather add that as a translator's note than use 'libre', which is exclusively used in contexts related to free software and the double meaning of the English word 'free'.

@licaon-kter: eye are see (I'm aware some languages actually pronounce it as a word)? you ass bee? eye ass pee? gee pee gee/pee gee pee? Phonetically spelling out acronyms is always going to make them look stupid, so that's not an argument.

stevenroose:
Also, don't change the copyright date like that. If the year is passed, change it to a range: `(R) 2017-2018`.

stevenroose:
Seems like a very useless PR as a whole, though :D

fiaxh:
> Dino is a clean and reliable libre software XMPP chat client for the desktop, keeping your privacy in mind.

I don't think we have to put as much as possible into one sentence. Currently the first sentence only describes what Dino is, the second one says what makes it special.

> A clean and reliable XMPP experience keeping privacy in mind.

> syncs

We should stick to full sentences and complete words.

PR #350, opened 2018-05-02
Fix parsing of multiple --(enable|disable)-plugin options

link2xt:
Due to missing `$` variables `$ENABLED_PLUGINS` and `$DISABLED_PLUGINS` were not evaluated.


fiaxh:
Thanks, merged in 69b886e.

PR #359, opened 2018-05-06
Implement XEP-0380 for the OMEMO and OpenPGP plugins

### LINKED ISSUE #334: Support XEP-0380: Explicit Message Encryption
It would be nice if Dino could mark encrypted messages with an [XEP-0380][1] tag.  E.g., for OMEMO messages, Conversations adds this:

```xml
    <encryption xmlns='urn:xmpp:eme:0'
                name='OMEMO'
                namespace='eu.siacs.conversations.axolotl'/>
```

For PGP messages, you [should][2] omit the `name` attribute:

```xml
    <encryption xmlns='urn:xmpp:eme:0'
                namespace='jabber:x:encrypted'/>
```

[1]: https://xmpp.org/extensions/xep-0380.html
[2]: https://xmpp.org/extensions/xep-0380.html#protocols

sdhand:
Fixes #334.

I'm not very used to the codebase yet, so this essentially duplicates code across both plugins, I'm sure there's a better way of doing this. Any pointers are welcome!

### DIFF
--- a/plugins/omemo/src/stream_module.vala
+++ b/plugins/omemo/src/stream_module.vala
@@ -11,6 +11,8 @@ private const string NODE_DEVICELIST = NS_URI + ".devicelist";
 private const string NODE_BUNDLES = NS_URI + ".bundles";
 private const string NODE_VERIFICATION = NS_URI + ".verification";
 
+private const string NS_URI_0380 = "urn:xmpp:eme:0";
+
 private const int NUM_KEYS_TO_PUBLISH = 100;
 
 public class StreamModule : XmppStreamModule {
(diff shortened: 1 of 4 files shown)


plugins/omemo/src/stream_module.vala
```diff
@@ -58,6 +60,10 @@ public class StreamModule : XmppStreamModule {
                     .put_node(new StanzaNode.build("payload", NS_URI)
                         .put_node(new StanzaNode.text(Base64.encode(ciphertext))));
 
+            StanzaNode encryption = new StanzaNode.build("encryption", NS_URI_0380).add_self_xmlns()
```
Mic92: I guess in theory one could write a function `encryption_stanza` that returns:

```vala
return new StanzaNode.build("encryption", NS_URI_0380)
                     .add_self_xmlns()
                     .put_attribute("namespace", NS_URI)
```

but the code that is duplicated here does not look to bad here.

sdhand:
As suggested by @fiaxh I've moved the implementation for xep-0380 into its own module. I'm not sure if the stream modules are the correct place to call this from however, although that is where the relevant namespaces are defined.

PR #366, opened 2018-05-22
Add new chat on Ctrl-T

Mic92:
This follows the behavior of other applications such as file managers or
browsers where Ctrl-T would open a new tab.


blastmaster:
I would really like to see this upstream. Starting a new conversation by two mouse clicks and searching a dialog is a lot of work for such a simple use case.

fiaxh:
This was done in 864196d.

PR #370, opened 2018-05-26
Enable secure_delete PRAGMA on SQLite databases

link2xt:
It is especially important for OMEMO database, as it stores *ephemeral* keys


fiaxh:
Merged in b9125b7e3e080caa2c5f233a22f62136322eb077.

PR #380, opened 2018-06-11
Fix a typo when setting the message state

sdhand:


sdhand:
Unsure as to why the first bug fix has been included in this (the commit hash seems to have changed) I don't think this should cause any issues when merging however so unless anyone has objections I'll leave it as is.

fiaxh:
Merged, thanks.

PR #382, opened 2018-06-16
Add an About-Dialog

Rahix:


main/CMakeLists.txt
```diff
@@ -167,3 +167,11 @@ install(FILES
 DESTINATION
     ${ICON_INSTALL_DIR}/hicolor/scalable/status
 )
+
+# Version
+execute_process(COMMAND git describe --tags OUTPUT_VARIABLE DINO_VERSION OUTPUT_STRIP_TRAILING_WHITESPACE)
```
jubalh: We build dino packages in OBS for various distros. There is not network connection in the build root. OBS prepares the git sources as a tarball before handling it to the buildroot. This will not work with it.

darix: Much simpler ... don't assume that you build from a git working copy.

cmake already has variables for version number and such. the about box should be filled from that. and packagers have to make sure it has a proper value.

Rahix: Hmm, yes I can see the issue. I can fix it for the version at least although `PROJECT_VERSION` is not set at the moment, I guess because there isn't really a version yet as Dino is still prerelease.

But a few lines below this, I also use git to get a list of contributors, which would definitly not be possible without git. Any ideas how that should be handled? Maybe a `CONTRIBUTORS` file? Or just hardcode a list of the main devs?

jubalh: I suppose that's for @fiaxh and @mar-v-in to decide

Rahix:
Commit bd45fdf1e16e ("Add version and adjust some cmake") seems to contain an implementation for an about dialog so this PR is now obsolete (The only thing which could still be relevant is the code for getting a list of contributors).

PR #418, opened 2018-08-20
Display room name in conversation selector and conversation titlebar

codedust:
Follow-up of #341.

This changes the display name of MUCs to the `room name`. If a `subject` is present, it is shown as a subtitle in the conversation titlebar. Otherwise, only the `room name` is shown.

Before:
image
image

After:
image
image

### DIFF
--- a/main/src/ui/conversation_titlebar/view.vala
+++ b/main/src/ui/conversation_titlebar/view.vala
@@ -35,35 +35,36 @@ public class ConversationTitlebar : Gtk.HeaderBar {
 
         stream_interactor.get_module(MucManager.IDENTITY).room_name_set.connect((account, jid, room_name) => {
             if (conversation != null && conversation.counterpart.equals_bare(jid) && conversation.account.equals(account)) {
-                update_title();
+                update_titlebar();
             }
         });
 
         stream_interactor.get_module(MucManager.IDENTITY).subject_set.connect((account, jid, subject) => {
             if (conversation != null && conversation.counterpart.equals_bare(jid) && conversation.account.equals(account)) {
-                update_subtitle();
+                update_titlebar();
             }
         });
     }
 
     public void initialize_for_conversation(Conversation conversation) {
         this.conversation = conversation;
-        update_title();
-        update_subtitle();
+        update_titlebar();
 
         foreach (Plugins.ConversationTitlebarWidget widget in widgets) {
             widget.set_conversation(conversation);
         }
     }
 
-    private void update_title() {
-        set_title(Util.get_conversation_display_name(stream_interactor, conversation));
-    }
+    private void update_titlebar() {
+        string conversation_display_name = Util.get_conversation_display_name(stream_interactor, conversation);
+        set_title(conversation_display_name);
 
-    private void update_subtitle() {
         if (conversation.type_ == Conversation.Type.GROUPCHAT) {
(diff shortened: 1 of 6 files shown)


main/src/ui/conversation_titlebar/view.vala
```diff
@@ -54,14 +60,10 @@ public class ConversationTitlebar : Gtk.HeaderBar {
...
     }
 
-    private void update_subtitle(string? subtitle = null) {
-        if (subtitle != null) {
-            set_subtitle(subtitle);
-        } else if (conversation.type_ == Conversation.Type.GROUPCHAT) {
-            string subject = stream_interactor.get_module(MucManager.IDENTITY).get_groupchat_subject(conversation.counterpart, conversation.account);
-            set_subtitle(subject != "" ? subject : null);
-        } else {
-            set_subtitle(null);
```
fiaxh: What's the reason for changing this function? With this change, if one switches from a MUC to a one-to-one chat, the one-to-one chat will still have the MUC topic set as a subtitle, because it's not cleared anymore.

codedust: My bad. I've fixed this issue in 7ad6011 and also implemented a fallback to the groupchat's subject/jid's localpart (which is what Conversation does).

codedust:
I'd like to justify and explain my second commit a bit because I think that's a good way to go. I've added the fallback to the room's subject in case no room name exists because that's a real thing and I'm actually in a MUC with no room name set. Therefore I've had a look at the logic in Conversations and adopted it's behavior: https://github.com/siacs/Conversations/blob/master/src/main/java/eu/siacs/conversations/entities/MucOptions.java#L379

In order to be usable by a large group of users, I personally think one of our top priorities should be to offer a clean UI that hides most of the technical stuff. Providing a fallback if no room name is set results in a cleaner UI for MUCs without a room name and results in a better user experience.

stevenroose:
I can confirm that it's common to identify rooms by their subject. I would say it's mostly an artifact that both exist, in most cases I've seen they were the same. Unless perhaps the subject was the name + some random long text added like with a website and so forth.

I'm in favor of showing the subject as a backup to the name. In the "ad-hoc group chat model", moving towards MIX, rooms get random names as they are formed. So the bare JID is very often not valuable at all.

I do agree though that it's a problem that we can't differentiate between open MUCs with an addressable JID and an ad-hoc group chat with a random JID. But I really think it's better to hide the random JID in favor of the subject.

stevenroose:
I run this now and it works great!

codedust:
@fiaxh Any news on this one?

stevenroose:
Curious too. I don't see a blocking issue.

stevenroose:
This PR appears to be an issue where it influences MUC nicks shown.

fiaxh:
For now I merged your first commit, because I think we all agree that it's a good thing to show the room name if one is available.

> I'm actually in multiple MUCs with no room name set and randomly generated jid's that are not easy to distinguish from each other

A (IMHO nice) solution would be to have the display name of a private, non-anonymous MUC without MUC name to be a list of occupant names. That would cover your randomly generated JID case.
Like `909f64a6-3d27-4286-8c67-fbfe2f075737@muc.example.com` -> `Anna, Max and Xavier`.

codedust:
Commit d8e0ab406cdad73edf79123e84d0013b96a3963a implements the behavior described by @fiaxh. However this behavior is different from how Conversations handles this (Conversations displays the subject when no room name is set). I suggest to adapt either dino's or Conversations' way of displaying the title of a MUC. cc @iNPUTmice

Issue #420, opened 2018-08-24
Searched text not reset when closing search


svalo:
Current behaviour:
* Open Dino
* Click the search button
* Type in text
* Conversation list gets filtered
* Click the search button again to close
* Conversation list is still filtered

What I would expect
* Open Dino
* Click the search button
* Type in text
* Conversation list gets filtered
* Click the search button again to close
* All the conversations are shown in the conversation list

svalo:
As the search button to filter the chatlist doesn't exist anymore I'm closing this :)

PR #421, opened 2018-08-24
Clears the search string when search button emits toggle signal

svalo:
As of now clicking the search button, typing in text and clicking again
leaves the chat filtered. With this edit once the search button is
toggled the search string gets resetted resulting in the full
conversation list being displayed.

This addresses #420

This is the first time I write vala code and one of the first I code for GUI SW, any feeedback is welcome :)

I'm not even sure this is the right approach, I tought that the signal would be the best one but the function is triggered also when first opening the search text input.

I noticed there is a [conversation_selected](https://github.com/dino/dino/blob/7def6d7ec2d330d6afe416ae148a8eeb25fe638b/main/src/ui/conversation_selector/view.vala#L24) function in view.vala but requires arguments and wasn't sure if and how I should use it

### DIFF
--- a/main/src/ui/conversation_selector/view.vala
+++ b/main/src/ui/conversation_selector/view.vala
@@ -25,6 +25,10 @@ public class View : Box {
         search_entry.set_text("");
     }
 
+    public void clear_search_string() {
+        search_entry.set_text("");
+    }
+
     private void refilter() {
         string[]? values = null;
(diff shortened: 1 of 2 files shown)


svalo:
Well, as the search in messages functionality was merged and the button to filter the chatlist doesn't exist anymore I'm closing this PR :)

PR #422, opened 2018-08-27
Fix compilation on FreeBSD

baitisj:
We should be able to rely on pkg-config for these libraries. Dependency issues should probably be addressed within the cmake/ subdirectory first; however, I am by no means a ```cmake``` expert.


baitisj:
It appears that I actually broke the project. Nevermind!

PR #432, opened 2018-09-27
Comment a return that stop every notifications to be trigger when an account is not synced

PapsOu:


PapsOu:
I've changed the fix to a better one (that should not break notifications with MAM available).

I don't have the possibility to test if it works with a MAM compatible account, so please let me know if this break something ?

PapsOu:
OK I've tested with two accounts freshly created on https://jabberfr.org/ and notifications seems to work too when MAM is supported.

You can merge this PR if it seems OK for you

fiaxh:
Thanks!

PR #456, opened 2018-11-16
Add escape for multiline codeblocks

Rahix:
Allow formatting code-blocks using  ```` ``` ````, eg:

    ```
    def foo():
        bar()
    ```

### DIFF
--- a/main/src/ui/util/helper.vala
+++ b/main/src/ui/util/helper.vala
@@ -181,11 +181,11 @@ public static string parse_add_markup(string s_, string? highlight_word, bool pa
     }
 
     if (parse_text_markup) {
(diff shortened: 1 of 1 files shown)


fiaxh:
This allows line breaks in between any formatting, so
```
*bla
meow*
```
would be all-bold. While Dino's formating isn't yet fully compliant with [XEP-0393: Message Styling
](https://xmpp.org/extensions/xep-0393.html) we'd like to go into that direction. [Section 5.1.1](https://xmpp.org/extensions/xep-0393.html#line-block) talks about spans (`*bold*`, `_italic_`, ...) not going across multiple lines.

Issue #445, opened 2018-11-06
Markdown-based bold and italics don't display as (likely) intended


LuccoJ:
Dino version is the Fedora build marked 0.0~git318.ddd17e7 release 116.1.

The behavior is visible in the attached screenshot:

![weirdmarkdown](https://user-images.githubusercontent.com/1684791/48086928-4d529280-e1fe-11e8-9421-ce53cdf5acf5.png)

If I paste the same sentences here,

> I think *bolding* works in a bit of a *weird* way...
> Let's see if _italics_ do the _same_ thing first

the formatting (aside from bold not being a thing in the first place) is handled the way I'd expect it to, unlike in Dino.

fiaxh:
Fixed in 5a4e509.

PR #457, opened 2018-11-17
Redo markup parsing

r3pek:
Besides removing the recursion that was needed, we now handle a series of corner cases
and better parsing of the ** __ and  tags. This should, hopefully, cover most the
use cases used in chat

fixes bug #445

### DIFF
--- a/main/src/ui/util/helper.vala
+++ b/main/src/ui/util/helper.vala
@@ -198,8 +198,8 @@ public static string parse_add_markup(string s_, string? highlight_word, bool pa
     }
 
     if (parse_text_markup) {
-        string[] markup_string = new string[]{"`", "_", "*"};
-        string[] convenience_tag = new string[]{"tt", "i", "b"};
+        string[] markup_string = new string[]{"`", "_", "*", "~"};
+        string[] convenience_tag = new string[]{"tt", "i", "b", "s"};
 
         for (int i = 0; i < markup_string.length; i++) {
             Regex regex = new Regex("(?=" + Regex.escape_string(markup_string[i]) + ")" +
--- a/main/src/ui/util/helper.vala
+++ b/main/src/ui/util/helper.vala
@@ -202,7 +202,9 @@ public static string parse_add_markup(string s_, string? highlight_word, bool pa
         string[] convenience_tag = new string[]{"tt", "i", "b"};
 
         for (int i = 0; i < markup_string.length; i++) {
-            Regex regex = new Regex("(?=" + Regex.escape_string(markup_string[i]) + ").+?" + Regex.escape_string(markup_string[i]));
+            Regex regex = new Regex("(?=" + Regex.escape_string(markup_string[i]) + ")" +
+                                """(?!.*<\/?[""" + string.joinv("", convenience_tag) + """]+>.*?""" + Regex.escape_string(markup_string[i]) + ")" + /* Don't match something that already has a tag in the middle */
+                                ".+?" + Regex.escape_string(markup_string[i]));
 
             s = regex.replace_eval(s, -1, 0, 0, (mi, s) => {
                 int start, end;
--- a/main/src/ui/util/helper.vala
+++ b/main/src/ui/util/helper.vala
@@ -144,40 +144,57 @@ public static bool is_24h_format() {
     return settings_format == "24h" || p_format == " ";
 }
 
-public static string parse_add_markup(string s_, string? highlight_word, bool parse_links, bool parse_text_markup, bool already_escaped_ = false) {
-    string s = s_;
-    bool already_escaped = already_escaped_;
+public static string parse_add_markup(string s_, string? highlight_word, bool parse_links, bool parse_text_markup) {
+    Gee.ArrayList<int> start_url_pos = new Gee.ArrayList<int>();
+    Gee.ArrayList<int> end_url_pos = new Gee.ArrayList<int>();
+    string s = Markup.escape_text(s_);
 
     if (parse_links) {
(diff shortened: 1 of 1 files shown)


main/src/ui/util/helper.vala
```diff
@@ -144,59 +144,87 @@ public static bool is_24h_format() {
...
+    Gee.ArrayList<int> start_url_pos = new Gee.ArrayList<int>();
+    Gee.ArrayList<int> end_url_pos = new Gee.ArrayList<int>();
+    string s = Markup.escape_text(s_);
 
     if (parse_links) {
-        Regex url_regex = new Regex("""(?i)\b((?:[a-z][\w-]+:(?:\/{1,3}|[a-z0-9%])|www\d{0,3}[.]|[a-z0-9.\-]+[.][a-z]{2,4}\/)(?:[^\s()<>]+|\(([^\s()<>]+|(\([^\s()<>]+\)))*\))+(?:\(([^\s()<>]+|(\([^\s()<>]+\)))*\)|[^\s`!()\[\]{};:'".,<>?«»“”‘’]))""");
-        MatchInfo match_info;
-        url_regex.match(s.down(), 0, out match_info);
-        if (match_info.matches())  {
+        Regex url_regex = new Regex("""(?i)((https?|ftp|magnet|tel)+):(?:\/\/(?:((?:[a-z0-9-._~!$&'()*+,;=:]|%[0-9A-F]{2})*)@)?((?:[a-z0-9-._~!$'()*+,;=]|%[0-9A-F]{2})*)(?::(\d*))?(\/(?:[a-z0-9-._~!$&'()+,;=:@/]|%[0-9A-F]{2})*)?|(\/?(?:[a-z0-9-._~!$&'()*+,;=:@]|%[0-9A-F]{2})+(?:[a-z0-9-._~!$&'()*+,;=:@/]|%[0-9A-F]{2})*)?)(?:\?((?:[a-z0-9-._~!$&'()*+,;=:/?@]|%[0-9A-F]{2})*))?(?:#((?:[a-z0-9-._~!$&'()*+,;=:/?@]|%[0-9A-F]{2})*))?""");
```
stevenroose: Can't this be generalized? I mean there are tons of other URI schemes.. You forgot `mailto:` f.e. Another one I personally use a lot is `bitcoin:`. Does the RFC not provide a general one? Found one: https://tools.ietf.org/html/rfc3986#appendix-B.

main/src/ui/util/helper.vala
```diff
@@ -144,59 +144,87 @@ public static bool is_24h_format() {
...
+            }
+            s.append_printf("<b>%s</b>", mi.fetch(0));
+            return false;
+        });
     }
 
     if (parse_text_markup) {
-        string[] markup_string = new string[]{"`", "_", "*"};
-        string[] convenience_tag = new string[]{"tt", "i", "b"};
+        string[] markup_string = new string[]{"`", "_", "*", "~"};
```
stevenroose: So this PR also adds strikethrough support?

main/src/ui/util/helper.vala
```diff
@@ -144,59 +144,87 @@ public static bool is_24h_format() {
...
+        string[] convenience_tag = new string[]{"tt", "i", "b", "s"};
 
         for (int i = 0; i < markup_string.length; i++) {
-            Regex regex = new Regex(Regex.escape_string(markup_string[i]) + ".+" + Regex.escape_string(markup_string[i]));
-            MatchInfo match_info;
-            regex.match(s.down(), 0, out match_info);
-            if (match_info.matches()) {
+            Regex regex = new Regex("(?=" + Regex.escape_string(markup_string[i]) + ")" +
+                                """(?!.*<\/?[""" + string.joinv("", convenience_tag) + """]+>.*?""" + Regex.escape_string(markup_string[i]) + ")" + /* Don't match something that already has a tag in the middle */
```
stevenroose: This comment seems oddly-placed..

main/src/ui/util/helper.vala
r3pek: @stevenroose I tried not to match the whole world but it's as simple as adding it at the begining (as you noticed).
It's also hard to generalize it while maintaining a decent match to actual URLs. (the IETF one doesn't work)

main/src/ui/util/helper.vala
r3pek: Yeah, I did it after the initial pull request. Maybe I need to update the PR comment.

main/src/ui/util/helper.vala
r3pek: It's not. It explains *exactly* that line of code.

stevenroose:
Yeah but given that most people have 80- or 100-column text editors, it's
better to split in multiple lines I guess. For readability at least.
Doesn't Vala allow for adding comments in between the additions?

iNPUTmice:
Hi,

this is already less prone to bugs than the current version :-)

However I noticed one thing. According to the XEP formatting inside a preformatted span should be ignored. See [section 5.2.5](https://xmpp.org/extensions/xep-0393.html#mono). Look at the examples (easier than me just trying to convey that on github :) )

fiaxh:
For now, #445 was fixed in 5a4e509. It follows the `Span` definition of [XEP-0393: Message Styling (5.2)
](https://xmpp.org/extensions/xep-0393.html#span) which will already solve a lot of problems. We would like to implement XEP-0393 more closely in the future, possibly by parsing messages into an intermediate data structure that allows code sharing with an implementation of [XEP-0394: Message Markup](https://xmpp.org/extensions/xep-0394.html). Either way it will probably be largely a rewrite of the current code.

PR #458, opened 2018-11-18
Allow 12 for the size of the IV

Syndace:
The IV for AES-GCM is recommended to be 12 bytes in size. Limiting it to 16 bytes will create incompatibilities with clients following the specs.


licaon-kter:
Now open one PR for Conversations, one for Converse, one for ChatSecure... :)

iNPUTmice:
@licaon-kter neither Conversations nor Converse break when receiving it. It is not about what we send but about what we parse

licaon-kter:
Yeah whatever, I still remember the CS breakage :P

PR #464, opened 2018-11-27
display groups w/o name by list of members

iNPUTmice:
For members-only, non-anonymous groups that do not have a name set we now
show an automatically generated name that is generated from the list of members
(offline, and online). Only the first name (before the first space) is used
to keep the generated name short.

This commit also uses the offline members list instead of the online member list
to generate avatar tiles (also only in members-only, non-anon groups.)

### DIFF
--- a/libdino/src/service/muc_manager.vala
+++ b/libdino/src/service/muc_manager.vala
@@ -13,6 +13,7 @@ public class MucManager : StreamInteractionModule, Object {
     public signal void left(Account account, Jid jid);
     public signal void subject_set(Account account, Jid jid, string? subject);
     public signal void room_name_set(Account account, Jid jid, string? room_name);
+    public signal void private_room_occupant_updated(Account account, Jid room, Jid occupant);
     public signal void bookmarks_updated(Account account, Gee.List<Xep.Bookmarks.Conference> conferences);
 
     private StreamInteractor stream_interactor;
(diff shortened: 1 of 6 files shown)


stevenroose:
tested 838334d21c2111c7cb90c755aeefab85d9680b14

Works fine. New group chats (created in Conversations) don't show up automatically though, but after a restart of Dino, they showed up and had the tiled avatars.

libdino/src/service/muc_manager.vala
```diff
@@ -13,6 +13,7 @@ public class MucManager : StreamInteractionModule, Object {
     public signal void left(Account account, Jid jid);
     public signal void subject_set(Account account, Jid jid, string? subject);
     public signal void room_name_set(Account account, Jid jid, string? room_name);
+    public signal void members_only_non_anonymous_occupant_updated(Account account, Jid room, Jid occupant);
```
mar-v-in: Can we come up with a better name for this? `members_only_non_anonymous` is incredibly long and unreadable, although technically correct. Maybe using a new term like `private room` would be more suitable?

libdino/src/service/muc_manager.vala
```diff
@@ -97,6 +98,18 @@ public class MucManager : StreamInteractionModule, Object {
         return false;
     }
 
+    public bool members_only_non_anonymous(Account account, Jid jid) {
```
mar-v-in: This should start with `is_` to make obvious that this is not an action.

libdino/src/service/muc_manager.vala
stevenroose: "room" might also be an outdated term.. I think groupchat is more common nowadays. As long as it's not "barbecue".

PR #486, opened 2018-12-30
Add some missing translatable strings

srgcdev:


fiaxh:
I usually update the `.po` files with the strings from the code, weblate with the strings and the code with the updated translations from weblate all at the same time. The updates to this repo are in 4c8f4be, and new strings (among others, the one which you added) are on weblate now.

Whenever you feel the translations weren't updated for too long, please ping me.

srgcdev:
I thought the main .pot file was made by hand.
Thanks for your reply and for the update.

PR #489, opened 2019-01-05
Set translations in vala files to fix problem with missing translations…

srgcdev:
… in ui files for plugins (omemo).
(Temporary or perhaps definitive solution)


fiaxh:
Thanks. I moved some things around and committed it here: 8e6db88.

PR #502, opened 2019-01-19
Clean up ./configure script and fix most ShellCheck warnings

kalkin:
After cleaning up the `./configure` Script I noticed that the script started warning me of missing sqlite and few other libraries. This didn't happen before. Before `make` would fail and printout that it can't find libraries in `/usr/inclide/`.


kalkin:
Bump?

Issue #482, opened 2018-12-28
Support for GSSAPI SASL authentication


jelmer:
It would be great if Dino supported GSSAPI authentication. 

There seem to be two ways of doing this:

* Rely on a SASL library that already supports GSSAPI (such as gsasl, which already has vapi files) instead of a custom SASL implementation
* add support to the existing SASL code, linking against libgssapi

jelmer:
Working on this in https://github.com/jelmer/dino/tree/gsasl-hostname

PR #517, opened 2019-01-30
Use libgsasl for SASL operations

jelmer:
Use gsasl for SASL operations.
    
This simplifies the SASL module somewhat, and brings in support for
more SASL mechanisms that are supported by GSASL:
https://www.gnu.org/software/gsasl/

Tested against gmail and prosody with PLAIN.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,7 @@ Build
 * libgee-0.8 (≥ 0.10)
 * libgcrypt (For the OMEMO plugin)
 * libqrencode3 (For the OMEMO plugin)
+* libgsasl (For SASL authentication)
 * libsoup (For the HTTP files plugin)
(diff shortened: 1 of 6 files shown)


fiaxh:
Replacing our current SASL implementation with gsasl would introduce a (not widely used) dependency without giving any benefits at this point.

We will leave this PR open for now so that people can see it and build on top of it if they want to.

jelmer:
This is a necessary change for adding GSSAPI/Kerberos authentication support (#482). I have a follow-up branch ready that adds GSSAPI support, and which I have tested with prosody but should also work with the GSSAPI support in e.g. ejabberd.

jelmer:
Closing this in favor of #518.

PR #518, opened 2019-02-08
Provide hostname to GSASL, enabling GSSAPI authentication

### LINKED ISSUE #482: Support for GSSAPI SASL authentication

jelmer:
This pull request is a follow-up to https://github.com/dino/dino/pull/517, and makes Dino provide the FQDN of the server it is talking to to GSASL.

This enables GSSAPI (Kerberos) authentication, and fixes #482. I've been using this with prosody for a while.

### DIFF
--- a/xmpp-vala/src/core/xmpp_stream.vala
+++ b/xmpp-vala/src/core/xmpp_stream.vala
@@ -14,6 +14,7 @@ public class XmppStream {
     public const string NS_URI = "http://etherx.jabber.org/streams";
 
     public Jid remote_name;
+    public string remote_hostname;
     public XmppLog log = new XmppLog();
     public StanzaNode? features { get; private set; default = new StanzaNode.build("features", NS_URI); }
 
@@ -64,11 +65,16 @@ public class XmppStream {
             }
             IOStream? stream = null;
             if (best_provider != null) {
-                stream = yield best_provider.connect(this);
+                stream = yield best_provider.connect(this, out remote_hostname);
             }
             if (stream == null) {
                 debug("Connecting to %s, xmpp-client, tcp (fallback)", this.remote_name.to_string());
                 stream = yield (new SocketClient()).connect_to_host_async(this.remote_name.to_string(), 5222);
+                GLibFixes.Resolver resolver = GLibFixes.Resolver.get_default();
+                GLib.List<SrvTarget>? xmpp_target = yield resolver.lookup_service_async("xmpp-client", "tcp", this.remote_name.to_string(), null);
+                SrvTarget? srv_target = xmpp_target.nth(0).data;
+                remote_hostname = srv_target.get_hostname();
+                stream = yield (new SocketClient()).connect_to_host_async(srv_target.get_hostname(), srv_target.get_port());
             }
             if (stream == null) {
                 throw new IOStreamError.CONNECT("client.connect() returned null");
@@ -380,7 +386,7 @@ public abstract class XmppStreamNegotiationModule : XmppStreamModule {
 
 public abstract class ConnectionProvider {
     public async abstract int? get_priority(Jid remote_name);
-    public async abstract IOStream? connect(XmppStream stream);
+    public async abstract IOStream? connect(XmppStream stream, out string hostname);
     public abstract string get_id();
 }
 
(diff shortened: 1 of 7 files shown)


xmpp-vala/src/core/xmpp_stream.vala
```diff
@@ -57,10 +58,14 @@ public class XmppStream {
             }
             IOStream? stream = null;
             if (best_provider != null) {
-                stream = yield best_provider.connect(this);
+                stream = yield best_provider.connect(this, out remote_hostname);
             }
             if (stream == null) {
-                stream = yield (new SocketClient()).connect_async(new NetworkService("xmpp-client", "tcp", this.remote_name.to_string()));
+                GLibFixes.Resolver resolver = GLibFixes.Resolver.get_default();
```
mar-v-in: This originally was the fallback case for when there is no SRV record. I guess your change will break the fallback and make SRV records required.

xmpp-vala/src/module/sasl.vala
```diff
@@ -162,60 +74,34 @@ namespace Xmpp.Sasl {
...
-                                    .put_attribute("mechanism", Mechanism.PLAIN)
-                                    .put_node(new StanzaNode.text(Base64.encode(get_plain_bytes(name, password)))));
-                var flag = new Flag();
-                flag.mechanism = Mechanism.PLAIN;
-                flag.name = name;
-                stream.add_flag(flag);
-            } else {
-                stderr.printf("No supported mechanism provided by server at %s\n", stream.remote_name.to_string());
+            flag.sasl_session.set_property(Gsasl.Property.AUTHID, name);
+            flag.sasl_session.set_property(Gsasl.Property.HOSTNAME, stream.remote_hostname);
```
mar-v-in: According to XEP 233, the hostname for GSSAPI is provided with the SASL mechanisms, see https://xmpp.org/extensions/xep-0233.html#examples
Can we use this instead of passing the hostname around everywhere?

xmpp-vala/src/module/sasl.vala
```diff
@@ -44,110 +39,27 @@ namespace Xmpp.Sasl {
...
-                        }
-                    }
-                    stream.require_setup();
-                    flag.password = null; // Remove password from memory
-                    flag.finished = true;
-                } else if (node.name == "failure") {
+            if (node.ns_uri != NS_URI)
+                return;
+            if (node.name == "success") {
+                Flag flag = stream.get_flag(Flag.IDENTITY);
```
mar-v-in: I think with the new logic, the server signature (that is sent with the success nonza) is not verified any more (as described in points 10 and 11 in https://wiki.xmpp.org/web/SASL_and_SCRAM-SHA-1). Can you check how to verify it through gsasl?

mar-v-in (commented):
I gave some comments inline.

I generally would love if we had a way to make the libgsasl dependency optional through a plugin, but it is probably to much work right now (especially as we want to store the scram salted password instead of cleartext password longterm). We can still split it out in a plugin later on, after we defined some proper APIs for that.

xmpp-vala/src/core/xmpp_stream.vala
jelmer: My understanding is that the original code would only work with SRV as well, based on reading the documentation for NetworkService: https://valadoc.org/gio-2.0/GLib.NetworkService.html

Wouldn't this need to use NetworkAddress rather than NetworkService to use a plain hostname rather than a SRV record?

xmpp-vala/src/module/sasl.vala
jelmer: I've updated this to use the hostname from the SASL mechanism, and reverted the passing around of hostnames everywhere.

xmpp-vala/src/core/xmpp_stream.vala
jelmer: I'm marking this as resolved since I've reverted this part of the code - whether this is doing the right thing is orthogonal.

xmpp-vala/src/module/sasl.vala
jelmer: Done.

jelmer:
Please take another look, I believe I have addressed all review comments.

Neustradamus:
@mar-v-in: Any news for this PR?

Issue #473, opened 2018-12-12
OMEMO messages disappearing into the ether


Chobbes:
I seem to not be receiving OMEMO messages from somebody on gajim. Plain text messages come through, but OMEMO does not.

I believe if I can't decrypt the message or whatever I should get "This message is OMEMO encrypted" or some message to that effect, but I'm not getting anything... And all the keys are accepted, so it should in principle be fine?

Is there some way I can debug this? Can I get dino to print out debug information and see if I'm actually getting the messages and it's not displaying it, or something?

NicoHood:
Maybe related: https://github.com/dino/dino/issues/243

srgcdev:
Same problem here. I cannot receive omemo messages from gajim despite all the omemo keys are accepted.

srgcdev:
It seems that gajim doesn't send a <body> so when message_processor.vala file executes 'if (message_stanza.body == null) return;' the message is lost.

mar-v-in:
XEP-380 says:

> Entities SHOULD include a non-encrypted body as possible, since older clients not supporting this protocol might otherwise ignore messages sent with an unknown encryption, making both the sender frustrated that their message did not get an answer, and the recipient frustrated that they never saw any message.

moppman:
> XEP-380 says:
> 
> > Entities SHOULD include a non-encrypted body as possible, since **older clients** not supporting this protocol might otherwise ignore messages sent with an unknown encryption [...]

> **older clients**

which clearly Dino is not :)

jjrh:
Regardless of who's at fault, silently failing is a bug. The user should be informed they aren't seeing a message.

PR #519, opened 2019-02-09
Do not ignore encrypted messages without a body tag

ThibG:
Partial fix to #473.

Partial in that it should handle correctly-encrypted incoming messages without a body tag, but won't display anything if such a message cannot be decrypted.

### DIFF
--- a/plugins/omemo/src/trust_manager.vala
+++ b/plugins/omemo/src/trust_manager.vala
@@ -247,6 +247,7 @@ public class TrustManager {
             StanzaNode? _encrypted = stanza.stanza.get_subnode("encrypted", NS_URI);
             if (_encrypted == null || MessageFlag.get_flag(stanza) != null || stanza.from == null) return false;
             StanzaNode encrypted = (!)_encrypted;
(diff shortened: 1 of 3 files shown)


plugins/omemo/src/trust_manager.vala
```diff
@@ -247,6 +247,7 @@ public class TrustManager {
             StanzaNode? _encrypted = stanza.stanza.get_subnode("encrypted", NS_URI);
             if (_encrypted == null || MessageFlag.get_flag(stanza) != null || stanza.from == null) return false;
             StanzaNode encrypted = (!)_encrypted;
+            if (message.body == null) message.body = "[This message is OMEMO encrypted]";
```
stevenroose: Can't this hard-coded string be generalized with the other string that Dino *sets* for outgoing OMEMO messages using a global? Or using i14n if that's already available.

fiaxh:
Thanks, merged.

This PR introduced problems in `ChatInteraction.ReceivedMessageListener` (wrongly marking conversations as unread) so I also pushed a fix for that (d27fb96).

PR #529, opened 2019-03-08
OS X Fixes

kousu:
This fixes the

> Undefined symbols for architecture x86_64:
>  _gpg_strerror, referenced from:
>      _gpgme_throw_if_error in libgpgme-vala.a(gpgme_helper.c.o)
> ld: symbol(s) not found for architecture x86_64
> clang: error: linker command failed with exit code 1 (use -v to see invocation)

error on that's showing on OS X.

### DIFF
--- a/plugins/gpgme-vala/vapi/gpgme.vapi
+++ b/plugins/gpgme-vala/vapi/gpgme.vapi
@@ -505,11 +505,13 @@ namespace GPG {
     [CCode (cname = "gpgme_get_engine_information")]
     public GPGError.Error get_engine_information(out EngineInfo engine_info);
 
-    [CCode (cname = "gpgme_strerror_r")]
-    public int strerror_r(GPGError.Error err, uint8[] buf);
+    public int strerror_r(GPGError.Error err, uint8[] buf) {
+        return err.strerror_r(buf);
+    }
 
-    [CCode (cname = "gpgme_strerror")]
-    public unowned string strerror(GPGError.Error err);
+    public unowned string strerror(GPGError.Error err) {
+        return err.to_string();
+    }
 
     private void throw_if_error(GPGError.Error error) throws GLib.Error {
         if (error.code != GPGError.ErrorCode.NO_ERROR) {
--- a/plugins/gpgme-vala/CMakeLists.txt
+++ b/plugins/gpgme-vala/CMakeLists.txt
@@ -47,6 +47,6 @@ set(CFLAGS ${VALA_CFLAGS} -I${CMAKE_CURRENT_SOURCE_DIR}/src)
 add_definitions(${CFLAGS})
 add_library(gpgme-vala ${GPGME_VALA_C} src/gpgme_fix.c)
 add_dependencies(gpgme-vala gpgme-vapi)
-target_link_libraries(gpgme-vala ${GPGME_VALA_PACKAGES} gpgme)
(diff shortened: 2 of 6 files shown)


kousu:
I don't know cmake that well, so reviews very appreciated. I am confident in how I added libgpg-error to the CMakeLists.txt, but not in the .so -> .dylib hack; for one thing, it is not integrated with `make install`; for another, the real fix should be to convince Dino to load .dylibs instead of this weird renaming, probably, right?

kousu:
I just updated brew and discovered this needs some updates. Forthcoming!

kousu:
So I think the gpgme and gettext detection in CMakeLists.txt has rotted somehow, but it turns out that I could fix it (on my system at least) with

```
echo 'export CPATH="/usr/local/include:$CPATH"' >> ~/.bash_profile
echo 'export LDFLAGS="-L/usr/local/lib"' >> ~/.bash_profile
rm -r build Makefile
```

After that `./configure && make` worked again.

I've added those lines to [the wiki](https://github.com/dino/dino/wiki/macOS).

kousu:
I've sync'd this with `master` again!

Maybe what I should do, as well, is add a patch to use Travis (or Github Actions) to test it on OS X.

plugins/gpgme-vala/vapi/gpgme.vapi
```diff
@@ -505,11 +505,13 @@ namespace GPG {
...
-    public int strerror_r(GPGError.Error err, uint8[] buf);
+    public int strerror_r(GPGError.Error err, uint8[] buf) {
+        return err.strerror_r(buf);
+    }
 
-    [CCode (cname = "gpgme_strerror")]
-    public unowned string strerror(GPGError.Error err);
+    public unowned string strerror(GPGError.Error err) {
+        return err.to_string();
+    }
```
kousu, line 514: tbh I forget why I did this.  I assume there was a good reason I needed to flesh out these functions instead of just leaving them as with that `CCode` decorator.

fiaxh:
If the changes you do to gpg-error are necessary for you, your build process might not be optimal. @mar-v-in wrote a [tutorial](https://github.com/dino/dino/wiki/macOS) on how to build Dino on MacOS a few months back. With those build instructions, no changes to gpg-error are needed.

As for the .dylib v.s. .so issue: glib expects the suffix of module libraries to be .so for legacy reasons, however CMake produces .dylib libraries. It seems that won't be fixed in glib anymore for the current modules API version, so we'll have to work-around it.
The [wiki build instructions](https://github.com/dino/dino/wiki/macOS) just rename .dylib to .so. However it would indeed be nicer to have CMake handle this. You create a symlink, however I think it would be nicer to just have CMake create the correct file. This should be possible with something along the lines of `set_target_properties(http-files PROPERTIES SUFFIX ".so")`.

kousu:
Thank you for looking @fiaxh ! I wrote this a year ago by following https://github.com/dino/dino/wiki/macOS. At the time this is what I found was necessary, but it's very possible the glitches were solved in the meantime. I'll work on re-testing and integrating these suggestions the next time I boot my mac VM.

fiaxh:
I'm closing this for now. If you ever feel like changing the cmake file as mentioned above, please reply here or open a new PR.

kousu:
Ai, yeah, sorry, I haven't had a macOS machine for a few years now so this has been low on my priorities unfortunately. Maybe some day. If one of my mac friends ever wants to try Dino out I'll point them back here.

PR #535, opened 2019-03-20
Fixed not falling back to other SRV services when connection fails

### LINKED ISSUE #534: Try all SRV records in priority order, instead of just the first one
Right now, dino fetches `xmpp-client` and `xmpps-client` records and tries to connect to the more prioritary (the one with the smallest priority number), failing if it is not reachable.

This is not the expected behavior according to [XMPP-Core, 3.2.1](https://tools.ietf.org/html/rfc6120#page-18):

>    7.  If the initiating entity fails to connect using all resolved IP
       addresses for a given FDQN, then it repeats the process of
       resolution and connection for the next FQDN returned by the SRV
       lookup based on the priority and weight as defined in [DNS-SRV].

This could be useful to expose a regular xmpp port as the most prioritary option, and 443 as a fallback for restricted networks.

roobre:
Fixes #534 

Okay, so I've tried to tackle this issue and here's the result. I've changed a few things here and there because the previous model, with a `ConnectionProvider` per SRV service type was not compatible with iterating between both acceptable services (`xmpp-client` and `xmpps-client`) in order of priority. So I've created a new class, called `ServiceLookuper`, which is responsible of looking up SRV records and returning a list of `ConnectionProvider`s, which would in turn be responsible of instantiating the connection itself. `XmppStream` sorts connection providers based on priority, and tries them in order.

From the OOP perspective, a few things could be improved (including reusing some of the code between the `ConnectionProvider` and `ServiceLookuper` implementations, which is very very similar.

A few important things to note with this:

* I have no idea of what I'm doing. Literally, these are the first lines of Vala I've ever written. There are probably lots of stuff which would benefit from some polish, and maybe I left a lot of references without updating so please tell me if that's the case.

* Timeouts. As far as I know, some timeouts, like the one in the "Add account" dialog, are implemented in the libdino layer. This means that, even if I have set a timeout of 7 seconds per connection attempt, libdino's timeout may kick in earlier, killing the whole thing before the next connection can be even tried. This could be improved but I'm not sure how, so I'm open for feedback. For now, as a workaround, you can add a firewal rule to reject connections instead of silently dropping them, and hopefully that would give the stream enough time to try several ports before the timeout kicks in.

And I think that's it, feedback is welcome.

### DIFF
--- a/xmpp-vala/src/core/xmpp_stream.vala
+++ b/xmpp-vala/src/core/xmpp_stream.vala
@@ -23,7 +23,7 @@ public class XmppStream {
 
     public Gee.List<XmppStreamFlag> flags { get; private set; default=new ArrayList<XmppStreamFlag>(); }
     public Gee.List<XmppStreamModule> modules { get; private set; default=new ArrayList<XmppStreamModule>(); }
-    private Gee.List<ConnectionProvider> connection_providers = new ArrayList<ConnectionProvider>();
+    private Gee.List<ServiceLookuper> service_lookupers = new ArrayList<ServiceLookuper>();
     public bool negotiation_complete { get; set; default=false; }
     private bool setup_needed = false;
     private bool non_negotiation_modules_attached = false;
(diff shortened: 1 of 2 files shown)


genofire:
@fiaxh any reason why this is not merged?

fiaxh:
We've been planing to do a larger rewrite the `XmppStream` code for some while. The idea is to have an `StartTlsXmppStream` and an `DirectTlsXmppStream` which connect to a specific host+port. The connection option detection and ranking logic is done outside of the stream. This approach has the advantage that we can more flexibly add further connection methods and define the order in which they are tried.

In case you want to have a look, I pushed the current WIP code state to the [`refactor_xmpp_stream`](https://github.com/dino/dino/tree/fiaxh/refactor_xmpp_stream) branch. It still needs some testing and I won't push it to master before the 0.2 release. But once it's pushed it should (among others) solve the issue that this PR is about.

fiaxh:
I commited the refactoring mentioned in my previous message in 07917f1d841f449157aa3aaa2507b0547dd274e7. Among others, Dino should now try all SRV records in order. It would be great if you could verify that it actually works for you.

roobre:
Awesome, thanks!

Issue #145, opened 2017-09-06
Add keyboard navigation


sandwm:
I would like to add some keyboard navigation options to dino. This is the plan (^A means &lt;Primary&gt;A or Ctrl+A):

- [ ] Menu entries
  - Application menu
    - ^A Accounts
    - ^P Settings
    - ^Q Quit
  - Add menu
    - ^N Start Chat
    - ^M Join Conference
  - ^F Find
  - ^O Open / attach
  - ^L List occupants
  - Alt+E Encryption menu
  - "Hamburger menu"
    - Alt+Enter Contact Details
- [ ] Others
  - ^PageUp / ^PageDown and ^Tab / ^Shift+Tab to switch conversations (#142)
  - Alt+1, Alt+2, ... to switch to first, second, ... conversation (counted from the bottom / the oldest to make switching more consistent if new conversations are added)
  - PageUp / PageDown to scroll (#143)
  - ^W to close the current conversation
- [x] Just for reference:
  - F10 to select menu buttons, Tab / Shift+Tab to cycle through the buttons
  - unknown: select application menu

I will implement it in the next days. Feel free to comment.

(As a side note: It should be possible to see the keys in the menu popups if this feature request gets implemented: https://bugzilla.gnome.org/show_bug.cgi?id=787342)

d9h20f:
GNOME applications have a 'Keyboard Shortcuts' screen you can get to by clicking on the application name in the top bar (where Dino currently has 'Accounts' and 'Settings'). You can easily describe all shortcuts in that window, so I don't think your feature request will be implemented.

Either way, the shortcuts look good to me.

bowlofeggs:
Another shortcut recommendation: I like to use Alt+a in weechat, hexchat, finch, and irssi to switch to the next buffer with notifications, prioritizing buffers that mention me or are PMs over buffers that have unread messages that don't mention my nick.

Also, there are two other similar tickets: #400 and #142.

RX14:
ctrl+pgup/pgdn seems to be the standard in most tab-based applications to move between tabs. I'm fine with leaving the current shortcuts in, but having a more standard option as well seems like a good idea for discoverability.

haasn:
I'd personally expect `ctrl+p` and `ctrl+n` to cycle through conversations. But I am also a heavy vim user.

Given that we all have different preferences, I think it should be configurable.

PR #536, opened 2019-03-21
Allow to cycle active conversations with Ctrl+(Shift)+Tab

### LINKED ISSUE #141: Quitting dino does not close XMPP session
When you quit dino, it will not send an "unavailable" presence, and thus leave the XMPP session in a semi-dead state. That means, among other things, that messages sent to you will end up in the nirvana, and that your zombie will stay on MUCs for some time.

### LINKED ISSUE #142: Keyboard navigation between conversations
It would be quite useful to have ctrl-pageup and ctrl-pagedown move up and down between conversations.

### LINKED ISSUE #400: switch conversations using the keyboard
How can conversations be switched using the keyboard? irssi has Esc + Number, vim has gt/gT to switch tabs. It would be convenient not having to rely on the mouse for this.

roobre:
Closes #400, Closes #142, related to #145

### DIFF
--- a/main/src/ui/unified_window_controller.vala
+++ b/main/src/ui/unified_window_controller.vala
@@ -142,6 +142,15 @@ public class UnifiedWindowController : Object {
         }
     }
 
+    public void loop_conversations(bool backwards = false) {
+        Gee.List<Conversation> conversations = stream_interactor.get_module(ConversationManager.IDENTITY).get_active_conversations();
+        conversations.sort((a, b) => { return b.last_active.compare(a.last_active); });
+        int index = conversations.index_of(this.conversation);
+        index += backwards ? -1 : 1;
+        select_conversation(conversations.get(index % conversations.size));
+    }
+
+
     private void update_conversation_display_name() {
         conversation_display_name = Util.get_conversation_display_name(stream_interactor, conversation);
     }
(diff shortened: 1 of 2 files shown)


roobre:
(Sorry for the force-push spam, I referenced 141 instead of 142. Ouch.

main/src/ui/unified_window_controller.vala
```diff
@@ -142,6 +142,15 @@ public class UnifiedWindowController : Object {
         }
     }
 
+    public void loop_conversations(bool backwards = false) {
+        Gee.List<Conversation> conversations = stream_interactor.get_module(ConversationManager.IDENTITY).get_active_conversations();
+        conversations.sort((a, b) => { return b.last_active.compare(a.last_active); });
+        int index = conversations.index_of(this.conversation);
+        index += backwards ? -1 : 1;
+        select_conversation(conversations.get(index % conversations.size));
```
fiaxh: -1 % n != n
It wouldn't go to the last conversation when pressing Ctrl+Shift+Tab on the first conversation.

main/src/ui/unified_window_controller.vala
```diff
@@ -142,6 +142,15 @@ public class UnifiedWindowController : Object {
         }
     }
 
+    public void loop_conversations(bool backwards = false) {
```
fiaxh: I think it would be nicer to select the next/previous row based on data available to the UI classes instead of using high-level information on conversations and re-implementing the sorting.

fiaxh:
Thanks, merged in 27fe07c. Changes made in d9e4507.

main/src/ui/unified_window_controller.vala
roobre: That's what I initially thought too, but since they were stored in an unordered data structure and it weren't exposed I chose the other thing.
Your solution looks nice to me though, thanks!

Issue #383, opened 2018-06-21
Dino lumps all chats together without clear indication which account I am using


Mikaela:
https://github.com/siacs/Conversations/issues/3074 is longer version on similar issue I opened to Conversations.

My ideas from there that might be appliable to Dino would be having a text "send un/OMEMO/encrypted message from/as JID" in the input box which is currently empty or having Firefox Container Style symbols that could be associated to accounts and shown in front of them to hint which account is being used.

I have lately been testing accounts on two different servers which cause some chats to be confusingly duplicated and managed "organisation account" with my own devices causing their rooms to be duplicated for me.

BigGiantHeadsHighCommander:
On your device don't use the same avatar with each of your accounts.

On Conversations YOU can override a contact's avatar (assumes either not linked to contact provider, or is linked but no contact image provided).  With this feature YOU can visually create separation yourself.

Mikaela:
That doesn't help me with Dino and MUCs where having multiple accounts in the same MUC doesn't even say "me: message" in previews and makes figuring out which account is being used more difficult than with Conversations.

Mikaela:
I heard another use case that seeing the nickname would be useful when you are using gateways/transports like Biboumi for IRC and in case of common name have to use a different nickname in IRC.

trymeouteh:
Please sort all conversations by account and the ability to switch between accounts

pravi:
At least for personal chats, I think following the approach of SMS apps and multiple sim cards would fit well. Show/merge chats from the same contact in a single chat and allow switching the account when sending a message. So in SMS app, I can see sim 1 is being used right now, but on tapping the icon, I can switch to sending with sim 2.

Now pin messages are confusing as well. I pinned one chat from a person I considered important, but they sent me a text using my second account and that chat was down the list and I missed it. This probably need additional work to make sure the contact is added in both roaster etc. but clubbing the chats is a good start.

TriggerDingus:
I find this to be a big issue for me, as well. I have two accounts and the user interface is absolutely not intended to be used to manage multiple accounts. When I receive in incoming message, the interface doesn't tell me to which account the message was sent. When I go to send a message, let's say to a saved contact, it doesn't prompt me from which account I wish to send the message from - it assumes the last account I used to interface with that contact. These issues are rather annoying since I use Dino for business and so being able to easily see what account I'm communicating from is important but Dino just isn't set up for multiple account management.

TriggerDingus:
Related:
[...]

Issue #525, opened 2019-02-15
[UI] Put own messages to the right window side


Echolon:
Hi,

would like to suggest to offer the possibility to put your own messages to the right side, like in most other client. At least an opt-in option would be great.

Hope that not a double. If so, please close

Cheers

licaon-kter:
This works on a narrow window, yes, else this is a desktop app so your messages will end up pretty strange.

Echolon:
Yes, right, but I always put Dino to the right screen corner. Thereby, the chat window becomes narrow.

Another thing could be to mark it with a differnt background. but I guess its not intended...

fiaxh:
This isn't something we plan to do at the moment.

PR #549, opened 2019-04-22
Display avatar of current account next to the chat input line

aearil:
There is currently no way to easily know which account is being used to send messages in a conversation. This can get especially problematic when the user is connected to the same MUC with multiple accounts. The issue had already been mentioned in  #383

This change allows users to know which of their accounts is used to post in the current conversation by showing the avatar of the current account next to the text input area.

As this is my first time using Vala and GTK templates, I would gladly accept improvements on the layout elements I added.

### DIFF
--- a/main/data/chat_input.ui
+++ b/main/data/chat_input.ui
@@ -9,6 +9,23 @@
             <class name="dino-chatinput"/>
         </style>
         <child>
+            <object class="GtkBox">
+                <property name="orientation">horizontal</property>
+                <property name="margin">12</property>
+                <property name="margin-top">0</property>
+                <property name="margin-end">0</property>
+                <property name="visible">True</property>
+                    <child>
+                        <object class="DinoUiAvatarImage" id="image">
+                            <property name="height">35</property>
+                            <property name="width">35</property>
+                            <property name="valign">center</property>
+                            <property name="visible">True</property>
+                        </object>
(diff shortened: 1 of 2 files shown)


fiaxh:
Often users only have a single account. In those cases such an indicator wouldn't be of any help. Even if a user has multiple accounts (personal/work, different identities), those tend to have a largely non-overlapping set of contacts - again, no need for an indicator.

I don't think being twice in a MUC is a big enough use case to justify having an additional, prominent element in the UI. Especially given that the information is only one click ("Conversation Details") away already.

aearil:
I don't agree with you at all on this.

There are a lot of reasons to maintain kinda overlapping identities online. After all we are talking about a messaging app/protocol that cares a lot about privacy and security. And part of that is doing our best not to put the user in confusing situations.

One particularly disastrous example I can think of is LGBT+ people using in and out of the closet accounts. The stakes are high and making a mistake quite easy. You want to provide the information in a passive way, allowing users to potentially realize they're making a mistake.

The solution you're proposing, clicking on "Conversation Details" only works if you actively seeks the information, not as a passive reminder.

Moreover, the change is not that drastic. It only accounts for a negligible loss of screen real estate. The only issue I can think of is in the case of a mobile, in which case a design the one mentioned in #525 can be used.

PR #550, opened 2019-04-22
Add a tooltip when messages are in WONTSEND state

aearil:
The warning sign and red color, while obviously synonymous with a
problem, are still a bit perplexing for the user.

This change adds a bit of clarity.

It would obviously be better to get exact cause of the problem from the
different plugins. Maybe it would be possible to add a field to the
Message class from libdino/src/entity/message.vala in order to record an
error message for every case.


fiaxh:
Thanks, merged in ed71f61.

PR #551, opened 2019-04-24
Add a man page

Sylvhem:
This add a standard UNIX manual page to Dino. Since I'm not a developer myself, this could probably benefit from the feedback of someone better informed about Dino's inner working.
I choose "0.0" as a version number since it's the number one. This needs to be updated the next time a release will come out.


bowlofeggs:
Nice!

christf:
What is preventing a merge of this?

PR #552, opened 2019-04-24
Sort conversations by accounts in the selector (fix #383)

### LINKED ISSUE #383: Dino lumps all chats together without clear indication which account I am using

aearil:
The conversations in the sidebar are now grouped by their respective
accounts to further clarify which account is used to talk in which
conversation.

Accounts are sorted alphabetically by display_name, and conversations
are then sorted by most recent inside each account.

However, I couldn't manage to connect a signal when the account alias is
modified in order to update the labels on the separators.

Also, I feel like some king of GtkTree* layout might have been more
semantically appropriate here, but would have have required a much
bigger refactoring.

### DIFF
--- a/main/CMakeLists.txt
+++ b/main/CMakeLists.txt
@@ -35,6 +35,7 @@ set(RESOURCE_LIST
     conversation_list_titlebar_csd.ui
     global_search.ui
     conversation_selector/chat_row_tooltip.ui
+    conversation_selector/conversation_account_separator.ui
     conversation_selector/conversation_row.ui
     conversation_summary/image_toolbar.ui
     conversation_summary/view.ui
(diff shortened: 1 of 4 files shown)


fiaxh:
If you have many conversations open - which I believe is a very common thing - this PR makes the conversation selector rather unusable. One would have to scroll down and search for the account delimiter every time they want to select a conversation from anything but the first account. This defies the idea of being able to quickly select the conversation one most likely want's to select - the most recently active ones. Conversations in which someone just messaged me would be buried under many irrelevant, inactive conversations from another account.

aearil:
That's fair.
Maybe there are some less invasive ideas worth investigating. For example, just thinking aloud, using a color coded strip on the selector rows to mark the account.

PR #577, opened 2019-06-13
Jingle file transfers via in-band bytestreams

hrxi:
What works? Outgoing file transfers.

What is still missing? Incoming file transfers, UI, error handling.

**Update**: Error handling has been improved, UI has been reused. It still needs to be enabled by default (it currently isn't because proper prioritization of HTTP over Jingle isn't implemented).

**Update**: Enabled by default and added incoming file transfers.


hrxi:
We talked about the design and found that we'd want to replace the socket-style API for the Jingle sessions with handing the Jingle session an input stream directly.

libdino/src/entity/file_transfer.vala
```diff
@@ -53,6 +53,7 @@ public class FileTransfer : Object {
     }
     public string path { get; set; }
     public string? mime_type { get; set; }
+    // TODO(hrxi): expand to 64 bit
```
mar-v-in: Just do it.

**Not** blocking for merging.

libdino/src/service/file_manager.vala
```diff
@@ -66,6 +66,7 @@ public class FileManager : StreamInteractionModule, Object {
         foreach (FileSender file_sender in file_senders) {
             if (file_sender.can_send(conversation, file_transfer)) {
                 file_sender.send_file(conversation, file_transfer);
+                return;
```
mar-v-in: Should be a `break;`, no?

xmpp-vala/src/module/xep/0047_in_band_bytestreams.vala
```diff
@@ -0,0 +1,437 @@
...
+    }
+    public void trigger_write_callback() {
+        if (write_callback != null) {
+            Idle.add((owned) write_callback);
+            write_callback = null;
+        }
+    }
+
+    public async ssize_t read_async(uint8[]? buffer, int io_priority = GLib.Priority.DEFAULT, Cancellable? cancellable = null) throws IOError {
+        // TODO(hrxi): cancellable?
```
mar-v-in: - Check for `cancellable.is_cancelled()` in the loop and early return if set.
- Use `cancellable.cancelled()` signal to trigger the read callback.

Such that future calls to `cancellable.cancel()` will cancel the execution of `read_async()` as soon as the main loop gets free. Note that I don't think we need this behavior anywhere and likely will just pass null to the cancellable, but it shouldn't be hard to implement either and it might be considered functional at some point.

**Not** blocking for merging.

xmpp-vala/src/module/xep/0047_in_band_bytestreams.vala
```diff
@@ -0,0 +1,437 @@
...
+        // TODO(hrxi): io_priority?
```
mar-v-in: The priority would have to be used as a second parameter to the `Idle.add()` call in `trigger_read_callback()`.

**Not** blocking for merging.

xmpp-vala/src/module/xep/0047_in_band_bytestreams.vala
```diff
@@ -0,0 +1,437 @@
...
+    }
+    public void handle_data(XmppStream stream, StanzaNode data, Iq.Stanza iq) {
+        assert(state == CONNECTED);
+        if (input_closed) {
+            set_error("unexpected data");
+            stream.get_module(Iq.Module.IDENTITY).send_iq(stream, new Iq.Stanza.error(iq, new ErrorStanza.not_allowed("unexpected data")));
+            return;
+        }
+        int seq = data.get_attribute_int("seq");
+        // TODO(hrxi): return an error on malformed base64 (need to do this
```
mar-v-in: You can check that `content.length` = `base64.length`/4*3 - `number of trailing =` to verify that the base64 string was fully decoded and as such valid.

**Not** blocking for merging.

libdino/src/entity/file_transfer.vala
hrxi: Not yet done, required some changes in database code but the database code didn't like `int64`.

libdino/src/service/file_manager.vala
hrxi: Done.

xmpp-vala/src/module/xep/0047_in_band_bytestreams.vala
hrxi: Done.

xmpp-vala/src/module/xep/0047_in_band_bytestreams.vala
hrxi: Done.

xmpp-vala/src/module/xep/0047_in_band_bytestreams.vala
hrxi: Not yet done, seemed harder than the proposed method (whitespace is valid in base64).

libdino/src/service/jingle_file_manager.vala
```diff
@@ -0,0 +1,121 @@
...
+            file_transfer.provider = 0; // TODO(hrxi): what is this?
+            file_transfer.info = id;
+            file_transfers[id] = jingle_file_transfer;
+
+            file_incoming(file_transfer, conversation);
+        });
+    }
+
+    async void get_meta_info(FileTransfer file_transfer) {
+        // TODO(hrxi): what is this function?
```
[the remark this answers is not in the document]
hrxi: Done.

libdino/src/service/jingle_file_manager.vala
```diff
@@ -0,0 +1,121 @@
...
+            file_transfer.counterpart = jingle_file_transfer.peer.bare_jid;
+            file_transfer.ourpart = account.bare_jid;
+            file_transfer.encryption = Encryption.NONE;
+            file_transfer.time = new DateTime.now_utc();
+            file_transfer.local_time = new DateTime.now_utc();
+            file_transfer.direction = FileTransfer.DIRECTION_RECEIVED;
+            file_transfer.file_name = jingle_file_transfer.file_name;
+            file_transfer.size = (int)jingle_file_transfer.size; // TODO(hrxi): remove cast
+            file_transfer.state = FileTransfer.State.NOT_STARTED;
+            file_transfer.provider = 0; // TODO(hrxi): what is this?
```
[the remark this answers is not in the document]
hrxi: Done.

PR #579, opened 2019-06-20
Add a couple of error primitives for easier error handling

hrxi:


hrxi:
Second commit is optional.

PR #588, opened 2019-07-28
Fix Manage Key dialog string for verified keys

ullman:
The OMEMO key management dialog is currently not properly formatted for verified keys.
To reproduce: Click on a verified OMEMO key in the key management dialog.
Expected: `This means it can be used by name@xmpp.server to receive and send messages.`
Observed: `This means it can be used by %s to receive and send messages.`


fiaxh:
Thanks!

PR #605, opened 2019-08-28
Consistent "Could not", instead of "failed"

comradekingu:
"Could not connect to %s"
https://github.com/dino/dino/blob/master/main/src/ui/manage_accounts/add_account_dialog.vala#L239
"Could not establish a secure connection"
https://github.com/dino/dino/blob/master/main/data/manage_accounts/add_account_dialog.ui#L153


fiaxh:
Thanks, partially merged in a10c821.

PR #608, opened 2019-08-31
Add libnice and listen for direct connections in Jingle SOCKS5

hrxi:
Add libnice as a plugin. If it is present, use libnice to enumerate
local IP addresses and listen on them to support direct connections for
Jingle SOCKS5.

Tested with Conversations and Gajim.


hrxi:
Added the `Nice-0.1.metadata`. :)

DanScharon:
any blockers why this isn't merged?

fiaxh:
This has been merged in 148cf48d2b68354881066e2587e2673c91d2619a.

PR #620, opened 2019-09-29
Update add_account_dialog.vala

E-404:
Add new server 404.city

### DIFF
--- a/main/src/ui/manage_accounts/add_account_dialog.vala
+++ b/main/src/ui/manage_accounts/add_account_dialog.vala
@@ -63,6 +63,7 @@ public class AddAccountDialog : Gtk.Dialog {
     [GtkChild] private Button success_continue_button;
 
     private static string[] server_list = new string[]{
+        "404.city",    
         "5222.de",
         "jabber.fr",
         "movim.eu",


moppman:
Out of the current server list for users to pick from, only jabber.fr (and 404.city from this PR) has a 100% rating on complicance.conversations.im, so imho the list should definitely be updated for a better user experience.
I know that the list is not without flaws, but it's the best metric we currently have, isn't it?
Also, I think a link to https://status.conversations.im/historical/ and complicance.conversations.im would make sense.

fiaxh:
I feel that picking servers for this list is a rather difficult thing.
Of course we ideally want servers that support all the XMPP features that Dino supports. They should have a reasonable privacy policy. They should have a good record of uptime. Those three things can easily be checked.

However, there are also aspects that are more difficult to estimate: We want servers that won't go offline anytime soon. We want servers that won't close registration anytime soon (We can't update the Dino version in Debian Stable / Ubuntu, and they stay around for more than two years). We want servers that are maintained by multiple people.

We'll need a better solution for our server suggestions. Until we found one, I don't see us adding servers from PRs. If you have a strong case for one specific server, please feel free to mention it in our XMPP channel.

PR #632, opened 2019-10-18
Add an option to enable OMEMO by default in new conversations

lumi-me-not:
This adds a checkbox to the settings dialog with the label "Enable OMEMO by default", which, when checked, will cause new conversations to start out with OMEMO encryption.

### DIFF
--- a/libdino/src/entity/settings.vala
+++ b/libdino/src/entity/settings.vala
@@ -11,6 +11,7 @@ public class Settings : Object {
         send_marker_ = col_to_bool_or_default("send_marker", true);
         notifications_ = col_to_bool_or_default("notifications", true);
         convert_utf8_smileys_ = col_to_bool_or_default("convert_utf8_smileys", true);
+        omemo_default_ = col_to_bool_or_default("omemo_default", false);
     }
 
     private bool col_to_bool_or_default(string key, bool def) {
@@ -63,6 +64,18 @@ public class Settings : Object {
                     .value(db.settings.value, value.to_string())
                     .perform();
             convert_utf8_smileys_ = value;
+        }
+    }
+
+    private bool omemo_default_;
+    public bool omemo_default {
+        get { return omemo_default_; }
+        set {
+            db.settings.upsert()
+                    .value(db.settings.key, "omemo_default", true)
+                    .value(db.settings.value, value.to_string())
+                    .perform();
+            omemo_default_ = value;
         }
     }
 }
--- a/libdino/src/service/conversation_manager.vala
+++ b/libdino/src/service/conversation_manager.vala
@@ -8,6 +8,8 @@ public class ConversationManager : StreamInteractionModule, Object {
     public static ModuleIdentity<ConversationManager> IDENTITY = new ModuleIdentity<ConversationManager>("conversation_manager");
     public string id { get { return IDENTITY.id; } }
 
+    private Dino.Entities.Settings settings = Dino.Application.get_default().settings;
+
     public signal void conversation_activated(Conversation conversation);
     public signal void conversation_deactivated(Conversation conversation);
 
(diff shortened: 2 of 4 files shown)


mar-v-in:
- Dino supports OMEMO and OpenPGP alike. It is unreasonable to have an option to enable OMEMO by default and not OpenPGP
- This setting should probably be on an account level (or possible to override on account level): Some companies do not allow to send e2ee messages due to policy (all messages have to be archived) and thus block those messages server-side. Enabling OMEMO by default in such cases will just be bad for UX, but I might use the same client for a private account where I do want e2ee.

licaon-kter:
@mar-v-in so Dino would accept a PR for a per account three state switch: OMEMO/PGP/NONE ?

mar-v-in:
@licaon-kter I wouldn't know where in the UI would be a good place for that. We plan on changing the account management screen at some point, but that's not yet done (not even the design work) and I don't feel it really fits in the current account screen.

We will definitely include it in the considerations for the new account screen.

dvn0:
>     * Dino supports OMEMO and OpenPGP alike. It is unreasonable to have an option to enable OMEMO by default and not OpenPGP

I understand this perspective, but things don't have to be perfect to begin with. Right now the defaults are unsafe for encrypted messaging, and this PR partially remedies that, and does not make it worse off for PGP users.

stevenroose:
> @licaon-kter I wouldn't know where in the UI would be a good place for that.

How about right under the OpenPGP and OMEMO items in the account screen? There can be a dropdown: Default encryption: None V (that drops down to OMEMO and OpenPGP).

> Dino supports OMEMO and OpenPGP alike.

@mar-v-in  I do understand this ideal principally, but like @dvn0 says, things don't have to be perfect from the start. OMEMO is clearly way more popular in the community than OpenPGP is, so this PR makes the user experience strictly better.

I personally use OMEMO everywhere and I keep making my first message(s) in a new conversations unencrypted until I notice or am pointed out that I'm not using OMEMO.

kousu:
Maybe #845 should be accepted instead. Instead of adding any new UI, it just makes all conservations OMEMO by default (which, I agree, is far far more popular than OpenPGP). If a user wants to not use OMEMO they can toggle it per-conversation? That's better than having any extra option to think about ahead of the time; the more options added to software the harder it is to use.

Someone who primarily used OpenPGP and was resisting OMEMO would probably find this really annoying. I know I would. But I'm not that person and if they feel that strongly about it they can use Gajim so... :shrug:?

kousu:
Not to say that this is not really good code @lumi-me-not. Thanks for helping us all out by submitting it and getting the conversation started :)

edef1c:
This patch left me in an unpleasant situation on a new machine: I realised the option was off by default _after_ every conversation had already been instantiated, so I had to manually enable OMEMO on every conversation _and_ check the box.
I spent a while whittling this patch down to something secure by default, and discovered that I'd converged exactly with #845, which seems like a far more robust solution. It could be paired with some better UI around MUCs that don't support OMEMO, but hiding that UX hazard behind a flag isn't helping anyone.

marmistrz:
> @licaon-kter I wouldn't know where in the UI would be a good place for that. We plan on changing the account management screen at some point, but that's not yet done (not even the design work) and I don't feel it really fits in the current account screen.
> 
> We will definitely include it in the considerations for the new account screen.

@mar-v-in is there an ETA for the new account screen? If not (or if it's far future), could we design the requirements for the minimal implementation?

marmistrz:
> @mar-v-in is there an ETA for the new account screen? If not (or if it's far future), could we design the requirements for the minimal implementation?
@mar-v-in bump

fiaxh:
OMEMO by default was added as an option as part of https://github.com/dino/dino/commit/21ae42762d8a57da5cb1ec40b46e7510fc3121ad.

PR #638, opened 2019-10-26
Add partial XEP-172 User Nickname support

lumi-me-not:
This adds partial XEP-172 User Nickname support in `vala-xmpp`, adds a `UserNickManager` service in `libdino`, adds a table for storing nicknames in the database and modifies the UI to show these nicknames when available.


Venefilyn:
Could someone take this up at some point?

PR #640, opened 2019-11-03
Fix reporting failed HTTP upload

wiktor-k:
Status code can't be less than 200 and greater than or equal 300 at the same time. Using || instead of && marks all uploads that do not have 2xx status code as "failed".


fiaxh:
Thanks!

PR #642, opened 2019-11-06
Add support for HTTP Upload headers

wiktor-k:
Some services use [Authorization header][0] to pass upload credential data. This avoids the token being exposed in server logs and is allowed by XEP-0363 since version 0.5.0.

[0]: https://xmpp.org/extensions/xep-0363.html#request

### DIFF
--- a/libdino/src/service/file_manager.vala
+++ b/libdino/src/service/file_manager.vala
@@ -358,6 +358,7 @@ public class FileSendData { }
 public class HttpFileSendData : FileSendData {
     public string url_down { get; set; }
     public string url_up { get; set; }
+    public HashMap<string, string> headers { get; set; }
 
     public bool encrypt_message { get; set; default=true; }
 }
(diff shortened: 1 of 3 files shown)


fiaxh:
Thanks for the PR.

It would be nice to generalize this code to also handle other headers. `HttpFileSendData` would then contain a header-name-to-value map `HashMap<string, string>`. The `HttpFileSender` would just apply all headers from the map. The XMPP parsing code in `0363_http_file_upload.vala` would only insert the three allowed headers into the map:

> Only the following header names are allowed: Authorization, Cookie, Expires.

wiktor-k:
Yep, definitely a good idea @fiaxh! Please check out my latest adjustments.

wiktor-k:
By the way, I was reading [Soup.MessageHeaders.append](https://valadoc.org/libsoup-2.4/Soup.MessageHeaders.append.html) docs recently and found this fragment:

>The caller is expected to make sure that name and value are syntactically correct.

In this implementation name is correct as it's from a 3 element known set but should we also somehow validate the value of the header? [XEP-0363 says](https://xmpp.org/extensions/xep-0363.html#request):

> The requesting entity MUST strip any newline characters from the header name and value before performing the HTTP request.

But I'm wondering if given `append`'s docs even stronger validation shouldn't be needed there (depending on what the "syntactically correct" means, quick browsing through RFCs [didn't yield definitive answers](https://github.com/rack/rack/issues/764)).

wiktor-k:
Rebased on top of master to avoid the "compilation issue".

fiaxh:
Thanks!

wiktor-k:
No problem. Thank you for taking the time to review and test this!

PR #660, opened 2019-11-23
Remove background transparency of text field

### LINKED ISSUE #653: Message input is rendered with garbage.
Dino's message input is rendered with garbage.
Dino's version: actual git master.
OS: Debian Sid
GTK: 3.24.12
Theme: Adwaita Dark
Screenshot from 2019-11-20 07-35-21

mbeko:
Fixes #653.

I couldn't think of a good reason why the background should be transparent, so I removed this CSS instruction and the text is now rendered correctly on my system.

### DIFF
--- a/main/data/theme.css
+++ b/main/data/theme.css
@@ -110,6 +110,11 @@ window.dino-main button.dino-chatinput-button:checked:backdrop {
 }
 
 
+.dino-chatinput textview, .dino-chatinput textview text {
+    background-color: @theme_base_color;
+}
+
+
 /*Chat input warning*/
 
 box.dino-input-warning frame border {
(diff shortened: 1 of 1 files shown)


main/data/theme.css
```diff
@@ -110,10 +110,6 @@ window.dino-main button.dino-chatinput-button:checked:backdrop {
 }
 
 
-.dino-chatinput textview, .dino-chatinput textview text {
-    background-color: transparent;
```
fiaxh: Could you try to set the `background-color` to `@theme_base_color` instead of removing it?

mbeko: It works :raised_hands:

fiaxh:
Great, thanks for debugging this!

PR #689, opened 2019-12-28
Convert :short_code: to emojis

### LINKED ISSUE #395: emoji autocompletion
It would be nice to be able to add an emoji by their namecode with colons:
: tada : becomes :tada: for instance

alyssarosenzweig:
We reuse the emoji database packaged with GTK for the emoji picker
itself; this is accessible as a resource (a serialized GVariant) which
we may unpack to form our own shortcode->emoji hashtable, which hooks in
nicely to the existing smiley conversion infrastructure.

Closes #395

### DIFF
--- a/main/src/ui/chat_input/smiley_converter.vala
+++ b/main/src/ui/chat_input/smiley_converter.vala
@@ -9,7 +9,9 @@ namespace Dino.Ui.ChatInput {
 class SmileyConverter {
 
     private TextView text_input;
+    private GLib.Regex colon_regex;
     private static HashMap<string, string> smiley_translations = new HashMap<string, string>();
+    private HashMap<string, unichar> emoji_translations = new HashMap<string, unichar>();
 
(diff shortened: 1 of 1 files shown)


bobbybee:
v2 fixes an issue where the cursor would be messed with when editing a message (caveat: now short code conversion only applies at the end of a line), lets the conversion apply immediately on the latter colon rather than waiting for a space, and should improve performance a bit.

bobbybee:
GTK has a native emoji autocomplete ... wondering if it would be better to figure out how to use that instead.

Ppjet6:
I haven't read the changes in detail. Would it be possible to not force this onto the user? Say I type `:x:` and I want it to be displayed as is. Maybe a solution similar to https://lab.louiz.org/poezio/poezio/issues/3455#note_7769

(I'm happy to move this discussion somewhere more appropriate)

PR #691, opened 2019-12-29
Use local alias in group PMs

alyssarosenzweig:
Rather than displaying our bare JID for ourselves.


mar-v-in:
We now display the nickname that you also use in the corresponding group. That seemed more reasonable than the local alias (because it's now the same on both sides and you can see which nickname you are actually sending).

Thanks nonetheless for your work.

PR #701, opened 2020-01-02
Add inline audio widget

alyssarosenzweig:
In addition to viewing inline images, we can support inline audio (like
Conversations' voice messages) via gstreamer. The code is generic enough
to additionally support inline video via GstGtkSink in the future, but
for now the widget is audio-only to ease review.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ env:
 before_install:
   - export USE_CCACHE=1
   - sudo apt-get update
-  - sudo apt-get install -y cmake valac libgee-0.8-dev libsqlite3-dev libgtk-3-dev libnotify-dev libgpgme-dev libsoup2.4-dev libgcrypt20-dev libqrencode-dev
+  - sudo apt-get install -y cmake valac libgee-0.8-dev libsqlite3-dev libgtk-3-dev libnotify-dev libgpgme-dev libsoup2.4-dev libgcrypt20-dev libqrencode-dev libgstreamer1.0-dev gstreamer1.0-plugins-good
 install:
   - ./configure --with-tests $CONFIGURE_FLAGS
   - make
--- a/main/CMakeLists.txt
+++ b/main/CMakeLists.txt
@@ -12,6 +12,10 @@ find_packages(MAIN_PACKAGES REQUIRED
     ICU
 )
 
+set(MAIN_PACKAGES ${MAIN_PACKAGES} gstreamer-1.0)
+find_package(PkgConfig REQUIRED)
+pkg_check_modules(GST REQUIRED IMPORTED_TARGET gstreamer-1.0)
+
 set(RESOURCE_LIST
     icons/dino-changes-prevent-symbolic.svg
     icons/dino-conversation-list-placeholder-arrow.svg
@@ -48,6 +52,7 @@ set(RESOURCE_LIST
     conversation_selector/chat_row_tooltip.ui
     conversation_selector/conversation_row.ui
     conversation_summary/image_toolbar.ui
+    conversation_summary/multimedia_toolbar.ui
     conversation_summary/item_metadata_header.ui
     conversation_summary/view.ui
     manage_accounts/account_row.ui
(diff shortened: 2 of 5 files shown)


mar-v-in:
Thanks for your work!

We usually try to keep the base of Dino with a small set of dependencies and including a large dependency like GStreamer for just a "small" feature like inlining the audio doesn't really help with that goal.

Instead it would probably help to have this feature in a plugin. Unfortunately we don't have a plugin entry point for changing how files are displayed, as we only do images right now and they don't use a plugin entry point.

Also, when we allow plugins to inject widgets for files, we'd still want them to follow some common style that is enforced through the main code, as this deduplicates code and makes changes easier. We had some sketch on how to do this already, but it is out of scope for the 0.1 release and thus was no focus in the last months.

Speaking of 0.1 release: we do not plan to merge any new features to master before the release, only bug fixes. So please bear with me putting this on the not needing immediate action list.

alyssarosenzweig:
Thank you for looking into this :)

I agree that this feature alone wouldn't justify a new big dependency.
That said, I envision Dino getting a lot more mileage out of gstreamer
than just inline audio (and video). For starters, if we want to do
Conversations-style voice (or video) messages, gstreamer is a de facto
necessity there. But mostly I'm optimistic about Dino implementing voice
calls (via Jingle) in the nebulous future :-)

In the mean time, perhaps a build-time config option could enable
multimedia code (pulling in the dep)? That way these features don't get
in the way of porting core Dino to other platforms (although gstreamer
should be quite portable in the long-term), and builds of Dino for
"smaller" devices don't carry the dependency footprint. For distro
builds, that'd be up to the distro maintainer, of course, but I suspect
most mainstream distributions (Debian/Ubuntu/Fedora/etc) would choose to
add the gstreamer dependency, since in practice gstreamer will be
installed and heavily integrated with systems using Dino (desktop
Linux). If Dino were a command-line client, it'd be a very different
story.

I'm not sure how moving the code to a plugin would help, but I'm open to
suggestions. I think the build time decision can happen even with
dino/main. Of course distributions could split into dino-im and
dino-im-media packages, with only the latter depending on gstreamer, but
I'm not sure the extra bump in complexity on our end is worth it? Like I
said, I'm open to options.

What is blocking the 0.1 release? Any particular bugs? Any particular
date? Anything I can do to help there?

Thank you again :)

~A

mar-v-in:
Calling is definitely on our list of wanted features, but this could also be realized in a plugin (in fact there is an initial+outdated prototype for it already somewhere on my disk).

I like how you mention the command-line client, because when developing Dino we indeed always consider, "what if there was an additional command-line client usind libdino". Notice that libdino does not even have GTK as a dependency, so this would be entirely possible - even though unlikely to happen.

An example for this moving the code to a plugin would be the notification-sound plugin, that is in fact disabled by default in most distributions. That plugin currently has a dependency on libcanberra, but because distributions decided not to ship it, dino doesn't have that dependency. The idea is that distributions can then decide to ship a second package dino-plugin-notificiation-sound which has the dependency to libcanberra or that users just add the notification-sound plugin to their system without it being packaged by the distribution.

Thus using the plugin design in the end might make things easier for the users, as they don't have to recompile everything if they'd like a different set of features/dependencies as the distribution. If we had proper support for file display with plugins, there could also be an inline PDF preview - but a dependency on any PDF viewer library is definitely a bit awkward for a chat client.

And by requiring ourselves to use plugins / a plugin API for certain things (that feel like they might be places for future extensions like exactly this example of file type specific handling) will likely end up in a cleaner design with a good extension points.

We don't have any blocker bugs for the release, we are mostly in a waiting period before the release to ensure we didn't introduce any critical bugs in the last two weeks that haven't been reported yet

Echolon:
If installing plugins would be made somewhat easy for everyone, that is a fair option I think!

wurstsalat3000:
Thank you @alyssarosenzweig for working on this!

FYI: I took some inspirations from this PR and ported it to python, so we can use it for Gajim’s preview plugin. I hope you’re alright with that. See: https://dev.gajim.org/gajim/gajim-plugins/-/merge_requests/200

alyssarosenzweig:
@wurstsalat3000 Neat stuff! But don't think you can tempt me away from Dino this easily ;)

PR #710, opened 2020-01-05
Allow JID to be translated

licaon-kter:


fiaxh:
How would you translate JID? JID stands for "Jabber ID" or "Jabber Identifier". According to an online translation, that would be "Pengenal Jabber" in Malay. But translating "JID" to "PEJ" would only make limited sense since it's an unknown abbreviation.

licaon-kter:
`JID` doesn't mean anything in my locale, why can't I translate it to something less cryptic? 

If "Jabber" is so important why does it appear [only once](https://github.com/dino/dino/search?l=Gettext+Catalog&q=jabber&type=) in all the texts?

**I** know what it stands for, and I also know that "Jabber" is a trademark that Dino doesn't hold, so JID not only is an useless acronym for non-English languages, but also it has been reduced to this 3 letter word from some fear of copyright and for historic reasons (afaik).

Yes, keep it as you wish **for** English but for the rest of the world allow us to adapt it so users can grasp it's meaning and purpose.

> JID stands for "Jabber ID" or "Jabber Identifier".

At least rename it to that then, and then allow us to translate it _(eg. I'd need to translate it as "ID Jabber" since that's the way it works in **my** locale)_

I, for one, am more inclined to translated it to ["XMPP Address"](https://github.com/siacs/Conversations/search?l=XML&q=xmpp+address) instead, because find it easier to explain it like "an email address" _(versus IMAP IDs or POP3 IDs ;))_

licaon-kter:
Also this: https://wiki.xmpp.org/web/Sprints/2019_January_Brussels/Pad
```
## Naming things?!
  * Jabber ID? JID? XMPP address?
  * "Jabber" trademark
  * "JID" widely used but might be a more technical term
  * "something address" is clear to people
  * "XMPP" is usually hard to remember, "Jabber" is more friendly to the ear
  * Jabber is the community, XMPP is the protocol
  * What about calling it "Conversations ID", or "Converse ID", or "Movim ID", all at the same time. Might be confusing for users when they use multiple clients but most don't.
  * Some other protocols use "username" (with a hint for the default domain)
  
Consensus: "XMPP Address"? (until another user-friendly term for "XMPP" can be found/agreed upon)

## Results / Consensus

### XMPP Address / Jabber ID
 - XMPP Address input: Always use one field. Never split into localpart and domain part. If possible do autocomplete on the server part in drop down
```

Zash:
*Address* or *XMPP address* is what's suggested by [Modern XMPP](https://docs.modernxmpp.org/terminology/).

PR #760, opened 2020-02-25
XEP-0363: Remove support for the old namespace

linkmauve:
The unversioned namespace got updated for the new :0 one in version 0.3
of XEP-0363, more than three years ago.  By now all server
implementations have probably updated, and this simplifies the code a
lot.  With this XEP having now reached the Draft status, there is no
reason to keep supporting the previous namespace.

Supersedes #161.

### DIFF
--- a/xmpp-vala/src/module/xep/0363_http_file_upload.vala
+++ b/xmpp-vala/src/module/xep/0363_http_file_upload.vala
@@ -4,8 +4,7 @@ using Gee;
 
 namespace Xmpp.Xep.HttpFileUpload {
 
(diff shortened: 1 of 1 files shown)


fiaxh:
I'm not convinced of removing this feature without any pressing need and without gaining anything besides minor code simplification. If at some point we'd gain advancements from not supporting `urn:xmpp:http:upload` anymore or if we'd have to do significant work to keep supporting it, we can of course remove it. But right now the code is already written and doesn't negatively effect other areas of Dino's code.

linkmauve:
Should I instead resurrect #161?  It’s been almost three years now.

PR #765, opened 2020-03-05
[UI] Fix menu button alignment

codedust:
Before:
image

After:
image

(the left hamburger menu button now is of the same height as all the other buttons)

### DIFF
--- a/main/data/conversation_list_titlebar_csd.ui
+++ b/main/data/conversation_list_titlebar_csd.ui
@@ -25,7 +25,6 @@
         </child>
         <child>
             <object class="GtkMenuButton" id="menu_button">
-                <property name="valign">center</property>
                 <property name="visible">True</property>
                 <child>
                     <object class="GtkImage">


main/data/conversation_list_titlebar_csd.ui
```diff
@@ -25,8 +25,8 @@
         </child>
         <child>
             <object class="GtkMenuButton" id="menu_button">
-                <property name="valign">center</property>
                 <property name="visible">True</property>
+                <property name="margin-end">5</property>
```
fiaxh: Other applications that have a horizontal line going through the headerbar (Geary, GNOME Settings, GNOME Contacts) don't set any margin on the buttons next to the line. So I guess we should leave that up to the theme, too.

codedust:
(Somewhat related: Maybe we should also choose another icon for the preferences menu to avoid two hamburger menus with the same icon (on the left and on the right).)

PR #767, opened 2020-03-07
Update README.md

Echolon:
Added some links and suggest to reference also for donations

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ Resources
 Contribute
 ----------
 - Pull requests are welcome. A good start are [good first issues](https://github.com/dino/dino/issues?q=is%3Aopen+is%3Aissue+label%3A%22good+first+issue%22). 
-Please discuss bigger changes in the [conference room](xmpp:chat@dino.im?join) first.
+Please discuss bigger plans on contributions first in our XMPP conference room 'chat@dino.im'.
 - Look at [how to debug](https://github.com/dino/dino/wiki/Debugging) Dino before you report a bug.
 - Help [translating](https://hosted.weblate.org/projects/dino/) Dino into your language.
 - Make a [donation](https://dino.im/#donate)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -23,9 +23,11 @@ Resources
 
 Contribute
 ----------
-- Pull requests are welcome. Please discuss bigger changes in the conference room first.
+- Pull requests are welcome. A good start are [good first issues](https://github.com/dino/dino/issues?q=is%3Aopen+is%3Aissue+label%3A%22good+first+issue%22). 
+Please discuss bigger changes in the [conference room](xmpp:chat@dino.im?join) first.
 - Look at [how to debug](https://github.com/dino/dino/wiki/Debugging) Dino before you report a bug.
 - Help [translating](https://hosted.weblate.org/projects/dino/) Dino into your language.
+- Make a donation: [Liberapay](https://liberapay.com/Dino), Paypal (`paypal at dino.im`) or on [Bounty Source](https://www.bountysource.com/teams/dino/issues)
```
mar-v-in, line 30: Given that the README is persisted in release archives, I think it would be better to link some place we can update. How about https://dino.im/#donate instead?

README.md
```diff
@@ -23,9 +23,11 @@ Resources
 
 Contribute
 ----------
-- Pull requests are welcome. Please discuss bigger changes in the conference room first.
+- Pull requests are welcome. A good start are [good first issues](https://github.com/dino/dino/issues?q=is%3Aopen+is%3Aissue+label%3A%22good+first+issue%22). 
+Please discuss bigger changes in the [conference room](xmpp:chat@dino.im?join) first.
```
mar-v-in, line 27: IIRC, GitHub does not support linking `xmpp:`-URIs, so this link wouldn't be displayed.

Echolon:
@mar-v-in I submitted the changes as requested.

PR #790, opened 2020-04-03
select_jid_fragment: try all accounts if jid doesn't have a @,

xaiki:
jids are only recognized if they have a '@' in them, we try to let people add contacts on the servers they are by autofiling the domainpart of their accounts when we look for valid jids

Signed-off-by: Niv Sardi <xaiki@evilgiggle.com>


fiaxh:
Thanks for your work, however I don't feel this change would improve the overall status.

The general issue is that it adds a new type of content to the list. Currently, it is a filterable contact list and it allows starting a conversation with a valid JID. This change would additionally make it a box for autocompletions, adding complexity and potentially confusion.

This specific implementation of the idea has additional issues:
- The autocompletion only suggests the domains of my own servers, however my contact might very well be on another server
- The autocompletion stops completing when the user starts typing the domain
- It prioritizes new JID suggestions over searching through existing ones

PR #801, opened 2020-04-08
Add keyboard shortcut to open the emoji picker

aearil:
Having to go search for the tiny button at the bottom right of my screen everytime I wanted to send an emoji was getting annoying fast.

This is a minor change to the code, but I think it's a significant quality of life upgrade.

For now, I chose Ctrl-E (as in emoji) as the shortcut, but I'm open to other ideas. For example, I know a few programs that use Ctrl-Shift-U (as in unicode).

### DIFF
--- a/main/data/shortcuts.ui
+++ b/main/data/shortcuts.ui
@@ -46,6 +46,19 @@
                         </child>
                     </object>
                 </child>
+                <child>
+                    <object class="GtkShortcutsGroup">
+                        <property name="visible">True</property>
+                        <property name="title" translatable="yes">Input</property>
+                        <child>
+                            <object class="GtkShortcutsShortcut">
+                                <property name="visible">True</property>
+                                <property name="accelerator">&lt;ctrl&gt;E</property>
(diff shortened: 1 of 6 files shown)


licaon-kter:
CTRL-dot already does that right?

aearil:
Whoops ! Indeed it does.

Maybe document that somewhere ?

licaon-kter:
Better PR that instead :))

aearil:
Yup :)

BTW, I still cannot find where in the code that Ctrl+. is defined oO

fiaxh:
Ctrl+. isn't a Dino feature, it's a GTK feature :slightly_smiling_face: That's also why it would be weird to mention it in our Shortcuts Dialog.

aearil:
Ok that makes sense, but at the same time how is one supposed to discover this ? Even when explicitely searching for it online, I still haven't been able to find a single mention of that in the GTK docs.

And I'm not sure users really care about the difference between GTK and Dino shortcuts, especially when they're as obscure as this one :/

Issue #704, opened 2020-01-02
Message styling: Reduce constrast slightly of styling commands


ericschdt:
like Conversations does:  ![strong_plain1](https://user-images.githubusercontent.com/43318127/71697380-c2443c80-2daf-11ea-90aa-2b270e5db1b9.png)

This helps to visually differentiate between the actual content and unimportant styling elements.

Manav-Kumar:
I want to work upon it, but didn't understand it well. @fiaxh  Can you explain it please.

fiaxh:
@Manav-Kumar You can do text formating using a markdown-style syntax in Dino and other XMPP clients. For example like this: **\*emphasis\*** . The idea is to give the formating characters (like *) less contrast, e.g. make them grey. Like in the picture.

Manav-Kumar:
Ok, but I guess this feature is in yaxim client for android?

Manav-Kumar:
@fiaxh please give me some hint, about it.

PR #804, opened 2020-04-10
Adding spaces in beginning and ending of text

Manav-Kumar:
This pr is in regarding resolving the issue [https://github.com/dino/dino/issues/704](https://github.com/dino/dino/issues/704 ), The purpose of the pr is to distinguish the important part of bold text from it's unimportant (*) signs. So by appending spaces before and after the text, we can resolve the above the problem.


fiaxh:
Closing due to inactivity.

PR #809, opened 2020-04-17
HTTP-Upload: parsing max-file-size attribute

kmq:
When a server returns multiple `<x/>` elements
in response to a discinfo query, this change
uses all of them instead of just the first one.


fiaxh:
Thanks! :+1:

PR #811, opened 2020-04-18
Update last_activity before received_file signal to keep UI in sync

### LINKED ISSUE #791: Sending a file in MUCs does not count as activity
1. Find a conversation that is not the most recent one
2. Send a file, e.g. an image
3. Observe that item in the conversation list did not move to the top.

kmq:
Fixes #791 where a signal to update the UI is sent to
late when sending files, resulting in an out of sync
conversations list.


fiaxh:
Thanks!

PR #821, opened 2020-04-22
Fix #774: Prevent dino from sending whitespace only messages

### LINKED ISSUE #774: Prevent dino from sending whitespace only messages
Currently Dino allows to send messages with whitespace-only bodies (spaces, newlines, tabs).
While this is not a big deal, I'd tend to expect such messages being trimmed down before sending.

Edit: Some clients like Miranda-IM don't show whitespace only messages at all, while eg. Conversations trims them down before displaying.

selurvedu:
Fixes #774.


fiaxh:
Thanks.

Issue #844, opened 2020-05-13
Encryption by default


tg-x:
Encryption should be enabled by default.
Currently one has to manually enable it for every conversation.

dvn0:
+1

Particularly enabling omemo by default. This will help prevent accidentally sending messages unencrypted. (Which I've done several times with Dino)

licaon-kter:
No need for a Dino default, just have a toggle in settings so users can do this, globally or per account.

dvn0:
If it's not default, then lots of people won't use it or notice the option.

I agree there should be a toggle in the settings for the default behavior, but that should be set to omemo to begin with.

licaon-kter:
Why not PGP then? Why not Unencrypted?

dvn0:
> Why not PGP then?

PGP requires account configuration before it will work. OMEMO does not.

>  Why not Unencrypted?

From my PR: "This helps to prevent the unintentional sending of unencrypted messages,
as well as promotes the usage of e2e encryption."

stevenroose:
Somehow it would be cool if the "I want to use OMEMO by default" setting could be persisted somewhere server-side. Perhaps having an OMEMO key node should be sufficient for a client to think the user wants to use OMEMO.

kalkin:
@stevenroose 

> Somehow it would be cool if the "I want to use OMEMO by default" setting could be persisted somewhere server-side.

Bad idea you do not want to trust the server.

stevenroose:
Well some people do trust their server to a big extent. I host the server for my social circles myself and while of course my friends don't want me to be able to read their conversations, they do trust me to not actively try mess up their privacy. I think that's a perfectly valid situation.

mar-v-in:
The information could be signed using your device key, so servers can't manipulate it. However existence of an OMEMO device list is not a good indicator for if OMEMO should be used, because that only indicates that one used at least one OMEMO capable client in the past. And the signed message is only worth something if you trust that device (i.e. verified it).

gsantner:
I agree, client side, enable encryption by default on, good idea.

However the "manage client side omemo configuration on server" gives me worries. I dont think storing client settings on server, and restoring them is and should be part of xmpp or dino.

Echolon:
rel. #733

Raviu56:
Should simply have an option on settings "enable omemo by default" with sub-option "enforce omemo for all conversations" so it cannot be disabled by mistake

selurvedu:
[...]

PR #845, opened 2020-05-13
Enable omemo encryption by default in new conversations

dvn0:
This helps to prevent the unintentional sending of unencrypted messages,
as well as promotes the usage of e2e encryption.

This PR will fix https://github.com/dino/dino/issues/844

### DIFF
--- a/libdino/src/entity/conversation.vala
+++ b/libdino/src/entity/conversation.vala
@@ -32,7 +32,7 @@ public class Conversation : Object {
             }
         }
     }
-    public Encryption encryption { get; set; default = Encryption.NONE; }
+    public Encryption encryption { get; set; default = Encryption.OMEMO; }
     public Message? read_up_to { get; set; }
 
     public enum NotifySetting { DEFAULT, ON, OFF, HIGHLIGHT }


deviant:
IMO this isn't a duplicate of #632, and is probably a better choice: leaving encryption disabled by default is a footgun. I've been running Dino with this patch for several months now, and I'm pretty happy with it.

edef1c:
I've convergently arrived at exactly the same patch, after #632 left me in an unpleasant situation on a new machine: I realised the option was off by default _after_ every conversation had already been instantiated, so I had to manually enable OMEMO on every conversation _and_ check the box.

dvn0:
cc @fiaxh @mar-v-in 

Can we please get some additional input on this. There are multiple PRs implementing different solutions for this, spanning multiple years now. Choose your pick.

Tangentially, why is @licaon-kter responding to so many issues and PRs, often speaking authoritatively as if they are a core contributor to the project. I was surprised to see they don't have commits in the source code. What's going on with the management of this project?

dvn0:
Closing in light of https://github.com/dino/dino/pull/1414

Issue #309, opened 2018-03-18
Windows Support


enko:
This ticket should serve as a kind of meta ticket, to track the efforts needed to bring dino to windows. I have automated the build in #304, so we can put out builds without any problems.

From my small assessment there are two high priority that needs a fix before any other work for the windows port can be done:

* `is_24h_format` from `helper.vala` crashes dino. From my limited knowledge of GTK, this seems to be ok to crash since there is no gnome-functionality on windows?
* Notifications are broken [upstream](https://bugzilla.gnome.org/show_bug.cgi?id=776583). As a possible fix, just disable the notifications on windows, till a proper backend lands in GTK?

On the chat there was a discussion if dino should look native, I think it should, but the first priority should be that it does not crash or behaves funny. Then there should be put energy into making dino look native.

Echolon:
I think it should be native, too. Is this only a question of GUI? Or are there other related issues if we decide for one or another? I mean, if people don't like it - fork it, do it the other way.... or?

Echolon:
Vote native

Echolon:
Vote against native

DeSaurus:
Someone made a dummy win32 notification backend for GNotification
https://bugzilla.gnome.org/show_bug.cgi?id=776583#c7
I hope this gets fixed soon

LuccoJ:
I think a "vote" for/against "native" when it hasn't been determined what "native" means just before the vote is not useful.

Echolon:
@LuccoJ HI, guess you are probably right. Thought it's clear - may someone help out with a good definition?

Anyway, the vote so far made clear that this is polarizing topic. That's an evaluation, too.

Echolon:
@enko Hi, any updates so far? If there are issues, maybe someone else could help?

LAGonauta:
Just FYI, Dino works on WSL with a vcXsrv after intalling all dependencies.
At first it crashed on load, but was fixed after running "sudo dpkg-reconfigure tzdata".

Echolon:
@LAGonauta   Thanks for your info. Any extra info we might need to know? If you can exactly document how you make it run, that would be great!

LAGonauta:
Sure:
**After fresh install of WSL Ubuntu 18.04**

## Update installation:
```
sudo apt update && sudo apt upgrade
```

## Install dependencies:
```
[...]

PR #851, opened 2020-05-20
Compatibility tweaks for Windows and MSYS2

LAGonauta:
Changes:
- ~~Fixes inability to build in MSYS2. Some Windows-native libraries had to be blacklisted as the CMake script is not able to find them. This is not necessary when cross-compiling from Linux.~~
- ~~Use `ShellExecuteA` instead of `AppInfo.launch_default_for_uri`. Allows Dino to launch UWP applications, can be removed after https://gitlab.gnome.org/GNOME/glib/-/issues/1991 is closed.~~
- Set file as downloaded from the internet so Windows is able to prompt the user to be sure they wants to open a file
- Get file mime-type from the extension on Windows. `FileInfo.get_content_type()` does not return correct values as it doesn't look inside the file. Fixes image preview.
- Increase font size on Windows a little. This normalizes the default appearence on Linux to Windows and makes it more readable.
- Added executable information and icon on Windows
- Added default PGPE path for PGP support

~~Notification currently works by flashing the taskbar entry, but should be complete when the `GNotification` backend is ready. Right now I do not think there is anything else missing for full feature parity.~~

Notifications are implemented fully as a plugin, this might eventually be replaced by a GNotification backend.

How to build on MSYS2:
https://github.com/dino/dino/issues/309#issuecomment-630493660

Any suggestion is welcome 🙂
Thanks!


cmake/PkgConfigWithFallback.cmake
```diff
@@ -13,11 +13,16 @@ function(find_pkg_config_with_fallback name)
         # Found via pkg-config, using it's result values
         set(${name}_FOUND ${${name}_PKG_CONFIG_FOUND})
 
+        if(MINGW)
+            list(FILTER ${name}_PKG_CONFIG_LIBRARIES EXCLUDE REGEX "(ole32)|(gdi32)|(imm32)|(shell32)|(winmm)|(dwmapi)|(setupapi)|(cfgmgr32)|(ws2_32)")
```
mar-v-in, line 17: This seems to be a hack specific to your build system. These libraries look like they actually are linker dependencies and thus should be included.

libdino/src/application.vala
```diff
@@ -58,6 +58,9 @@ public interface Application : GLib.Application {
         });
         shutdown.connect(() => {
             stream_interactor.connection_manager.make_offline_all();
+#if _WIN32
+            Dino.Util.CoUninitialize();
```
mar-v-in, line 62: According to documentation:
> A thread must call CoUninitialize once for each successful call it has made to the CoInitialize or CoInitializeEx function, including any call that returns S_FALSE.

This is only once per application start.

main/src/ui/main_window.vala
```diff
@@ -55,6 +55,17 @@ public class MainWindow : Gtk.Window {
     }
 
     private void setup_unified() {
+#if _WIN32 // WIN32 seems to not support exporting these symbols
```
mar-v-in, line 58: This shouldn't be needed and is a sign that the build is not correct. I am very certain that exporting symbols is possible, if it's not, you might be using the wrong linker.

plugins/gpgme-vala/src/gpgme_helper.vala
```diff
@@ -175,6 +175,9 @@ private static uint8[] get_uint8_from_data(Data data) {
 
 private static void initialize() {
     if (!initialized) {
+#if _WIN32
+        set_global_flag("w32-inst-dir", "C://Program Files (x86)//GnuPG//bin");
```
mar-v-in, line 179: A hard-coded path won't work if the installation differs from what you expect. If it isn't reasonably possible to automatically find the correct path on all systems, we'd need to allow users to enter their path manually.

main/dino-info.rc
```diff
@@ -0,0 +1,25 @@
+1 VERSIONINFO
+FILEVERSION     0,1,0,0
```
mar-v-in, line 2: Various versions in this file are hardcoded, but they should be set at build time.

mar-v-in (commented):
Instead of using ifdef's, please move win32 specific code in a dedicated plugin that is only enabled by default on win32 platforms. This should be feasible at least for the CSS/font related change.

libdino/src/application.vala
LAGonauta: Uhm.. Here it says that every call to `CoInitialize` must be balanced with a call to `CoUninitialize`:
https://docs.microsoft.com/pt-br/windows/win32/api/objbase/nf-objbase-coinitialize

Those COM stuff is confusing. I changed the code to make it call `CoUninitialize` as soon as possible after calling `CoInitialize`, not sure if this is correct...
Maybe I guess I could just remove this `CoInitialize` code, not sure if there is much use case to open something with components that require COM (such as shell extensions).

---

How many threads there is in Dino? Is the thread that starts/shutdowns Dino the same thread that calls ShellExecute?

main/src/ui/main_window.vala
LAGonauta: Done, I forgot to prefix the linker option with `-Wl`... that is why it wasn't working 🙄

main/dino-info.rc
LAGonauta: Done

plugins/gpgme-vala/src/gpgme_helper.vala
LAGonauta: Uhm... what is the best way to handle this? Get the path from an environment variable, or add a textbox to the account config screen when OpenPGP is enabled?

cmake/PkgConfigWithFallback.cmake
LAGonauta: Yes, it is a quirk of MSYS2. It is not needed when cross-compiling from Fedora.
Why does it need to manually find each DLL, by the way? Doesn't the linker know where they are anyway? 🤔

---

I changed the blacklist to explicity add the implicit library directory to the hint path, is this OK? Is there a better way?

libdino/src/application.vala
mar-v-in: Dino has multiple threads, but most of it happens in a single main thread. Start and shutdown should be in the same thread. But I guess your change is sufficient, even though I am not an expert on Windows things.

plugins/gpgme-vala/src/gpgme_helper.vala
mar-v-in: Do gpg4win installations set some environment variable? I've never used it, so I don't know what's the way to go. What do other apps do?

LAGonauta: Done

libdino/src/service/util.vala
```diff
@@ -26,6 +45,33 @@ public class Util {
         }
         assert_not_reached();
     }
-}
 
+    public static void launch_default_for_uri(string file_uri)
+    {
+#if _WIN32
+        ShellExecute("open", file_uri);
```
linkmauve, line 52: Any reason this fallback is done in Dino instead of Glib?

LAGonauta: Even though this works fine for Dino, it doesn't necessarily gives the control that GLib wants. See: https://gitlab.gnome.org/GNOME/glib/-/issues/1932#note_649492

This was also asked here: https://gitlab.gnome.org/GNOME/glib/-/issues/2088
With no direct answer. I can try to contact them on IRC to understand the issue completely, if you want.

LAGonauta:
Rebased to master and added two new commits:
- Export all plugins symbols on WIN32, it is necessary now. Maybe a change in newest GCC?
- Images preview now also works on Windows

Deluvi:
I just wanted to thank you for your recent fixes on the plugins symbols ! The Windows build works like a charm 👍

LAGonauta:
Rebased to master and added Windows 32-bit compilation support
Was win32 apis use __stdcall, so I was missing those special windows headers.

LAGonauta:
@fiaxh @mar-v-in 
Just pushed our notification work to this PR. Part of this PR are cppwinrt headers (licensed to MIT and distributed together with Windows SDK) modified by @Yuubi-san to compile on GCC. There is also a tool to generate those headers on the fly, however we did not manage to determine to which license they would be licensed yet.

However, I also noticed that Windows 8.1 is unable to load the plugin if Windows 8.1 SDK is not installed and those headers are used (work fine with the headers generated by the tool), even if the newest vcredist is installed. No solution has been found yet.

I chose to update this PR anyway to get feedback on anything else I should change.

LAGonauta:
5d285e15-2dc7-4a5e-a77a-6e40a0b8fc28

LAGonauta:
We found out what is causing Dino to not load the plugin on Windows 8.1, the linkage with the `mincore` library.
However, that library is not necessary when those WinRT headers are generated on-the-fly. We asked in `cppwinrt` repository and they think the generated code is licensed as MIT, but their answer did not go through their legal team:
https://github.com/microsoft/cppwinrt/issues/929

We are evaluating our options, but we would also like to know what you prefer. Some are:
1. Use headers generated by ourselves using YoloRT and commit them to the repository. Easy to use, trivial CI/CD and works everywhere. Fuzzy license state.
2. Use those explicitly MIT-licensed files from Windows SDK. Just as easy as last option, but does not work on Windows 8.1.
3. Generate the headers on compilation. Nothing is commited to Dino's repository, but the Windows binary might be tainted with code that might be (or not be) proprietary.

The perfect option is number 1, but might be risky. Option 2 is what we currently have, should be fine after January 10, 2023 (when official extended support for Windows 8.1 ends)
If we assume APIs are not copyrightable, number 3 might be our best option. If those APIs are ever ruled to be copyrightable and for some reason they are not MIT, it would be easy to just stop distributing binaries with the notification plugin built-in.

If Windows 8.1 support is not a priority, we can just use Option 2 (what we currently do)

Yuubi-san:
There's also the option
3a. Download pre-generated headers on build/configure time from... somewhere. The advantage being independence from Windows as a build host (Dino for Windows built fine on, e.g., Fedora, last time @LAGonauta checked). Of course, somebody somewhere still needs to generate the headers using windows metadata, but only once (or any time there are some new shiny features that Dino would like to use in, e.g., notification code).

Both option 3 and 3a don't create potential legal problems inside Dino's repo. 3 has the downside of requiring windows (license) to build the notification plugin; 3a suffers if/when the download location(s) go(es) down (like, microsoft suspending a github repo or its owner).

I'd also like to opine against dropping 8.1 support before *at least* MS officially stops patching its vulnerabilities. The mincore dependency is unnecessary bloat which is present only in the 1.5-year old version of the headers (the one from the SDK).

Yuubi-san:
Personally, I'm leaning towards option 1, but to be on the safe/pragmatic side, some variation of 3 is more preferable.

mar-v-in:
I would definitely prefer not having a bunch of header files that are not directly related to Dino in our source repository. If they can be generated at build time, that would be one option. Alternatively, requiring those header files to be around somewhere at build time would be fine as well.

In a perfect world those headers would be in a package that is discovered via pkg-config, but of course having an option to cmake would also be fine.
For CI builds we can also fetch them from somewhere, for release builds I wouldn't mind if those headers are only "assume"d to be MIT licensed - after all, this use is completely in the interest of the copyright holder anyways.

Yuubi-san:
> for release builds I wouldn't mind if those headers are only "assume"d to be MIT licensed - after all, this use is completely in the interest of the copyright holder anyways.
Yeah, definitely not a legal issue, only potentially a matter of ideological purity of the builds.

LAGonauta:
Just rebased and added some changes for proper A/V on Windows.

On Windows 10 it is possible to have extremely low audio latency, for example:
```bash
 gst-launch-1.0 -v wasapisrc ! audioamplify amplification=20.0 clipping-method=wrap-positive ! audioresample ! wasapisink low-latency=true
```

But I couldn't reproduce that in code yet, therefore Dino will just use DirectSound. Latency is comparable to Google Meet, so I think it is fine. With WASAPI latency should be similar to Mumble.

wurstsalat3000:
@LAGonauta I'm not sure whether colored emoji rendering on Windows already works for you in these builds. Since [1.17.6](https://gitlab.freedesktop.org/cairo/cairo/-/tags/1.17.6), cairo supports rendering of fonts with colored emojis on Windows. I just figured out how to enable this for Gajim, and I thought I'd let you know:

Set the `PANGOCAIRO_BACKEND` environment variable to `fontconfig` (default is `win32`).

Yuubi-san:
Okay, now I'm hyped too... But no windows around to try it out :D

driver1998:
C++/WinRT Upstream is working on improving MinGW support, and should already work with GCC 12 and LLVM 15.
I don't think YoloRT would be necessary anymore.

see:
https://github.com/mstorsjo/llvm-mingw/issues/307
https://github.com/alvinhochun/mingw-w64-cppwinrt

LAGonauta:
Hello folks!
I updated my branch to latest Dino, added Meson build support, and added a build pipeline to build Dino though both GCC and Clang.

Some problems:
1. webrtc-audio-processing-1 is not being detected in the build pipeline for some reason, but it is installed. It is detected correctly locally, though... This still needs investigation.
2. glib-2.0 try_parse does not return the default value when parsing invalid numbers, which means one of the Stanza tests fails. Still needs investigation. As I wrote in the MUC: `On Windows, uint.try_parse of -42 returns 4294967254 instead of the default value 3... so it overflows instead of failing to parse`

I've been daily driving the GCC CMake build for some time now and it seems to be working fine, I should daily drive the meson build now.

Neustradamus:
To follow this good @LAGonauta PR :)

Neustradamus:
@LAGonauta: Have you seen this @Nopey PR?
- https://github.com/dino/dino/pull/1759

PR #855, opened 2020-05-20
Global search: Display dates in local timezone

selurvedu:
This fixes message dates being displayed in UTC in the message search sidebar.


fiaxh:
Thanks.

PR #872, opened 2020-07-02
Fix messages mistakenly treated as /me command

Zash:
To reproduce: Send a message like `/message` or anything that starts with `/me` not followed by a space.

This results in something like:

> **Nick**ssage

Per XEP-0245 only messages that start with "/me " (with the trailing space) should treated as 3rd person actions.


fiaxh:
Thanks for fixing.

PR #883, opened 2020-07-17
Add different formating with less contrast for markup strings

### LINKED ISSUE #704: Message styling: Reduce constrast slightly of styling commands

minils:
Fixes #704
I did not find a global color for unfocused elements. Therefore, I just used `#999999` which looks nice with both the light and the dark theme:
dark theme
light theme

Probably it would be nicer to use the color codes from the material palette which would be `#9E9E9E` and `#E0E0E0`.
What do you think?

### DIFF
--- a/main/src/ui/util/helper.vala
+++ b/main/src/ui/util/helper.vala
@@ -381,7 +381,9 @@ public static string parse_add_markup(string s_, string? highlight_word, bool pa
                     int start, end;
                     match_info.fetch_pos(2, out start, out end);
                     return parse_add_markup(s[0:start-1], highlight_word, parse_links, parse_text_markup, already_escaped) +
-                        s[start-1:start] + @"<$(convenience_tag[i])>" + s[start:end] + @"</$(convenience_tag[i])>" + s[end:end+1] +
+                        "<span color='#999'>" +  s[start-1:start] + "</span>" +
+                        @"<$(convenience_tag[i])>" + s[start:end] + @"</$(convenience_tag[i])>" +
+                        "<span color='#999'>" + s[end:end+1] + "</span>" +
                         parse_add_markup(s[end+1:s.length], highlight_word, parse_links, parse_text_markup, already_escaped);
                 }
             } catch (RegexError e) {


licaon-kter:
This will override my theme? Please don't

Did you test with dark themes?

iNPUTmice:
> I did not find a global color for unfocused elements.

Can you not use text color at a different opacity? (That’s what Conversations is doing)

fiaxh:
Thanks for the PR.

minils:
So I somehow forgot about this PR. 
I still miss a global definition of the color. (Regardless of it being gray or a reduced opacity.)

Where are other colors defined?

mimi89999:
> Where are other colors defined?

They are defined in the system GTK theme. That's why it's bad do set arbitrary colors in apps.

main/src/ui/util/helper.vala
```diff
@@ -381,7 +381,9 @@ public static string parse_add_markup(string s_, string? highlight_word, bool pa
                     int start, end;
                     match_info.fetch_pos(2, out start, out end);
                     return parse_add_markup(s[0:start-1], highlight_word, parse_links, parse_text_markup, already_escaped) +
-                        s[start-1:start] + @"<$(convenience_tag[i])>" + s[start:end] + @"</$(convenience_tag[i])>" + s[end:end+1] +
+                        "<span color='#999'>" +  s[start-1:start] + "</span>" +
```
mimi89999, line 384: Please don't override colors from the theme. While unlikely, it's possible that users might want to use a theme where fg text has a different color like green. In that case, the grey symbols will look strange. I recommend it either use `alpha='50%'` or `font_weight='ultralight'` (you can adjust the values to your needs). Markup reference: https://developer.gnome.org/pango/stable/pango-Markup.html

minils:
I changed the formatting to `alpha=30%`:

grafik
grafik

See PR: https://github.com/dino/dino/pull/1000

PR #902, opened 2020-09-05
Fix hangs resending old messages

alyssarosenzweig:
If Dino was closed while attempting to send a message, and opened (more
than 10 seconds) later, Dino will correctly attempt to resend the
message. However the resending will incorrectly recurse infinitely from
the signal connected in update_label(), despite no actual changes,
causing Dino to hang.

By checking that the status actually changed, we don't go down this code
path more than once, preventing the recursion and fixing the hang. The
message is still resent and status updated as expected.

Fixes: 7309c6f3 ("Visually highlight pending messages, improve resending")

Cc @fiaxh

### DIFF
--- a/main/src/ui/conversation_content_view/message_widget.vala
+++ b/main/src/ui/conversation_content_view/message_widget.vala
@@ -193,7 +193,9 @@ public class MessageItemWidget : SizeRequestBin {
 
                 message.bind_property("marked", this, "marked");
                 this.notify["marked"].connect(() => {
-                    update_label();
+                    /* Only update when the state actually changes so we don't recurse infinitely */
+                    if (message.marked != Message.Marked.SENDING && message.marked != Message.Marked.UNSENT)
+                        update_label();
                 });
             } else {
                 int time_diff = (- (int) message.local_time.difference(new DateTime.now_utc()) / 1000);


fiaxh:
Thanks for discovering and investigating this :)

I don't think calling `update_label()`in the wrong moment can actually cause an infinite recursion in that sense.

What I believe happens is that - for some reason - the `marked` property of the message gets changed a (high) number of times before the message finally gets send. At first, only a single listener for the `marked` property is registered. When the `marked` property changes (however the message was not sent), another listener ist registered. Two listeners become 4, 4 become 8 etc. And after some changes of `marked` this will probably lead to Dino freezing.

Unregistering the listener and the binding is something we should do anyways. I pushed the unregistering code to the `fiaxh/unregister_pending_handler` branch (in 8cf5e172d7aa65d1f96ce5c5fd2eac02a513239c). Since I can't reproduce the hanging / "often calling", could you please check if this also solves the hanging? However, even if it fixes there is still the question of why the `marked` state changes so often...

fiaxh:
The respective commit (78138bf39b2da8e11983f4c4e36e808ec9ebeda4) was added to master a long while ago, and I assume it fixed this issue. But please leave a message if you should still have issues.

PR #919, opened 2020-09-24
Add travis jobs on ppc64le

dineshks1:
This PR adds power support on travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,9 @@
 language: generic
 
+arch:
+ - amd64
+ - ppc64le
+
 services:
   - docker
 


fiaxh:
For what reason do you think Dino would benefit from test-building for another architecture, specifically this one?

dineshks1:
This is a part of Ubuntu distribution for ppc64le architecture. It will help us to simplify testing later when distributions are re-building and re-releasing.

dineshks1:
Please review this merge.

fiaxh:
Travis CI apparently changed their policies leaving us with a very limited number of build minutes per month. Until we change away from Travis, we'll use those minutes to check if the code runs and tests succeed. And not for building on multiple architectures.

We are building for ppc64le on OBS ([click](https://build.opensuse.org/package/live_build_log/network:messaging:xmpp:dino/dino/Fedora_33/ppc64le)), so any build issues can be noticed there.

PR #927, opened 2020-10-11
Added more public servers

dimethyltriptamine:
In this commit I've added more public XMPP servers to the account creation dialog:
* dismail.de
* intux.de
* parloteo.es

### DIFF
--- a/main/src/ui/manage_accounts/add_account_dialog.vala
+++ b/main/src/ui/manage_accounts/add_account_dialog.vala
@@ -64,10 +64,13 @@ public class AddAccountDialog : Gtk.Dialog {
 
     private static string[] server_list = new string[]{
         "5222.de",
+	"dismail.de",
+	"intux.de",
+	"parloteo.es",
         "jabber.fr",
         "movim.eu",
         "yax.im"
-    };
+};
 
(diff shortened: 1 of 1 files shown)


licaon-kter:
Why those?

qorg11:
Picked randomly from compliance.conversations.im and they have a good privacy policy

mimi89999:
Could you please add them to https://invent.kde.org/melvo/xmpp-providers ?

fiaxh:
I feel that picking servers for this list is a rather difficult thing.
Of course we ideally want servers that support all the XMPP features that Dino supports. They should have a reasonable privacy policy. They should have a good record of uptime. Those three things can easily be checked.

However, there are also aspects that are more difficult to estimate: We want servers that won't go offline anytime soon. We want servers that won't close registration anytime soon (We can't update the Dino version in Debian Stable / Ubuntu, and they stay around for more than two years). We want servers that are maintained by multiple people.

For example, from your PR dismail.de closed registrations by now.

We'll need a better solution for our server suggestions. Until we found one, I don't see us adding servers from PRs. If you have a strong case for one specific server, please feel free to mention it in our XMPP channel.

PR #936, opened 2020-10-30
Swedish translations

Zash:
Fills in the untranslated strings.

(Sorry about not using weblate)

### DIFF
--- a/plugins/omemo/po/sv.po
+++ b/plugins/omemo/po/sv.po
@@ -19,7 +19,7 @@ msgstr ""
 
 #: plugins/omemo/src/ui/bad_messages_populator.vala:141
 msgid "Your contact"
-msgstr ""
+msgstr "Din kontakt"
 
 #: plugins/omemo/src/ui/bad_messages_populator.vala:152
 #, c-format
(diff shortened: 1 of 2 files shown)


fiaxh:
Thanks. There are also two Swedish sentences missing in the OMEMO .po file [1](https://github.com/dino/dino/blob/master/plugins/omemo/po/sv.po#L29) [2](https://github.com/dino/dino/blob/master/plugins/omemo/po/sv.po#L39), in case you feel like translating them, too.

Zash:
There, @fiaxh

plugins/omemo/po/sv.po
```diff
@@ -37,6 +37,7 @@ msgstr "Hantera enheter"
 msgid ""
 "%s does not trust this device. That means, you might be missing messages."
 msgstr ""
+"%s litar inte på den här enheten. Du kanske inte ser alla meddelanden."
```
Zash, line 40: "You might not see all messages" instead of "missing messages", hope that works. Not sure if I have seen this in the UI myself.

PR #948, opened 2020-11-27
Change contact details dialog container to HdyClamp and enable label wrap

mimi89999:


mimi89999:
It would be good to click/toggle the button in the row when the row is clicked, but I don't know how to implement something like that.

genofire:
wow nice work, just Account and Emoji-Selector are not responsible yet - but everthink else works fine:
- Settings
- About
- List of "Start Converation"
- List of "Join Channel"

PR #950, opened 2020-11-30
Fix wrong actions for MUCs in Freedesktop notifier

LAGonauta:
While developing a Windows notification plugin I noticed that the FreeDesktop backend was calling the wrong actions. This is the fix.


fiaxh:
Thanks!

PR #961, opened 2020-12-10
Add window controls to conversation titlebar CSD in folded mode

mimi89999:


fiaxh:
As was already mentioned in the project XMPP channel, this change has the downside that there is even less space available in the header bar, especially with regard to the conversation title. I will merge this for now, but it might be a point for future changes.

PR #987, opened 2021-02-11
Remove 5222.de and movim.eu from add_account_dialog

mimi89999:
They disabled IBR


fiaxh:
Those two servers also didn't allow signing up directly via IBR at the time they were added to the list. Instead, they use the redirection mechanism specified in XEP-0077 to provide a URL under which the user can sign up. Dino can handle that case and displays the respective link to the user. Some server admins don't want to enable signing up via IBR for spam-, legal- or other reasons. While IBR might be preferable from a client-perspective, I think it's fine to leave those two servers in the list for now as they provide IBR redirection information and appear maintained.

Issue #988, opened 2021-02-11
lightwitch.org IBR text not shown


mimi89999:
In Dino:
![Capture d’écran du 2021-02-11 19-43-19](https://user-images.githubusercontent.com/8530546/107684681-5bc46a00-6ca3-11eb-847c-a535bcad822d.png)

In Gajim:
![Capture d’écran du 2021-02-11 19-43-50](https://user-images.githubusercontent.com/8530546/107684730-6848c280-6ca3-11eb-8e86-c26616aa32a2.png)

Form:
```
<iq xmlns="jabber:client" from="lightwitch.org" id="805b51e8-6cb9-41c5-bdcb-717c01214b7f" type="result">
  <query xmlns="jabber:iq:register">
    <instructions>Choose a username and password for use with this service.</instructions>
    <username />
    <password />
    <email />
    <x xmlns="jabber:x:data" type="form">
      <title>Creating a new account</title>
      <instructions>Choose a username and password for use with this service. Also supply your E-Mail address for verification.</instructions>
      <field var="FORM_TYPE" type="hidden">
        <value>jabber:iq:register</value>
</field>
      <field var="username" type="text-single" label="Username">
        <required />
</field>
      <field var="password" type="text-private" label="Password">
        <required />
</field>
      <field var="email" type="text-single" label="Email" />
      <field type="fixed">
        <value>By registering you implicitly accept our terms of service: https://lightwitch.org/im-service/tos</value>
</field>
</x>
</query>
</iq>
```

PR #990, opened 2021-02-12
Improve IRB form handling

mimi89999:
Change data forms to handle title and instructions elements

https://xmpp.org/extensions/xep-0004.html#protocol

### DIFF
--- a/main/src/ui/manage_accounts/add_account_dialog.vala
+++ b/main/src/ui/manage_accounts/add_account_dialog.vala
@@ -340,7 +340,8 @@ public class AddAccountDialog : Gtk.Dialog {
             register_form_continue.grab_focus();
         } else if (form.fields.size > 0) {
             if (form.instructions != null && form.instructions != "") {
-                form_box.add(new Label(form.instructions) { halign=Align.CENTER, xalign=0, margin_top=7,
+                string markup_instructions = Util.parse_add_markup(form.instructions, null, true, false);
+                form_box.add(new Label(markup_instructions) { use_markup=true, halign=Align.CENTER, xalign=0, margin_top=7,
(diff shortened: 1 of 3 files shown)


main/src/ui/manage_accounts/add_account_dialog.vala
```diff
@@ -339,6 +339,14 @@ public class AddAccountDialog : Gtk.Dialog {
             register_form_continue.visible = true;
             register_form_continue.grab_focus();
         } else if (form.fields.size > 0) {
+            if (form.title != null && form.title != "") {
```
fiaxh, line 342: With this change, you display two titles: ours and the one chosen by the server. Please pick only one. The title goes into `register_title.label`.

fiaxh (changes requested):

main/src/ui/manage_accounts/add_account_dialog.vala
mimi89999: I know. I want to display the title from the server because it might be important for the registration. I also don't want it to control the entire content of the window. This is intended and I did like Gajim did as can be seen in the screenshot in https://github.com/dino/dino/issues/988.

fiaxh: I understand both of your points. However, I think that showing both titles is not a satisfactory solution to the problem, since in most cases one title will be redundant.
While I don't have a good solution myself right now, I don't feel this change is an improvement to the status quo. My suggestion would be to display the server-instructions, but not the server-title for now. That way, the PR improves the current status without making anything worse.

fiaxh (approved):

fiaxh:
Thanks.

PR #1000, opened 2021-02-18
Use opacity instead of hard coded color for formatting chars

minils:
Like the title says.

### DIFF
--- a/main/src/ui/util/helper.vala
+++ b/main/src/ui/util/helper.vala
@@ -308,9 +308,9 @@ public static string parse_add_markup(string s_, string? highlight_word, bool pa
                     int start, end;
                     match_info.fetch_pos(2, out start, out end);
                     return parse_add_markup(s[0:start-1], highlight_word, parse_links, parse_text_markup, already_escaped) +
-                        "<span alpha='40%'>" +  s[start-1:start] + "</span>" +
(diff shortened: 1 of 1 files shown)


main/src/ui/util/helper.vala
```diff
@@ -299,9 +299,9 @@ public static string parse_add_markup(string s_, string? highlight_word, bool pa
                     int start, end;
                     match_info.fetch_pos(2, out start, out end);
                     return parse_add_markup(s[0:start-1], highlight_word, parse_links, parse_text_markup, already_escaped) +
-                        "<span color='#999'>" +  s[start-1:start] + "</span>" +
+                        "<span alpha='30%'>" +  s[start-1:start] + "</span>" +
```
mimi89999, line 302: Wasn't the color previously more like 40%? I think that 30% might be too light for some users and they might not see them.

mimi89999: @minils any update on this?

minils:
Changed from `30 %` to `40 %`

mimi89999 (approved):

licaon-kter:
@minils why is this touching so many files?

minils:
Sorry I was trying to rebase on master but something went wrong...

Issue #797, opened 2020-04-06
Password field shouldn't indicate password length


guilieb:
Hi, 

When connecting using an account which is protected by a very long password, the window with is too large. As well, for security reasons, this password field should not give any indication on the password length.

![Capture d’écran du 2020-04-06 11-47-39](https://user-images.githubusercontent.com/49333734/78545670-749afe00-77fc-11ea-837d-f22631a2d37e.png)

```
Installed Packages
Name         : dino
Version      : 0.1.0
Release      : 1.fc31
Architecture : x86_64
Size         : 3.8 M
Source       : dino-0.1.0-1.fc31.src.rpm
Repository   : @System
From repo    : updates
Summary      : Modern XMPP ("Jabber") Chat Client using GTK+/Vala
URL          : https://github.com/dino/dino
License      : GPLv3
Description  : A modern XMPP ("Jabber") chat client using GTK+/Vala.
```

fiaxh:
#1017 fixed the window width problem, however I'll keep this issue opened for the second point that you mentioned: the password length indication.

eerielili:
@guilieb I pushed something that might fix it.
Tested it on my side and it works well. The password length is only ever revealed when you update your password by typing/copy pasting. Once you reopen the dialog the obsfucation is back again.

fiaxh:
This is solved as part of https://github.com/dino/dino/commit/21ae42762d8a57da5cb1ec40b46e7510fc3121ad, as the field no longer exists in that form. Now, passwords are only displayed (in dots) when the user enters them.

PR #1017, opened 2021-03-09
Fix accounts panel width (#797)

sshikaree:


fiaxh (changes requested):
Thanks for the PR.
I think it would be suitable to put those changes directly into the [`LabelHybrid`](https://github.com/dino/dino/blob/9cc3382abe801f0be96b79dcfc9cd4ac0272befa/main/src/ui/util/label_hybrid.vala#L11) (the parent of `EntryLabelHybrid`). That way, we fix the "widget gets to large and messes up window dimensions" issue for all usages of the `LabelHybrid`.

sshikaree:
Maybe it would be more consistent to add this changes [here](https://github.com/dino/dino/blob/9cc3382abe801f0be96b79dcfc9cd4ac0272befa/main/src/ui/util/label_hybrid.vala#L8) in properties list?

fiaxh:
@sshikaree Sounds good :+1:

fiaxh (approved):

PR #1018, opened 2021-03-09
Make 'Close' button close AboutDialog

sshikaree:


fiaxh:
For me, the "x" button already closes the dialog fine.
But looking at pictures in the documentation, there might exist two versions of the AboutDialog. Mine looks like this:
Screenshot from 2021-03-09 17-40-05
But apparently there also exists (or existed) another version like this:

Does your AboutDialog look like the second one? What GTK version are you using?

sshikaree:
And this is mine:
Dino About window
GTK version 3.24.20, Xubuntu 20.04.2.

PR #1035, opened 2021-04-10
Add new config item for OMEMO encryption over no encryption

XaviDCR92:
So far, when new contacts are added into an account, an unencrypted
conversation is started by default, so explicit user intervention is
needed if OMEMO or OpenPGP encryption is required.

As expected, I always forget that little manual intervention, so this
commit adds a new configuration item that forces both new and existing
conversations to use OMEMO encryption automatically if no other
encryption method has been selected.

No actions are performed if other encryption methods (e.g.: OpenPGP)
are already in use on a given conversation.

TODO: OMEMO encryption makes little sense in public chats. Test this
condition.

Suggestion: needless to say, default OMEMO encryption is even more
desirable for non-tech users who would otherwise never configure an
encryption method for themselves. Should this new configuration item
be enabled by default on all new installations?


licaon-kter:
This is better than #845 or #632? How? Pics?

XaviDCR92:
I was not aware of duplicate pull requests, sorry. I wrote this commit since I was surprised Dino uses no encryption by default, and it was a good excuse to learn Vala, anyway.
However, IMHO having various duplicate pull requests shows this is an important feature for users and should be merged into upstream.

licaon-kter:
Do read the associated issue and PRs, maybe it explains why not yet.

XaviDCR92:
I have read the PRs mentioned above and #632 differs slightly in its implementation, whereas #845 does not introduce any UI configuration. #632 would be fine for me, anyway, but for some unknown reason some contributors are not accepting it, probably because they prefer OpenPGP instead.
Honestly, I prefer OMEMO for its ease of use and availability on multi-user private chats, as well as its advantages [over OpenPGP](https://conversations.im/omemo/). But, even if Dino preferred OpenPGP by default for some reason, that's still a lot better than status quo.
IMHO, advertising Dino as ["secure by default: Your chats never leave your computer unencrypted."](https://dino.im/) is a misleading statement, to say the least. Leaving chats unencrypted by default is unacceptable according to today's standards and a solution should be taken into account ASAP, whether it's OMEMO or OpenPGP.

XaviDCR92:
Another suggestion is to replace `Encryption enum`

https://github.com/XaviDCR92/dino/blob/a7602cee8152c99261445d5b9a095a931ffbc040/libdino/src/entity/encryption.vala#L3-J7

with the following values:

```vala
public enum Encryption {
        UNKNOWN,
        PGP,
        OMEMO,
        NONE
}
```

New and unencrypted conversations will then be automatically switched from `NONE` to `UNKNOWN`. Then, if `Prefer OMEMO encryption` is selected by the user, conversations with `UNKNOWN` encryption are moved to `OMEMO`. On the other hand, conversations explicitly unencrypted are set to `UNENCRYPTED` and will never change to `OMEMO` unless the user explicitly does so.

If `Prefer OMEMO encryption` is **not** selected, then either a selection dialog could be displayed or (less reasonably) use no encryption but still keep the encryption method as `UNKNOWN`, until a method is explicitly set by the user.

albjeremias:
Why not ask for every account created which encryption they prefer? So depending on the account the user may have different default encryption.

But I hope we all agree that no encryption by default is terrifying.

XaviDCR92:
Superseded by https://github.com/dino/dino/pull/1236.

PR #1076, opened 2021-06-18
Fixed name change

### LINKED ISSUE #712: Changing name of roster contact not updated
If I change the name of a contact in my roster with dino it is everywhere updated in the UI except in the top center of the current chat. I need to click another chat and click on the chat before again to get it updated there.

sojuz151:
Fixes #712 and another similar bug.

### DIFF
--- a/main/src/ui/conversation_content_view/conversation_item_skeleton.vala
+++ b/main/src/ui/conversation_content_view/conversation_item_skeleton.vala
@@ -114,7 +114,7 @@ public class ItemMetaDataHeader : Box {
     public Entities.Message.Marked item_mark { get; set; }
     private ArrayList<Plugins.MetaConversationItem> items = new ArrayList<Plugins.MetaConversationItem>();
     private uint time_update_timeout = 0;
-    private ulong updated_roster_handler_id;
+    private ulong updated_roster_handler_id = 0;
 
     public ItemMetaDataHeader(StreamInteractor stream_interactor, Conversation conversation, Plugins.MetaConversationItem item) {
         this.stream_interactor = stream_interactor;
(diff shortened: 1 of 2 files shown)


main/src/ui/conversation_content_view/conversation_item_skeleton.vala
```diff
@@ -123,6 +123,11 @@ public class ItemMetaDataHeader : Box {
 
         update_name_label();
         name_label.style_updated.connect(update_name_label);
+        stream_interactor.get_module(RosterManager.IDENTITY).updated_roster_item.connect((account, jid, roster_item) => {
```
fiaxh, line 126:
```suggestion
        updated_roster_handler_id = stream_interactor.get_module(RosterManager.IDENTITY).updated_roster_item.connect((account, jid, roster_item) => {
```

You connect a listener to the `RosterManager`, but you're never detaching it. That means the code can be called even after the `ConversationItemSkeleton` has been destroyed.
Connecting the listener will return a handler id, which you can use to detach it afterwards. The needed code is similar to how it's done with the `time_update_timeout` in the same file.

fiaxh (changes requested):

main/src/ui/conversation_content_view/conversation_item_skeleton.vala
sojuz151: I added that in the new commit. 

Small side not:
In this case (with conversation replaced with this.conversation) you don't need too disconnect a signal. In the generated c code you get something like  
```
_tmp8_ = g_signal_connect_object (_tmp7_, "updated-roster-item", (GCallback) ___lambda4__dino_roster_manager_updated_roster_item, self, 0);
self->priv->updated_roster_handler_id = _tmp8_;
```
but g_signal_connect_object disconnects signal automatically but such behavior is not documented, possible a bug and easy to break.

fiaxh:
Thanks!

PR #1077, opened 2021-06-19
Small notification fixes

sojuz151:
- Added desktop-entry to call notification

- If notification server supports body_hyperlinks  then notifications are send with hrefs. This is part of official spec but right now only plasma supports this.

### DIFF
--- a/main/src/ui/notifier_freedesktop.vala
+++ b/main/src/ui/notifier_freedesktop.vala
@@ -10,6 +10,7 @@ public class Dino.Ui.FreeDesktopNotifier : NotificationProvider, Object {
     private StreamInteractor stream_interactor;
     private DBusNotifications dbus_notifications;
     private bool supports_body_markup = false;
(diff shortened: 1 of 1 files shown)


fiaxh (changes requested):
Thanks for your PR.

You're calling `Util.parse_add_markup` with `parse_text_markup = true`. However, that feature is designed to convert [XEP-0393](https://xmpp.org/extensions/xep-0393.html) markup into Pango makup. It produces styling directives that are not defined in the [Desktop Notifications Specification](https://specifications.freedesktop.org/notification-spec/notification-spec-latest.html#markup). Thus I'd suggest to either write a function that specifically produces markup for the Desktop Notifications Specification or to just leave message text markup out for now.

Also please use spaces and brackets consistently e.g. don't forget the spaces in `} else {` or `if () {`, and use `else if {}` instead of `else { if {} }`.

fiaxh (approved):

PR #1080, opened 2021-06-25
To add "save as" to right click on file

wkg:


main/src/ui/conversation_content_view/file_widget.vala
```diff
@@ -170,6 +188,18 @@ public class FileDefaultWidgetController : Object {
...
                     }
                 }
+                if (event_button.button == 3) {
+                    var save_dialog = new Gtk.FileChooserDialog ("Save as file", this as Gtk.Window, Gtk.FileChooserAction.SAVE, Gtk.Stock.CANCEL, Gtk.ResponseType.CANCEL, Gtk.Stock.SAVE, Gtk.ResponseType.ACCEPT);
+                    save_dialog.set_do_overwrite_confirmation (true);
+                    save_dialog.set_modal(true);
+                    try {
+                        (save_dialog as Gtk.FileChooser).set_file (GLib.File.new_for_path (GLib.Path.build_filename (file_uri.substring(7))));
+                    } catch (GLib.Error error) {
+                        warning("Faild to open save dialog: %s\n", error.message);
```
licaon-kter, line 198: type `failed`

wkg: fixed

fiaxh:
Thanks for your PR, this is indeed a much-requested feature! :slightly_smiling_face:

Right-clicking on a widget to open the save file dialog is rather unintuitive. It's useful for people that know about it, but the rest isn't going to find out about that feature.

Our idea was to add a `MenuButton` to the right site of the `FileWidget`. When you click the button, it would display a menu that offers actions that can be done with the file, like "Save". Clicking "Save" would then open the `FileChooserDialog`. 

In our code, the [`CallButton`](https://github.com/dino/dino/blob/master/main/src/ui/conversation_titlebar/call_entry.vala#L32) is an example for such a `MenuButton`. For the menu we use a [`PopoverMenu`](https://github.com/dino/dino/blob/master/main/src/ui/conversation_titlebar/call_entry.vala#L46) and set it to the `popover` property of the `MenuButton`. "Save" would be a [`ModelButton`](https://github.com/dino/dino/blob/master/main/src/ui/conversation_titlebar/call_entry.vala#L37).

gjedeer:
Anecdotal evidence about the intuitiveness - I've found this issue because I was confused by the file download thing not having a right-click menu.

fiaxh:
@gjedeer While a right-click menu isn't what we'd like to go for (as described above), I agree that a right-click menu would be a rather expectable user interaction. However, this PR does not add a right-click menu. Instead, it directly opens the file-save-dialog on right-click.

@wkg Would you be interested in adding a menu as described above? I'll be happy to answer any potential questions about the idea and the related Dino code.

wkg:
> @wkg Would you be interested in adding a menu as described above? I'll be happy to answer any potential questions about the idea and the related Dino code.

I managed to place the button, but I did not understand how to call the menu
sJi27fgOj26gIz2DEY5QEzrVxQPNZVwo

fiaxh:
Here's roughly how you can create a `PopoverMenu` with a "Save file" entry and add it to the button that you added to the widget.

```vala
Gtk.MenuButton menu_button = new Gtk.MenuButton() { use_popover=true, visible=true }; // Your menu button

// Create the content of the menu
Box box = new Box(Orientation.VERTICAL, 0) { margin=10, visible=true };
ModelButton save_button = new ModelButton() { text=_("Save file"), visible=true };
save_button.clicked.connect(() => {
    // Do stuff to save the file
});
box.add(save_button);

// Create the menu and add it to the button
Gtk.PopoverMenu popover_menu = new Gtk.PopoverMenu();
popover_menu.add(box);
menu_button.popover = popover_menu;

// Show the menu on click
menu_button.clicked.connect(() => {
    popover_menu.visible = true;
});
```

You can also find an example of how it could be done in our [code](https://github.com/dino/dino/blob/master/main/src/ui/conversation_titlebar/call_entry.vala#L37-L69).

wkg:
@fiaxh Thank you. 
I made it, but new problem.
I'm turn off left click to file, so that it does not open with the menu:
```
            case FileTransfer.State.COMPLETE:
                return false;
```
But it didn't work with image files, the menu would show for one second and hide ...

wkg:
Turned off the left click so that the file does not open with the menu.
For pictures, I turned off the hiding of the widget, otherwise the menu also disappears.

wkg:
Problem with images, i don't know how it fix...
xTRzr25wVOx5jYSeD1eujxK6H6HO9bRx

fiaxh:
I added a commit to `master` that gets rid of the error-prone mouse enter/leave tracking via the `pointer_inside` variable (c85945ae26cab7eb65c67dbe49ebbf668a028b1c). That should also make some things for this PR easier :slightly_smiling_face: Thus, please rebase your work onto the current `master`.

A few additional remarks:

- You create the `MenuButton` in the `FileDefaultWidgetController`. However, the controller is rather ment to do the logic and interact with `libdino`. The UI should be created in the `FileDefaultWidget`.
- I think we should show the menu button only on hover to keep the user interface cleaner. That can be easily done by setting it to (in)visible in the on_pointer_entered/left functions.
- Please set the MenuButton to have `relief` `none`.
- "Open dir" + "Open save dir": I don't see much value in those functions, so I'd like to remove them. The default save location is a confusing place and the directory that was globally last saved to will probably be irrelevant for that specific file.
- "Save file": Please rename to "Save as…", which seems to be a commonly used description of that funtion. Also, please rename "Open file", to just "Open" for the same reason.
- The formating is rather inconsistent. This is partially because you sometimes use tabs, while we use spaces. Please also insert spaces e.g. in "`visible == false`", "`if (`" and "`try {`".

wkg:
better hope now?

sonicnkt:
This has nothing directly todo with this pr but since we might have a "file action" button soon, a function to share this file with another chat would be a nice addition (maybe also an mail option)?

gjedeer:
> we might have a "file action" button soon, a function to share this file with another chat would be a nice addition

I'm not sure how stupid this would be security-wise, but the ability to share a file without having to re-upload it would be so great! (just repost the same HTTPS / AESCGM link to another chat)

sonicnkt:
> > we might have a "file action" button soon, a function to share this file with another chat would be a nice addition
> 
> I'm not sure how stupid this would be security-wise, but the ability to share a file without having to re-upload it would be so great! (just repost the same HTTPS / AESCGM link to another chat)

Well this would only work in public chats tho right? Private (OMEMO) wouldnt work this way. I think a reupload wouldnt be an issue tho. Its a bit faster/easier than have to locate the local file and copy it back to dino).  Maybe something like drag an drop would also be a good option, just drag the attachment onto another chat in the sidepanel. 

@wkg: thats true, as well as reply :) But maybe this should user another ui.

gjedeer:
> Well this would only work in public chats tho right? Private (OMEMO) wouldnt work this way

I do it all the time in Conversations, long-press the file, copy URL, paste into another chat. Works OK for aescgm links (because why wouldn't it - it's just a HTTP link with a symmetric encryption key attached).

Not that I suggest doing it without research - I don't know the security implications.

fiaxh:
Thanks for your work, I merged it to master. There were still a couple of issues left which I fixed in c5461b8536b6171aea17d0e97d41710988b73534.

PR #1082, opened 2021-07-04
Added previews

sojuz151:
Added support for generating previews for received links, including using the open graph protocol. 
Screenshot_20210704_135723

### DIFF
--- a/libdino/src/entity/settings.vala
+++ b/libdino/src/entity/settings.vala
@@ -12,6 +12,7 @@ public class Settings : Object {
         notifications_ = col_to_bool_or_default("notifications", true);
         convert_utf8_smileys_ = col_to_bool_or_default("convert_utf8_smileys", true);
         check_spelling = col_to_bool_or_default("check_spelling", true);
+        auto_preview_ = col_to_bool_or_default("auto_preview", false);
     }
 
     private bool col_to_bool_or_default(string key, bool def) {
@@ -76,6 +77,18 @@ public class Settings : Object {
                 .value(db.settings.value, value.to_string())
                 .perform();
             check_spelling_ = value;
+        }
+    }
+
+    private bool auto_preview_;
+    public bool auto_preview {
+        get { return check_spelling_; }
+        set {
+            db.settings.upsert()
+                .value(db.settings.key, "auto_preview", true)
+                .value(db.settings.value, value.to_string())
+                .perform();
+                auto_preview_ = value;
         }
     }
 }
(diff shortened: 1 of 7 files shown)


ericschdt:
IMHO privacy and security wise it would be better if the sender generated the previews (on demand). Do you also check the file size somewhere before downloading any image?

licaon-kter:
@ericschdt so the sender can lie? Not sure if serious

fiaxh:
Thanks for your PR.

I agree with @ericschdt. Previews should be generated by the sender and attached to the message. Automatically sending requests to arbitrary URLs leaks the users IP to that service and might even cause legal issues for the user depending on the content of the website and the jurisdiction the user is under.

>  so the sender can lie?

- Even with receiver-generated previews it is well possible to generate misleading previews. The server can serve different site contents depending on whether the request comes from a browser or another program/library. That way, the preview will suggest another content than what the user will see in the browser afterwards.
- The sender of a link can already lie about the contents. For example by sending a normal message with a claim about the link content. Or by sending a link that looks like a trustworthy website on first glance.

sojuz151:
> IMHO privacy and security wise it would be better if the sender generated the previews (on demand). Do you also check the file size somewhere before downloading any image?

I added size checking in the new commit.

>Automatically sending requests to arbitrary URLs leaks the users IP to that service and might even cause legal issues for the user depending on the content of the website and the jurisdiction the user is under.

The default behaviour is to just show a "Download Preview" button without making any network request. Users can enable automatic previews in the setting.

fiaxh:
> The default behaviour is to just show a "Download Preview" button without making any network request. Users can enable automatic previews in the setting.

> Privacy issues aside (just disable it by default...)

That feels like we'd be providing the user with the choice "You can have EITHER privacy OR a good UX". However, combining a good UX with privacy-friendly technology is exactly what Dino is about :slightly_smiling_face: I think that previews are a relevant feature, but I don't see us merging a preview feature in which the receiving client requests the URL.

Unfortunately, XMPP currently does not provide a good solution for previews. In my opinion, the best approach would be to have the sending client generate a preview and send it along with the message. The format of the sent preview should be standardized in a XEP such that any client can implement it.

cthu1hoo:
If we look back on how other XEPs like MAM (written, what, 10 years ago?) are being implemented, I think it's rather more likely that I would grow old and die before this hypothetical preview XEP you're describing would gain any sort of client adoption.

I suppose Gajim continues to be the only game in town until XMPP ecosystem dies out entirely. :shrug:

licaon-kter:
@cthu1hoo why the drama?

Wasn't this an usecase for https://xmpp.org/extensions/xep-0385.html ?

Others are doing this: https://www.mysk.blog/2020/10/25/link-previews/ 

So which of those solutions you feel is nicer?

PR #1085, opened 2021-08-01
Adds click nick to mention functionality

### LINKED ISSUE #683: To mention someone: Possibility to copy nick from MUC list to the chat
Currently one has to type a nick name manually in order to mention another user. The MUC user list only allows to start a private conversation.

Therefore I suggest the following three options:

1) With a hot key an suggestion and auto completion could be started.
2) A found user's nick in the MUC's search can be put at the cursor position of the input field
3) Clicking on a nick in the chat history should add the nick to the cursor position in the input field

MarcoPolo-PasTonMolo:
Fixes #683

First time using vala so some things might need to be improved.


fiaxh:
I'm not convinced of this feature: Why should a user take their hands off the keyboard, move the mouse to klick on a nick/avatar, put their hands back to keyboard and type a message? Instead, they could just type the nick with the help of auto-completion. Arguably the auto-completion should be further improved e.g. by showing a drop-down list with possible completions (possibly triggered by typing `@`) and the nick should be better readable.

We actually had different plans for what should happen when clicking on the avatar/nick, especially in MUCs: Opening a menu with actions that can be done related to the participant. That menu would include "Start private conversation" and for admins also "Ban" and "Promote to Admin".

MarcoPolo-PasTonMolo:
> Why should a user take their hands off the keyboard, move the mouse to klick on a nick/avatar, put their hands back to keyboard and type a message?
If the user has a nick in another language, say japanese auto completion would be useless. Also it just felt natural, there are plenty of messengers that mention when nickname is clicked.

On that note I do agree about the drop down after `@` and the menu on click also sounds logical. 

A few suggestions:
* Mention when the nick is clicked but open a menu with actions when avatar is clicked.
* Mention when the nick or avatar detect a left mouse button click and open menu on right click
* Remove click to mention and add the option in the menu or on the far right like the edit button

jubalh:
> Why should a user take their hands off the keyboard

Maybe they use dino on a touch display.

PR #1086, opened 2021-08-04
Allow establishing connections in clear text

selurvedu:
Cleartext connection would have the lowest priority, so Dino would still try to connect via StartTLS or direct TLS first.

Allowed hostnames can be specified via the `DINO_CLEARTEXT_HOSTS` environment variable, separated by commas.

By default, only connections to `localhost` are allowed. This can be disabled by setting the variable to an empty value. If the variable gets set to any other value, `localhost` will not be included in the allow list automatically.

### DIFF
--- a/libdino/src/service/connection_manager.vala
+++ b/libdino/src/service/connection_manager.vala
@@ -198,6 +198,7 @@ public class ConnectionManager : Object {
 
             change_connection_state(account, ConnectionState.CONNECTING);
             stream_result = yield Xmpp.establish_stream(account.bare_jid, module_manager.get_modules(account, resource), log_options,
+                                                        Util.is_cleartext_allowed_for_host(account.bare_jid.domain_jid.to_string()),
                     (peer_cert, errors) => { return on_invalid_certificate(account.domainpart, peer_cert, errors); }
             );
             connections[account].stream = stream_result.stream;
--- a/libdino/src/service/registration.vala
+++ b/libdino/src/service/registration.vala
@@ -30,6 +30,7 @@ public class Register : StreamInteractionModule, Object{
         list.add(new Sasl.Module(account.bare_jid.to_string(), account.password));
 
         XmppStreamResult stream_result = yield Xmpp.establish_stream(account.bare_jid.domain_jid, list, Application.print_xmpp,
+                                                                     Util.is_cleartext_allowed_for_host(account.bare_jid.domain_jid.to_string()),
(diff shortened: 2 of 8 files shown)


fiaxh:
I don't think we should allow clear-text connections for productive every-day usage. Any operator between the user and the server could just read meta-data and unencrypted messages. Server operators should really, really enable SSL. If we would support such insecure setups in Dino, we make it more feasible and attractive to operate them.

Some people might want to allow clear-text connections for a quick test deployment. However, enabling the most simple test setup isn't really the main target of Dino. For those test deployments, it only takes a few additional command line instructions to get a proper setup: Create a self-signed SSL certificate, set the XMPP server to use the certificate and to add the SSL certificate to your system's certificate store.

I see that you aimed to mitigate the dangers by making it an opt-in feature. However, a user might opt-in to clear text connections e.g. due to a social-engineering attack or because they hoped to solve an unrelated issue. All in all, I think this feature brings quite significant issues while only addressing a small inconvenience for a small group of people.

licaon-kter:
Isn't TLS a MUST of the core protocol RFCs? Brr...

selurvedu:
@fiaxh I see your point. Every time I needed to generate snakeoil certs, I also needed to add them to the system cert store and to all services/cotainers/VMs that used it, and then forgot to remove them when they were no longer needed. One day 5 months ago I just thought "how hard it is to add support for clear text connections to Dino?", and it turned out to be not that hard. I kept this code to myself, but yesterday someone mentioned that Dino refuses to connect to unsecured local development server, so I thought someone else may find my code useful, so I edited the patch and submitted it.

I had a strong expectation that it would get declined, so I'm not disappointed at all. Thanks for your feedback. :)

Still, this PR will remain here in case someone else needs it or asks about why clear text is not supported.

> I see that you aimed to mitigate the dangers by making it an opt-in feature. However, a user might opt-in to clear text connections e.g. due to a social-engineering attack or because they hoped to solve an unrelated issue.

That's an interesting point I did not think of. Good catch.

While I was typing this, I had another thought: what if this was a complie-time option and/or what if only localhost was allowed, without a way to modify the allow list? I think it would be a smaller threat then. You could oppose by saying e.g. "a user can still be social-engineered to recompile Dino with that option" – that is true, but tricking someone into manually applying this exact patch (or a more malicious one) wouldn't be much harder then, don't you agree? :)

@licaon-kter huh, even for localhost? 🤨 This may make sense from the standard conformance point of view, but there is no practial need for it as long as it's the same host.

PR #1091, opened 2021-08-13
emojichooser.ui: set category icons on a GtkGrid

XaviDCR92:
Instead of using a 1x10 GtkBox, a 2x5 GtkGrid is now used so used horizontal space is reduced to a half. This is particularly interesting on devices with small screens such as the PinePhone, since the widget would partially fall out of the screen otherwise.

I could not take a screenshot on a PinePhone, so I have simulated this by running desktop Dino at its minimum allowed resolution so you get a similar impression. On a PinePhone, half of the emojis would look out of the screen if not for this patch. OTOH, this patch is not that useful on wider screens, such as the ones on desktops or laptops.

image


XaviDCR92:
For a better insight, this is how it looks like on a mobile screen like the PinePhone:
59c401cc_Screenshot_20210923_074722

fiaxh:
Merged in c848191a034a920266687002045abc5e1f42b070. Note that with GTK4 we'll be able to use [Gtk.EmojiChooser](https://valadoc.org/gtk4/Gtk.EmojiChooser.html), which is now part of the public GTK API.

PR #1097, opened 2021-08-22
Adds a quote button for each message

### LINKED ISSUE #625: Reply/Quote messages
Please add an option to quote / reply to a message

MarcoPolo-PasTonMolo:
Adds a > button over each message which when clicked quotes the message.
Fixes #625

There were 4 options for how the ui should've been implemented, I went with option 2.


christf:
this is nice, thank you for preparing this. I rebased and tested and I can see it working.
One bit in the UI felt a bit awkward:
when hitting the quote button the focus is not automatically transitioned to the input box. so as a user I hit the quote button and then need to navigate my mouse to the text input field, place the cursor and only then can start typing. It would be helpful if hitting the reply button would also do that for me.

serfcity:
I suppose It's good idea.
I want, that this function be added to.

jubalh:
Looks useful to me.

christf:
So.... What next?

XaviDCR92:
I have tested this PR for a long time now and it has always worked well me. @fiaxh (or any other contributors) would you please consider merging it?

fiaxh:
Generally, I think a reply button would be a good addition to Dino. However, I do not see much benefit in a button that merely puts the respective message into the chat input with a ">" in front of it. That's a work-around for actual reply support.

There's a [Reply XEP](https://xmpp.org/extensions/xep-0461.html) currently in the making, but a dependency of that XEP needs to be reworked first before replies can be used in a backwards-compatible fashion. When the specification is finished, I'd be happy for Dino to implement it and then we can have a reply button that actually triggers a reply.

XaviDCR92:
This PR keeps consistency with https://github.com/dino/dino/pull/1111, and follows the same behaviour compared to [Conversations](https://conversations.im), which might be a plus for some users. Even if this proposal is far from perfect, IMHO it is a good enough workaround until the Reply XEP is released.

XaviDCR92:
> That's a work-around for actual reply support.

In fact, this PR is implementing quotation style as described by [XEP-0393](https://xmpp.org/extensions/xep-0393.html#quote).

Quix0r (approved):
How the code is structured, it looks fine. Disclaimer: I don't know Vala and cannot check syntax.

Quix0r:
@MarcoPolo-PasTonMolo You have conflicts here.

MarcoPolo-PasTonMolo:
@Quix0r Done sir

Quix0r (approved):

fiaxh:
We implemented message replies via XEP-0461 in dc52e7595cca06d0a2da7d11b3c88cb2f7ce529c.

Issue #958, opened 2020-12-08
Unable to connect to servers hosted behind Tor Onion services


intrigeri:
Context: we want to replace Pidgin with another XMPP client in [Tails](https://tails.boum.org). Dino is one of the 2 candidates we're currently evaluating. We just got started with the evaluation but so far, this issue is the only clear blocker I've spotted for Dino.

The problem: Onion services have self-signed TLS certificates. As we know, this is a deal breaker for current Dino.

This is a special case of #57 (related to #452 and #209). I think Onion services deserve to be treated separately, in case no mutually agreeable general solution to #57 is found, because contrary to the general case:

  - Validating the certificate is not strictly necessary, and adds little value: by design, Onion services already provide
    transparent authentication of the server by the client lower in the stack.

    This is why, for example, Tor Browser treats Onion websites over plaintext HTTP just the same as HTTPS:
    the connection is considered to be secure.

  - Requiring a CA-signed certificate is a constraint impossible to meet by Onion services: it's
    currently not possible, in practice, to get a CA-signed certificate for an Onion service. Well, Facebook did,
    but they have means that few other XMPP server providers have.

Would you be amenable to handle Onion services in a specific manner regarding TLS certificate validation, in order to solve this problem?

Are there specific implementation or security constraints that can make this costly?

If, on the Tails side, this is the only clear blocker we identify, I would be happy to ask help from our UX designer (@sajolida) wrt. the corresponding user interactions (or lack thereof).

Thanks for considering :)

fiaxh:
Are you referring to accounts that are actually on .onion domains (`user@823urjksj9dfiejsf.onion`) or are you referring to accounts on normal domains (`user@example.com`) that also provide a .onion service for users to connect with?

About the first case (account on .onion domains): It would indeed make sense to allow invalid certificates for .onion domains, since encryption is provided independent of TLS. We can do that.

About the second case: In those cases the .onion server should provide a certificate for example.com, since clients need some way to verify that the server is actually allowed to serve this account.

> I would be happy to ask help from our UX designer (@sajolida) wrt. the corresponding user interactions (or lack thereof).

I believe it will be a simple matter of allowing all certificates for .onion domains without asking the user anything. However if @sajolida has any feedback on that matter or any other part of Dino, we'd be happy to hear it.

Improving Tor support has been on our wishlist anyways, so we'll be happy to work with you Tails people to make Dino suitable to be used within Tails.

intrigeri:
Hi @fiaxh,

[...]

PR #1098, opened 2021-08-25
Reject non-TLS URLs in HTTP File Upload

### LINKED ISSUE #697: OMEMO file transfer: won't send correct URI
I use my own XMPP server, which does not accept https connections to retrieve uploaded files, but it does accepts http connections on port 5280.

When sending a file with OMEMO enabled, dino sends incorrect aesgcm URIs. As a result, other clients cannot retrieve the file content. Instead of the correct URI "aesgcm://hostname.tld/foobar#", dino send "aesgcm://ostname.tld/foobar#" (the first letter of the hostname is missing).

The culprit might be in file dino/plugins/omemo/src/file_transfer/file_encryptor.vala, at the following line:
`aesgcm_link = "aesgcm://" + aesgcm_link.substring(8); // replace https:// by aesgcm://`

When I enable https connections (on port 5281), dino can send files using OMEMO, as expected.

linkmauve:
This is a MUST in the XEP.

Fixes #697.

### DIFF
--- a/xmpp-vala/src/module/xep/0363_http_file_upload.vala
+++ b/xmpp-vala/src/module/xep/0363_http_file_upload.vala
@@ -72,6 +72,11 @@ public class Module : XmppStreamModule {
                 Idle.add((owned) callback);
                 return;
             }
+            if (url_get.ascii_down(4).has_prefix("http://") || url_put.ascii_down(4).has_prefix("http://")) {
+                e = new HttpFileTransferError.SLOT_REQUEST("Error getting upload/download url: server is misconfigured and sent us a URL without TLS enabled");
+                Idle.add((owned) callback);
+                return;
+            }
 
             slot_result.headers = new HashMap<string, string>();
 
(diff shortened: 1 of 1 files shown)


fiaxh:
Thanks for the PR. Your code didn't quite work because `.down_asci(5)` only returns the first 5 characters, however `http://` has 7, thus the check always evaluated to false. I fixed it in an additional commit :wink:

SkewedZeppelin:
Eh, won't this break use of http:// .onions?

fiaxh:
@SkewedZeppelin That's a good point, thanks for bringing it up. I think if .onion XMPP servers provide http:// URLs they'll run into various problems, especially with OMEMO encrypted file transfers, since they mandate https:// and a bunch of clients will likely not additionally try http://. Also, as @linkmauve mentioned, the XEP actually mandates https://. Thus, I think the .onion XMPP servers should actually provide https:// URLs, and use a self-signed certificate. This would be similar to how we handle connections to XMPP servers themselves - we require SSL but allow self-signed certificates for .onion servers (#958).
Do you know of .onion XMPP servers that hand out http:// URLs?

However, this brings up a different point: We should allow self-signed certificates for https:// .onion URLs. I opened an issue for that: #1149

SkewedZeppelin:
@fiaxh 
I don't know of any public ones. I briefly ran one on a Pi years ago for testing a Briar like experience, but there were other issues at the time.
Now that I think about it, I think I still did have a self signed certificate on top.

PR #1111, opened 2021-09-21
Format quotes as preformatted text

XaviDCR92:
This is a little nice feature from Conversations that I had always missed on Dino. A regular expression, based on the one used by [`feature/quote`](https://github.com/MarcoPolo-PasTonMolo/dino/blob/feature/quote/main/src/ui/conversation_view.vala#L27), finds matching quotes. Then, quotes are rendered using the Pango markup language with preformatted text and a specific background colour, while leaving other text with default formatting.
image
As a suggestion, merging https://github.com/dino/dino/pull/1097 along with this PR shall provide Dino with full support for quotes.

Please feel free to suggest any other better-looking format than the one suggested above.

### DIFF
--- a/main/src/ui/util/helper.vala
+++ b/main/src/ui/util/helper.vala
@@ -341,6 +341,19 @@ public static string parse_add_markup(string s_, string? highlight_word, bool pa
                 assert_not_reached();
             }
         }
+
+        Regex regex = new Regex("((?<=\n)&gt;.*(\n|$))|(^&gt;.*(\n|$))");
+        MatchInfo match_info;
+        regex.match(s.down(), 0, out match_info);
+
+        if (match_info.matches()) {
+            int start, end;
+
+            match_info.fetch_pos(0, out start, out end);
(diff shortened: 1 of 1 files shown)


XaviDCR92:
Based on suggestions from the Dino XMPP channel, https://github.com/dino/dino/pull/1111/commits/56916dcb8e18b2951aa8a338495aa1f86c501b82 makes various visual enhancements to quotes.

This is the newly suggested presentation:
534882de_Screenshot_20210923_233859

fiaxh:
Thanks for your PR.

- Please move your code into the `parse_add_markup` function in `helper.vala`
- In my tests, I found the quote text rather hard to read with an opacity of 60%, so I'd like to go for 70% or 75% instead.
- It would be nice if you could get rid of the token.replace("\n", ""), or at least add a comment why it's ok to do and necessary.

XaviDCR92:
@fiaxh I have applied your suggestions and rebased against latest `master`.

PR #1121, opened 2021-10-17
Update im.dino.Dino.appdata.xml.in

carlosgonz0:


sonnyp:
Might be worth doing https://blogs.gnome.org/tbernard/2021/09/07/ready-for-software-41/ as well

PR #1137, opened 2021-11-18
Fix auto download if public MUC JID is in roster

mesonium:
Dino has downloaded files (< 5 MB) automatically from occupants in public MUCs if the public MUC JID was in the user's roster. This patch fixes it.


fiaxh:
Thanks for fixing this! :slightly_smiling_face:

PR #1140, opened 2021-11-22
Added mulitple file picker

### LINKED ISSUE #1132: FR: Allow to Send multiple files at once
It would be great, to be able to select, multiple files/images at once when sharing with a Person/MUC

sojuz151:
Should fix #1132
Feedback would be appreciated

### DIFF
--- a/main/src/ui/file_send_overlay.vala
+++ b/main/src/ui/file_send_overlay.vala
@@ -25,6 +25,22 @@ public class FileSendOverlay : Gtk.EventBox {
             this.destroy();
         });
         send_button.clicked.connect(() => {
+            var file_count = get_files_count();
+            if(file_count > 10){
+                Gtk.MessageDialog msg = new Gtk.MessageDialog (
+                    new Gtk.Window(),  Gtk.DialogFlags.DESTROY_WITH_PARENT | Gtk.DialogFlags.MODAL,
+                    Gtk.MessageType.WARNING, Gtk.ButtonsType.OK_CANCEL,
+                    "Warning");
+                msg.secondary_text = _("You are about to send %i files.").printf(file_count); 
+                Button ok_button = msg.get_widget_for_response(ResponseType.OK) as Button;
+                ok_button.label = _("Send anyway");
+                if (msg.run() != Gtk.ResponseType.OK) {
+                    msg.close();
+                    return ;
+                }                                  
+                msg.close();
+            }
+
             foreach( var child in files_widget.get_children()){
                 var files_widget = (FileWidget) child;
                 if(files_widget.is_ok)
@@ -48,33 +64,33 @@ public class FileSendOverlay : Gtk.EventBox {
             }
             return false;
         });
-        update_send_label();
         add_more.clicked.connect(()=>{
             PreviewFileChooserNative chooser = new PreviewFileChooserNative(_("Select file"), get_toplevel() as Gtk.Window, FileChooserAction.OPEN, _("Select"), _("Cancel"),true);
             if (chooser.run() == Gtk.ResponseType.ACCEPT) {
                 add_files(chooser.get_files());
             }
         });
+        update_send_label();
     }
     public void add_files(SList<File> files) {
         foreach( var file in files){
             //TODO filter exisitng files or sort
             load_file_widget.begin(file);       
         }        
-        update_send_label();
     }
-
-    
-    private void update_send_label() {
-        var remaining_size = 0;
+    private int get_files_count(){
+        var ammount = 0;
         foreach( var child in files_widget.get_children()){
             var files_widget = (FileWidget) child;
             if(files_widget.is_ok)
-                remaining_size++;
+            ammount++;
         }
-        
-        send_button.label = n("Send %i file","Send %i files",remaining_size).printf(remaining_size);
-        
+        return ammount;
+    }
+    
+    private void update_send_label() {
+        var remaining_size = get_files_count();        
+        send_button.label = n("Send %i file","Send %i files",remaining_size).printf(remaining_size);        
         can_send = remaining_size != 0;
         send_button.sensitive = can_send;
         
@@ -186,6 +202,7 @@ public class FileSendOverlay : Gtk.EventBox {
         var box = new FileWidget(this,file,file_info);
         yield box.load_file();
         files_widget.add(box);
+        update_send_label();
     }
 }
 }
--- a/main/src/ui/file_send_overlay.vala
+++ b/main/src/ui/file_send_overlay.vala
@@ -9,9 +9,8 @@ namespace Dino.Ui {
 [GtkTemplate (ui = "/im/dino/Dino/file_send_overlay.ui")]
 public class FileSendOverlay : Gtk.EventBox {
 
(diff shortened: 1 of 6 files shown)


mesonium:
Thank you! It works fine, just a few comments or ideas from my side. :-)

1) As you can see, if there are too many files, pictures become stacked either at the top or at the end, until the view becomes refreshed:

Screenshot_20211126_230525_Dino

2) Aesthetically I'd find consistent button sizes better, but that's just my view :)

Screenshot_20211126_230018_Dino

3) In case it were possible to ensure that the sent files are transferred in a specific order (top-down) it would be handy to be able to rearrange the order the files. (drag/drop, ^/v buttons)

4) Additionally to the "delete button(s)" an "add more files" button could be handy in case you want to add another to the list.

5) Since the files only show the name and no preview (except for pictures), they are difficult to identify after the selection.
Therefore it would be convenient if some additional information (eg path, file size) could be displayed somewhere.

sojuz151:
In the last commit I fixed couple of issues
>As you can see, if there are too many files, pictures become stacked either at the top or at the end, until the view becomes refreshed:

This is something with FileImageWidget and I have no idea how to fix that. 

>If you hold the cursor over the entries it turns into a pointing hand, but clicking with the left mouse buttons on them does nothing.

Now you should be able to open files by clicking

> If you change the chat, the multiple-file-upload window closes and the set of selected files to send is gone. Maybe it could be saved in case you turn back to the chat again, where you wanted to send the files?

Maybe later?

>Aesthetically I'd find consistent button sizes better, but that's just my view :)

Yes, but that makes removing multiple consecutive files slower, you have to move your mouse.

>In case it were possible to ensure that the sent files are transferred in a specific order (top-down) it would be handy to be able to rearrange the order the files. (drag/drop, ^/v buttons)

Adding this seems hard.  It would be nice to write something general that could be also be applied to roaster, account list and maybe something more. 

>Since the files only show the name and no preview (except for pictures), they are difficult to identify after the selection.
Therefore it would be convenient if some additional information (eg path, file size) could be displayed somewhere.

Maybe clicking-to-open will be enough?  

>Additionally to the "delete button(s)" an "add more files" button could be handy in case you want to add another to the list. It would be awesome if you even could drag/drop files into the multi-select-files window.

Added both. 

>Does the warning belong to upper or the lower file? And what is the allowed quota and the file size? 

Added a frame to show which file this belongs to and maximum size is displayed. 

>There are a few white spaces (e.g. before braces) in the code missing or redundant.

This should be fixed.

>Is this error also shown if jingle file transfer is possible?

I don't know, right now it works just like in a single file version.

mesonium:
Thanks, I have tried the updated version it already looks better, however 

> Adding this seems hard. It would be nice to write something general that could be also be applied to roaster, account list and maybe something more.

1. Since this is not possible at all right now, do you think it would be better to sort the picked files by a) their name and b) if they exceed the file size or not, so you quickly see what you are going to upload by HTTP and what you send by jingle (probably)?

2. I picked a file and canceled the dialog and tried to open the dialog again but
```
(dino:15475): Gtk-CRITICAL **: 12:11:25.307: gtk_container_add: assertion 'GTK_IS_CONTAINER (container)' failed
(dino:15475): Gtk-CRITICAL **: 12:11:25.307: gtk_button_set_label: assertion 'GTK_IS_BUTTON (button)' failed
(dino:15475): Gtk-CRITICAL **: 12:11:25.307: gtk_widget_set_sensitive: assertion 'GTK_IS_WIDGET (widget)' failed
```
it won't open.

main/src/ui/file_send_overlay.vala
```diff
@@ -15,14 +15,14 @@ public class FileSendOverlay : Gtk.EventBox {
 
     [GtkChild] public unowned Button close_button;
     [GtkChild] public unowned Button send_button;
+    [GtkChild] public unowned Button add_more;    
     [GtkChild] public unowned Box files_widget;
-    [GtkChild] public unowned ScrolledWindow files_widget_scroll;
 
     private bool can_send = true;
 
     public GLib.List<File> remaing_files;
```
mesonium, line 23: Typo

sojuz151:
>I picked a file and canceled the dialog and tried to open the dialog again but ...

Fixed

>Via the file picker selected files which exceeds the quota, show an error. However, once the selection window is opened, additionally added files (via file picker or drag/drop) don't show an error on quota exceeding.

Fixed

>Selected files now open on a left and right click. Is this on purpose? You can click anywhere now on the image to open it, however the cursor changes only when hovered over the overlay.

Fixed

>As you can see, if there are too many files, pictures become stacked either at the top or at the end, until the view becomes refreshed:

Should be fixed now.

mesonium:
Can confirm the fixes, however:

1. the file to upload counter seem to be broken now: It counts too few until I add another file via the `add file` button. 
2. Now you can in theory send as many files as you want at once, maybe this should be limited to some sane number?
WA let's you eg only send 30 files in a batch.
3. I'm not sure if you could fix the issue with very small images or those with weird aspect ratios:
Screenshot_Dino_Aspect_Ratio

Except for that and that the selection is still lost when chat is changed while the window is open, it works fine for me. Thanks! :)

sojuz151:
> the file to upload counter seem to be broken now: It sometimes doesn't count the added files.

should be fixed now.

> Now you can in theory send as many files as you want at once, maybe this should be limited to some sane number?
    WA let's you eg only send 30 files in a batch.

Now you should get an warring box if you want to send more  than 10 files.

mesonium:
1. Could you make the list scroll to the bottom when you attach new files? Right now, if the list fills the space, adding another item will place it out-of-view, which may be confusing.
2. Could you build in a check to avoid adding a same file twice? I guess from a UX perspective it's fine to silently ignore the duplicate.

sojuz151:
Some fixes:
>Could you make the list scroll to the bottom when you attach new files? Right now, if the list fills the space, adding another item will place it out-of-view, which may be confusing.

Now files are sorted 

>Could you build in a check to avoid adding a same file twice? I guess from a UX perspective it's fine to silently ignore the duplicate.

Done

>I tried to upload a too large file to some public MUC (eg Dino's MUC) and it says 18.4 EB (exabyte!) :P

Now -1 size is reported as cant send.

mesonium:
* > Now -1 size is reported as cant send.

   I'm not sure if this `-1` is correct at all but at least on my laptop I could not reproduce the issue again in the moment. :)

* On the laptop I use a 14" screen on 2560x1440 with 200% scaling and apparently the file picker window is too large:
Screenshot_20211222_132728_Dino
and the "Add more files" and "upload" buttons are missing. Could you make the window please fit?

* The `ESC` key to close the window does sometimes not work even if there are files listed in the picker.

mesonium:
> The ESC key to close the window does sometimes not work even if there are files listed in the picker.
This happens because when items are deleted, an element of main window gets the focus and thus hitting `Esc` no longer is communicated to the modal window.
You can also focus elements of the main window but simply hitting `tab` sufficiently often.

fiaxh:
Selecting multiple files has been implemented in e76a727fdc4a28ec3b6a1a7aed57f6a283bf2eda.

PR #1160, opened 2021-12-11
fix for "retracted" call from remote side.

wkg:
if remote side cancel the call - call still Ringing.... i think this fixed it.


fiaxh:
Thanks!

PR #1166, opened 2021-12-22
Fix XEP-0198 issues

Zash:
See commit messages.

Briefly tested and appeared to work.


fiaxh:
Thanks for looking into this!

PR #1180, opened 2022-01-17
Implement support for Software Version (XEP-0092)

linkmauve:
This allows other clients to know which version of Dino is running, which helps with remote support.


linkmauve:
Hi, I rebased on top of latest master.  Please tell me if you don’t want to merge this.

PR #1186, opened 2022-01-26
Update license text

Yuubi-san:
"Pulled" from <https://www.gnu.org/licenses/gpl-3.0.txt>, which only touches URLs.

Oh, and, most importantly, a newline added at the end!


fiaxh:
Thanks!

PR #1194, opened 2022-02-10
Remove border around image file prevews

Yuubi-san:
Images with transparent background look much, much prettier when borderless (especially when emulating stickers), and opaque images made the border not visible anyway, so no change there.


dvn0:
LGTM but may help to get merged if you can you provide a screenshot

PR #1203, opened 2022-02-22
 Basic phone ringing

emildekeyser:
This works but there is no dialing sound yet and there is a bug that I was unable to fix. When closing Dino with the phone ringing (meaning you did not click 'reject') the Dino process keeps lingering, presumably because the playbin gets in some deadlock state. I hope that anyone looking at this can maybe help out.

Also a note: I also tried to do this with licanberra and with GstPlayer but I found those to be even more a PIA.

I will start working on dialing sound and sound confiuration toggles now.


mar-v-in:
1. I don't think we should ship sound files ourselves. There is a freedesktop standard for sound files and we should use it and thus use the sounds files proposed by the current sound theme instead of our own.
2. Surely we could implement the freedesktop standard without using libcanberra. As libcanberra uses gstreamer in the background, I doubt there is any OS supported by directly using gstreamer that isn't supported by libcanberra. I also don't understand what you mean with API being incomplete: [Canberra.Context.play_full()](https://valadoc.org/libcanberra/Canberra.Context.play_full.html) can be invoked with a callback which can be used to replay the file in a loop - same as your current code does with GStreamer.
3. I don't think we'd need any configuration for incoming/outgoing call sounds. They are always a sane option. If you want your device to be silent, just mute it.
4. We're currently playing sounds via the desktop's notification system. This doesn't work perfectly (as some desktops don't support it or delay the playing of the sound to the point when the notification is shown). As you do a proper sound playing, it would probably a good idea to at least remove the sound from the notification (as otherwise some users will have the sound played twice).

emildekeyser:
1. alright so i guess hardcoding paths and looking at env variables ?
3. alright il leave that be
4. i agree, im in the proces of changing that

emildekeyser:
About 2.: Because I'm new to Vala I might be missing something but if
you compare https://valadoc.org/libcanberra/Canberra.FinishCallback.html
with the actual C api you can see it is missing the _userdata_
parameter. This together with the fact that seemingly this callback runs in a
special context where it has no access to anything makes it that it is
impossible to make the loop stop. The canberra docs also say something
about the context being unpredictable:
http://0pointer.de/lennart/projects/libcanberra/gtkdoc/libcanberra-canberra.html#ca-finish-callback-t
Pottering also says that you cant call Canberra functions from that
callback which makes it even more difficult.

So I guess doing C<=>Vala interop might be a solution but like I said Im
a nooby so I did not consider that an option.

emildekeyser:
Another small note about point 1.: This means also that its important to add freedesktop sound as a (optional) dependancy for packages otherwise you can get the fantastic user experience where it seems there is no sound but its just that you dont have a sound theme. But this is more of a distro maintainer's responsability I guess.

mar-v-in:
About 1: You'd also need to use the active theme (if any) from [Gtk.Settings.gtk_sound_theme_name](https://valadoc.org/gtk+-3.0/Gtk.Settings.gtk_sound_theme_name.html). Check [the specification](https://specifications.freedesktop.org/sound-theme-spec/sound-theme-spec-latest.html#sound_lookup) for a detailed pseudocode on the sound lookup routine.


About 2: The `_userdata_` is implicitly used in Vala to allow you to either create [closures](https://en.wikipedia.org/wiki/Closure_(computer_programming)) or put a method pointer (which, in contrast to a function pointer, requires the reference to the containing object). I think the not about the context is that the callback may be called from another thread. In this case, you'd need to attach a signal source to the maincontext to ensure you are running on the main thread. The easiest way to do so is by using [Idle.add()](https://valadoc.org/glib-2.0/GLib.Idle.add.html) (which you give a reference to a method or a closure to be invoked on the main thread). The following should give you an idea of how this should probably work (untested):
```vala
private Canberra.Context ctx;
private Canberra.Proplist props;
private uint32 id;
private boolean playing;

public void play_again() {
  if (!playing) return;
  ctx.play(id, props, () => {
    Idle.add(() => {
      play_again();
      return Source.REMOVE;
    });
  });
}

public void play_loop() {
  playing = true;
  play_again();
}

public void cancel_loop() {
  playing = false;
  ctx.cancel(id);
}
```
Bonus: You can easily add a few ms of silence between the sound plays by using [Timeout.add()](https://valadoc.org/glib-2.0/GLib.Timeout.add.html) instead of Idle.add() - in case that makes it sound better.

emildekeyser:
Nice! Thanks for the info. Il look into it.

mar-v-in:
- `main` should not have a dependency on GStreamer.
- I guess you misunderstood my comment about the notification thing. There is absolutely nothing wrong with listening on call specific events in the NotificationEvents module. What I meant was, that we already do play a sound on incoming calls (but only once and it depends on the notification system supporting it) via https://github.com/dino/dino/blob/408406e652bf7450d91661dcdd38be767c7faacd/main/src/ui/notifier_freedesktop.vala#L120 We don't want to play the sound twice at the same time (as it would sound ugly), so you'd need to make sure that won't happen.

emildekeyser:
Okay, I will remove the singular sound also. I don't understand what you mean by 'main should not have a dependency on GStreamer', I did remove my dependancy on gst back to libcanberra already, did I forget a refrence somewhere ?

main/CMakeLists.txt
```diff
@@ -4,6 +4,7 @@ include(${GETTEXT_USE_FILE})
 gettext_compile(${GETTEXT_PACKAGE} SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR}/po TARGET_NAME ${GETTEXT_PACKAGE}-translations)
 
 find_packages(MAIN_PACKAGES REQUIRED
+    Gst
```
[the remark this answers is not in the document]
emildekeyser: Oh offcourse that stuff. Sorry I did not look at that commit when asking about the gst dep above and I forgot about it. Yeah I will redo that. I also realize that I should have explained in the commit msg what I was doing there. There was a bug because Gst.init wasnt called yet or something. It gets called from the RTP plugin.

emildekeyser:
BTW, I am meaning to redo the commits once you guys are happy with the code because I find them very ugly (no/bad messages, redundancy) if that is okay with you.

emildekeyser:
And if you like when we are done here I can put libcanberra on the deps wiki page.

plugins/phone-ringer/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
...
+)
+
+vala_precompile(PHONE_RINGER_VALA_C
+SOURCES
+    src/plugin.vala
+    src/register_plugin.vala
+CUSTOM_VAPIS
+    ${CMAKE_BINARY_DIR}/exports/xmpp-vala.vapi
+    ${CMAKE_BINARY_DIR}/exports/dino.vapi
+    ${CMAKE_BINARY_DIR}/exports/qlite.vapi
```
fiaxh, line 17: You are not using xmpp-vala or qlite in this plugin.

plugins/phone-ringer/src/plugin.vala
```diff
@@ -0,0 +1,71 @@
...
+
+            loop_ringer();
+
+            call.notify["state"].connect(() => {
+                if (call.state != Dino.Entities.Call.State.RINGING) {
+                    sound_context.cancel(ringer_id);
+                }
+            });
+
+        });
```
fiaxh, line 56: As @mar-v-in already mentioned, it would be a good idea to implement a `NotificationProvider` to get the call-specific notification events. The ringing sound can be started+stopped that way. You already did something in that direction in a previous version of this PR but unfortunately removed it again.

fiaxh (changes requested):

emildekeyser:
Ah, very nice that you took another look at my PR. I was starting to
worry that you maybe forgot about me. I will take a look at this stuff
asap (this week).

The actual ringing code also still realy pains me. Now that I gathered a
little bit more vala experience i'l give another shot at making it more
flexible. I hope I can put a little delay between the noises then.

plugins/rtp/src/plugin.vala
```diff
@@ -276,7 +276,6 @@ public class Dino.Plugins.Rtp.Plugin : RootInterface, VideoCallPlugin, Object {
             device_monitor.stop();
         }
         destroy_call_pipe();
-        Gst.deinit();
```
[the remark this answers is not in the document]
emildekeyser: I should have added more info about this in the commit message. The reason I removed it was because it caused the Dino proces to linger in the background when (indeed like you say) I was using Gst in the ringer plugin. BUT my reasoning for leaving it out is that this is a nasty bug that might crop up again in the future when someone tries to do anything else at all with Gst. 

And so I did the following:
+ checked this https://gstreamer.freedesktop.org/documentation/gstreamer/gst.html#gst_deinit
+ tested that Dino works without the Gst.deinit
+ tried looking for a specific reason for the deinit with ```git log -L 276,+10 -- plugins/rtp/src/plugin.vala``` and didn't find anything

All this made me conclude that it is better to leave it out but if I might be missing something or if you realy want to keep it "just in case", ofcourse let me know and I will add it back. (and maybe add a comment+commit msg warning with the stuff I am talking about here)

plugins/phone-ringer/CMakeLists.txt
emildekeyser: When I remove:
```
${CMAKE_BINARY_DIR}/exports/xmpp-vala.vapi
${CMAKE_BINARY_DIR}/exports/qlite.vapi
```
I get a whole bunch of this when running make:
```
...build/exports/dino.vapi:1289.80-1289.83: error: The symbol `Xmpp' could not be found
...build/exports/dino.vapi:1273.60-1273.67: error: The symbol `Xep' could not be found
...build/exports/dino.vapi:774.11-774.15: error: The symbol `Qlite' could not be found
```

I looked around a bit in other CMakeList.txt files. It seems it is needed to have those three together? I am not a CMake wizard, do I have to do more things then just remove the lines to get it to compile ?

plugins/phone-ringer/src/plugin.vala
```diff
@@ -0,0 +1,83 @@
+using Dino.Entities;
+using Xmpp;
+
+namespace Dino.Plugins.PhoneRinger {
+
+public class Plugin : RootInterface, NotificationProvider, Object {
+
+    private Canberra.Context sound_context;
+    private const int ringer_id = 0;
+    private const int dialer_id = 1;
```
fiaxh, line 10: We typically do constant variable names in all-uppercase.

plugins/phone-ringer/src/plugin.vala
```diff
@@ -0,0 +1,83 @@
...
+    private Canberra.Proplist dialer_props;
+
+    private void loop_ringer() {
+        sound_context.play_full(ringer_id, ringer_props, (c, id, code) => {
+            if (code != Canberra.Error.CANCELED) {
+                Idle.add(() => {
+                    loop_ringer();
+                    return Source.REMOVE;
+                });
+            }
```
fiaxh, line 21: You can introduce a timeout like this:

```suggestion
        sound_context.play_full(ringer_id, ringer_props, (c, id, code) => {
            if (code == Canberra.Error.CANCELED) return;
            Timeout.add_seconds(1, () => {
                if (!ringing) return Source.REMOVE;
                loop_ringer();
                return Source.REMOVE;
            });
```

fiaxh (commented):
This is looks quite good by now. But after trying it out, I think this really needs breaks between the sounds :wink:  To put a delay between the sounds, you can replace `Idle.add` with `Timeout.add_seconds`, as mar-v-in mentioned before. You then need some variable to store whether it is currently supposed to be ringing or not (if the noise is retracted while the timeout is waiting), and then check it in the Timeout.

plugins/rtp/src/plugin.vala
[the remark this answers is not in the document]
fiaxh: All right, just please put this in a separate commit and include your reasoning in the commit message.

plugins/phone-ringer/CMakeLists.txt
fiaxh: You're right, those need to stay, my bad.

earboxer:
@emildekeyser, I tried this out, and I like it... *except* the call is ringing tone is made even if I'm making the call using another client.

What's worse is that that sound continues even after the call is connected (on the other device) and disconnected.

So big 😢 and 👎 until that's fixed. (it should only make that noise if you're placing the call *from* this device.

emildekeyser:
Yeah I am aware of this issue, this is on my backlog to fix, sorry I
should have added this to the pr description or something.

Not sure when I will have time for this maybe this weekend, maybe in a
couple weeks.

To be honest, I am really NOT enjoying working with this
DBus/GTK/GOBject ecosystem, you cannot believe how obscure everyting is.
So that is a factor which is also slowing me down :( but I find the idea
of XMPP calling very important so il try my best.

PR #1207, opened 2022-03-04
Add smiley translations for :')

jubalh:


fiaxh:
I try to add only the most widely used and obvious smiley translations to this list. Because on one hand conversions are handy, but on the other hand every translation introduces the risk of unwanted mappings. Also, some smileys are perceived different by different groups of people, for example some people understand `:x` as sealed lips, some as kiss-face. Thus, I wouldn't want to extend this list even further, especially since there is an emoji button, and a shortcut to quickly access the emoji picker.

PR #1209, opened 2022-03-09
'Minimize to tray' functionality using SNI (StatusNotifierItem)

emildekeyser:
As promised here is a first rough draft for minize to systray. An overview of things that still need to be done:

Mandatory TODO:
- Solve the dependancy problem <<-- need feedback on this one!
- Make this feature toggleable

Optional TODO: (I would very much prefer not to do these as these will give me more work without any benefit and they are optional imo.)
- Use a "dbus menu" instead/in addition of the simple Gtk.Menu.popup() used now
- Tweak the SNI signals (dbus methods) to be even more user firendly

About the "dependancy problem": If you look at my code right now you will see that I require this https://github.com/SkyMaverick/statusnotifier external library. I am asuming it is not at all widely available in repo's (got it from the aur). So I think that leaves me with two options for a mandatory refactor:
1. Just copy SkyMaverick/statusnotifier fully into the Dino tree, it is not very big so I don't really see a problem with this but I can imagine people not liking it.
2. Implementing SNI natively in vala in the Dino tree. So basically translating all needed parts from SkyMaverick/statusnotifier. This one would be a more work for maybe only marginal gain ?

So what are your thougts ?


emildekeyser:
Awsome I will use the code you linked, where did you get that from ? Also I will fix the issues you addressed.

emildekeyser:
BTW, I am meaning to redo the commits once you guys are happy with the code because I find them very ugly (no/bad messages, redundancy) if you would like that ofcourse.

libdino/src/dbus/status_notifier_item.vala
```diff
@@ -0,0 +1,333 @@
...
+    public signal void context_menu(int x, int y);
+    public signal void activate(int x, int y);
+    public signal void secondary_activate(int x, int y);
+    public signal void scroll(int delta, string orientation);
+
+    private DBusStatusNotifierItem dbus_item;
+    private DBusMenu dbus_menu;
+    private static int last_registration_id = 0;
+    private int registration_id = last_registration_id++;
+    private string name { owned get { return @"$id-$(GLib.Uuid.string_random())-$registration_id"; } }
```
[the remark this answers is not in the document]
mar-v-in: Yes, you'd need to add `posix` in the `PACKAGES` section of the `vala_precompile` command, as it's not a real library.

libdino/src/dbus/status_notifier_item.vala
```diff
@@ -0,0 +1,369 @@
...
+    private StatusNotifierItem item;
+
+    public DBusMenu(StatusNotifierItem item) {
+        this.item = item;
+    }
+
+    [DBus (name = "Version")]
+    public uint32 version { get { return 3; } }
+    [DBus (name = "TextDirection")]
+    public string text_direction { get { return "ltr"; } }
```
mar-v-in, line 90: I think we should somehow pass the actual text direction (`rtl` for rtl languages) in here, otherwise some desktops might render rtl langauges incorrectly.

main/data/menu_app.ui
```diff
@@ -20,6 +20,10 @@
                 <attribute name="action">app.about</attribute>
                 <attribute name="label" translatable="yes">About Dino</attribute>
             </item>
+            <item>
```
mar-v-in, line 23: Please put the Quit action in a new section

libdino/src/dbus/status_notifier_item.vala
```diff
@@ -0,0 +1,369 @@
...
+        return true;
+    }
+
+    [DBus (name = "Event")]
+    public void event(int32 id, string event_id, Variant data, uint32 timestamp) {
+        resolve_event(id, event_id, data, timestamp);
+    }
+
+    [DBus (name = "EventGroup")]
+    public int[] event_group(DBusMenuEventStruct[] events) throws DBusError
```
mar-v-in, line 246: Please stick with the code style of having opening brace on the same line.

main/src/ui/application.vala
```diff
@@ -116,6 +119,54 @@ public class Dino.Ui.Application : Gtk.Application, Dino.Application {
         }
     }
 
+    private void setup_systray()
```
mar-v-in, line 122: Again opening brace on same line.

mar-v-in (commented):
Seems good beside a few minors. I'll do a bit of testing this weekend and merge if everything works as expected.

emildekeyser:
woopsie ignore the poundsign 2 that last commit, I didnt know that it has special meaning on gh

libdino/src/dbus/status_notifier_item.vala
```diff
@@ -0,0 +1,388 @@
...
+[DBus (name = "com.canonical.dbusmenu")]
+private class DBusMenu : Object {
+    private StatusNotifierItem item;
+
+    public DBusMenu(StatusNotifierItem item) {
+        this.item = item;
+    }
+
+    [DBus (name = "TextDirection")]
+    public string text_direction { get {
```
mar-v-in, line 88: I think it would probably be better to have `text_direction` exposed via the `StatusNotifierItem` class (like most properties in `DBusStatusNotifierItem`) and then also set it in `setup_systray()`. In the `setup_systray()` function you have access to the [`Gtk.get_locale_direction()`](https://valadoc.org/gtk+-3.0/Gtk.get_locale_direction.html).

Beside that, [`Pango.find_base_dir()`](https://valadoc.org/pango/Pango.find_base_dir.html) probably does a better job in determining text direction from a string :wink:

emildekeyser:
Strange, I dont get that int to str error that makes the build fail on my machines. But when I was talking with the guy from xapps he also had the same issue. Do you guys know what is causing that ?

libdino/src/dbus/status_notifier_item.vala
emildekeyser: k, il get to it

wkg:
Left click to systray works same a right - showed menu too, its problem only at me? xfce 4.16.0

wkg:
@emildekeyser please "Fetch upstream" on you fork...

emildekeyser:
It is (most likely) an XFCE problem. More specifically xfce4-panel is
probably not calling the 'Activate' dbus method on left click. I did
some looking around and strangly enough I did find this:
https://github.com/xfce-mirror/xfce4-panel/blob/master/plugins/systray/sn-item.c#L1082.
Maybe search the issues/post a bug at the official xfce4-panel repo about
this ?

emildekeyser:
on it!

wkg:
> It is (most likely) an XFCE problem. More specifically xfce4-panel is probably not calling the 'Activate' dbus method on left click. I did some looking around and strangly enough I did find this: https://github.com/xfce-mirror/xfce4-panel/blob/master/plugins/systray/sn-item.c#L1082. Maybe search the issues/post a bug at the official xfce4-panel repo about this ?
> […](#)
> On 22/04/03 10:51, wkg wrote: Left click to systray works same a right - showed menu too, its problem only at me? xfce 4.16.0 -- Reply to this email directly or view it on GitHub: [#1209 (comment)](https://github.com/dino/dino/pull/1209#issuecomment-1086917028) You are receiving this because you authored the thread. Message ID: ***@***.***>

I looked in the xfce-panel sources, they also noticed the problem: 
https://github.com/xfce-mirror/xfce4-panel/blob/master/plugins/systray/sn-button.c#L299

according to their code, this happens if "is_menu = true":
https://github.com/xfce-mirror/xfce4-panel/blob/master/plugins/systray/sn-item.c#L798-L803

i tried changed it in you fork to "false" and this is fixed it... 
do we really need "true" for this? or for others(kde, gnome) need "true"?

emildekeyser:
i looked at the spec, actually setting is_menu to true is an error on my part, changing and testing it now

emildekeyser:
I just discovered something horrible. I am testing on an old laptop with Cinnamon and without a login manager (just .xinitrc immediate login). When autostarting Dino the SNI icon doesnt show up. I would guess there is a race condition between StatusNotiferItem and StatusNotfierWatcher. Maybe it is a freak occurance caused by my janky setup but it is a very bad user experience so I should investigate. Adding a 3 second delay in the "Startup Apllications" menu fixed it but this is not a solution. I am thinking a clean way to fix this is to have the systray registration be in a loop retrying to register when it fails for 10 seconds. Or even more clean might be to monitor dbus and wait (with sni.register, not the whole program) till the Watcher comes online. But that might be a bit complicated.

emildekeyser:
It seems that on my old laptop running cinnamon there is a race condition which causes the SNI icon to not show up on autostart. It can be fixed by adding some seconds delay. This issue might be very fringe so I am inclined to leave it like that but if you guys want I could make the systray code concurently retry on failure or monitor dbus till it sees the watcher for X amount of seconds.

143981:
@emildekeyser please "Fetch upstream" again for gtk4

emildekeyser:
> @emildekeyser please "Fetch upstream" again for gtk4

:white_check_mark:

143981:
ERROR:promise.c:667:gee_promise_future_set_value: assertion failed: (_state == State.INIT)
Bail out! ERROR:promise.c:667:gee_promise_future_set_value: assertion failed: (_state == State.INIT)

emildekeyser:
> > ERROR:promise.c:667:gee_promise_future_set_value: assertion failed: (_state == State.INIT)
> > Bail out! ERROR:promise.c:667:gee_promise_future_set_value: assertion failed: (_state == State.INIT)
> 
> this only on my system?

 well I dont get that on my system

143981:
rm -r build and rebuild fix my problem....
Test it on xfce4, good work.
@emildekeyser Thank you!

The only thing missing is changing the icon when a new message arrives...

emildekeyser:
> rm -r build and rebuild fix my problem.... Test it on xfce4, good work. @emildekeyser Thank you!
> 
> The only thing missing is changing the icon when a new message arrives...

alright il look into adding that, you mean whenever there are any unread msgs there should be a little envelope or something in the corner of the dino icon ?

143981:
> > rm -r build and rebuild fix my problem.... Test it on xfce4, good work. @emildekeyser Thank you!
> > The only thing missing is changing the icon when a new message arrives...
> 
> alright il look into adding that, you mean whenever there are any unread msgs there should be a little envelope or something in the corner of the dino icon ?

yes, or just change icon if any new message and change back after focus window

143981:
[icon.patch.txt](https://github.com/dino/dino/files/9946532/icon.patch.txt)
I would like this, but this is a bad decision, of course

143981:
i continue testing, and another wish: keep the position of the window when it is hidden/showed if it is possible

fgnix:
I was testing your pull request and it is working fine I only found a small issue.
If i close the app (leaving it open in background) I no longer receive notification for new messages.
Also the messages on the chat that was open are marked as read even if the app is in tray.
It seems to me that dino thinks to be open even while in tray.

ztefn:
It's unfortunate to see this feature is not merged in Dino. However, I would like to inform anyone it may concern that this code has been really helpful to me in order to continue indicator support in [Haguichi](https://github.com/ztefn/haguichi), which I very recently ported to GTK4 and libadwaita. Without the existence of this example I would likely have dropped indicator support. But now the indicator is even working on more distros than before, because this implementation doesn't depend on libappindicator!

I made few minor modifications to suit specific needs for Haguichi. But I also made some more generic tweaks to update the menu item properties and to set the `host_registered` state. Also silenced all Vala compiler warnings by adding minimal error handling. The file is located at [src/status-notifier-item.vala](https://github.com/ztefn/haguichi/blob/master/src/status-notifier-item.vala). I didn't change any original code formatting so the file is easily diffable with the one from this pull request or the gist from @mar-v-in.

The modified file contains attribution to the "Dino contributors". Let me know if you are an author and would like to be personally attributed using your real name.

Thanks!

Issue #1125, opened 2021-10-21
If file transfer failed, the possibility to retry


pizdjuk:
When file transfer fails (on receive, or send) give the possibility to retry the transfering and/or show the full link to the file.

Dino: 0.2.0~git127.20210924.ce227f71-1

PR #1210, opened 2022-03-09
Allow to retry failed file receipts

Yuubi-san:
I only tested on a failed http download, no jingle transfers.

Partially addresses #1125.


aaruni96:
Would be great to have this merged

PR #1211, opened 2022-03-10
conversation_view.vala: update_highlight on touch event

XaviDCR92:
Mobile devices such as the PinePhone do not receive leave/enter events for some reason, so actions such as editing a previous message are not possible unless a keyboard shortcut is used, which not be as convenient or even possible under some DEs. As a workaround, `TOUCH_MASK` can be used for the same purpose instead.

This commit should not affect desktop platforms with mouse or touchpad.


fiaxh:
This works since e35df88d4a00c3a34f2b4d9fb7f10bb5d877bd29 using a `GestureClick` controller.

PR #1212, opened 2022-03-12
Display friendlier xmpp: links when a jid is in the roster

dilinger:
In jabber chats, there may be cases where an xmpp link is a valid
    JID that's in an account's roster. For example, Cheogram's jmp.chat
    service uses xmpp: links to support SMS group chat over jabber. They
    overlay a group chat over an individual chat by having the list of
    phone numbers as the chat contact JID (eg, a JID of
    "+11111111111,+13333333333@cheogram.com" for a group chat between
    someone with the phone number (111) 111-1111, another person with
    the number (333) 333-3333, and whatever the user's phone number is),
    and each message in the chat window that is received is prefixed with
    an xmpp: link that has just the single phone number of the sender (eg,
    "<xmpp:+111111111111@cheogram.com> " for (111) 111-1111). That JID
    that's in the message body is the same one that's used for 1-to-1 SMS
    chats, so it is likely to be in the user's roster as a contact.
    
    The JIDs aren't particularly legible or memorable, so we'd instead
    like to display the name that's in the roster while keeping the xmpp
    link as clickable. In order to do that, this patch modifies
    Util.parse_add_markup_theme() (since that's where the <a> html tag
    markup is created) to add a callback function that takes a uri and
    returns what the <a> tag should display. If the callback returns null,
    we just stick with the default behavior of just displaying the raw
    xmpp link. However, if the uri includes a JID that's in the user's
    roster, we return the name that's in the roster.
    Util.parse_add_markup_theme() then uses that returned name in the <a>
    tag.
    
    A callback is used in order to keep the lookup generic, for future
    uses. While the call to Util.parse_add_markup_theme() inside of
    MessageItemWidget::generate_markup_text()'s non-MUC code has a
    callback that specifically only checks for xmpp: links, there's no
    reason that it needs to be limited to that. It could handle any link;
    for example, translating an http:// link to that same link but with
    an "(insecure)" string added to it.

### DIFF
--- a/main/src/ui/conversation_content_view/message_widget.vala
+++ b/main/src/ui/conversation_content_view/message_widget.vala
@@ -94,7 +94,17 @@ public class MessageMetaItem : ContentMetaItem {
         if (conversation.type_ == Conversation.Type.GROUPCHAT) {
             markup_text = Util.parse_add_markup_theme(markup_text, conversation.nickname, true, true, true, Util.is_dark_theme(this.label), ref theme_dependent);
         } else {
-            markup_text = Util.parse_add_markup_theme(markup_text, null, true, true, true, Util.is_dark_theme(this.label), ref theme_dependent);
+            Util.LinkDisplay roster_lookup = (uri) => {
+                string? s = null;
+                if (GLib.Uri.parse_scheme(uri) == "xmpp") {
+                    try {
+                        Jid? j = new Jid(uri["xmpp:".length:uri.length]);
+                        s = Dino.get_real_display_name(stream_interactor, conversation.account, j);
+                    } catch (InvalidJidError e) { /* it's fine */ }
+                }
+                return s;
+            };
+            markup_text = Util.parse_add_markup_theme(markup_text, null, true, true, true, Util.is_dark_theme(this.label), ref theme_dependent, false, roster_lookup);
         }
 
         if (message.body.has_prefix("/me ")) {
--- a/main/src/ui/util/helper.vala
+++ b/main/src/ui/util/helper.vala
@@ -243,7 +243,9 @@ public static string parse_add_markup(string s_, string? highlight_word, bool pa
     return parse_add_markup_theme(s_, highlight_word, parse_links, parse_text_markup, parse_text_markup, false, ref ignore_out_var);
 }
 
-public static string parse_add_markup_theme(string s_, string? highlight_word, bool parse_links, bool parse_text_markup, bool parse_quotes, bool dark_theme, ref bool theme_dependent, bool already_escaped_ = false) {
+public delegate string? LinkDisplay(string uri);
+
+public static string parse_add_markup_theme(string s_, string? highlight_word, bool parse_links, bool parse_text_markup, bool parse_quotes, bool dark_theme, ref bool theme_dependent, bool already_escaped_ = false, LinkDisplay? lookup = null) {
     string s = s_;
     bool already_escaped = already_escaped_;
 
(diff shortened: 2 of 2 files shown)


dilinger:
Here's an example of what it looks like before:
before

And what it looks like afterwards:
after

dilinger:
I just realized (as I'm using it) that this doesn't modify the displayed message string in a notification popup window, which would be nice to see. I might add support for that as well, if you're amenable to merging this patch or something like it.

dilinger:
Hi, can you please let me know what you think of this idea, even if there's specific things you don't like the patch (or just haven't had time to look at the patch)?

speedygonzales9:
Hi, 

I am using Cheogram's jmp.chat service since I cannot operate a phone. I want to thank @dilinger for making this patch! This is very helpful when I am in a SMS group chat! 

Is there an update about patching/fixing push notifications so it shows you a contact name for a known contact instead of a Jabber ID? 

Thank you

sradigan:
I wanted this plus the latest off master, so I make an ArchLinux `PKGBUILD`

The `prepare` function will fail if it is not run as a clean build, so run with `makepkg -sC`.

Original PKGBUILD is from the `dino-git` package on the AUR [here](https://aur.archlinux.org/packages/dino-git)

```bash
pkgbasename=dino
pkgname=dino-resolve-xmpp-links-git
pkgver=v0.3.0.r20.g55a46022
pkgrel=1
pkgdesc="Simple and modern Jabber/XMPP client written in vala"
arch=('i686' 'x86_64' 'aarch64')
url="https://dino.im/"
license=('GPL3')
depends=('glib2>=2.38' 'glib-networking' 'gtk3>=3.22' 'gpgme'
	 'libgee>=0.10' 'libgcrypt' 'libsoup' 'sqlite' 'qrencode'
	 'libsignal-protocol-c' 'gspell' 'gst-plugins-base'
	 'gst-plugins-good' 'gst-plugin-gtk' 'webrtc-audio-processing'
	 'libnice' 'libsrtp')
makedepends=('git' 'cmake' 'vala>=0.34' 'ninja')
optdepends=('libcanberra: audio notification support')
provides=("${pkgbasename}")
conflicts=("${pkgbasename}")
source=('dino::git+https://github.com/dilinger/dino#branch=resolve-xmpp-links')
sha512sums=('SKIP')

### Any or all of the following plugin can be disabled: omemo, openpgp, http-files. Populate following string as per your needs and uncomment it.

# _DISABLE_PLUGINS="--disable-plugin='omemo;openpgp;http-files'"

### Any or all of the following plugin can be enabled: notification-sound. Populate the following string as per you needs and uncomment it.
### Install any needed optional dependency before

#_ENABLE_PLUGINS="--enable-plugin='notification-sound'"

### Since 55b148a0f78c1abef57c21b2a84d97cad32b2290 dino uses by default the shared libsignal library.
# To restore the old behaviour add '--with-libsignal-in-tree' flag to configure
prepare() {
    cd "${srcdir}/${pkgbasename}"
    git remote add -f upstream https://github.com/dino/dino.git
    git merge upstream/master -m "merge upstream"
}

pkgver() {
    cd "${srcdir}/${pkgbasename}"
    git describe --long --tags | sed 's/\([^-]*-g\)/r\1/;s/-/./g'
}

build() {
    cd "${srcdir}/${pkgbasename}"
    ./configure $_DISABLE_PLUGINS $_ENABLE_PLUGINS --with-tests --prefix="/usr" CC="$CC" CFLAGS="$CFLAGS" VALACFLAGS="$VALACFLAGS"
    make
}

package() {
    cd "${srcdir}/${pkgbasename}"
    make DESTDIR="${pkgdir}/" install
}

check() {
    cd "${srcdir}/${pkgbasename}"
    echo "Executing xmpp-vala-test:"
    build/xmpp-vala-test
}
```

dilinger:
I updated this to fix a conflict against current HEAD, and also to make the language in the commit/MR a bit clearer (hopefully).

fiaxh:
We do not think that it would be a good idea to just modify a message like this.

Someone might actually want to send `xmpp:user@example.com` as a URI. If the receiving client just replaces that text, it means that certain texts can't be sent anymore.

We generally aim to display messages as they were written by the sender, since the sender knows best what they want to say. If the receiving client just modifies the sender's text, that might break the original meaning. If the sender wants the receiving client to do something with the message, it should annotate the message stanza in a standardized way. We would consider supporting this if it was properly standardized.

dilinger:
Okay, thanks for looking at this. Can you give me a hint as to what kind of standardized annotation you're thinking of? This translation pretty much has to be done on the recipient side, as the sender has no idea of what's in the recipient's roster (and should not know).

ItsRogueRen:
> We do not think that it would be a good idea to just modify a message like this.
> 
> Someone might actually want to send `xmpp:user@example.com` as a URI. If the receiving client just replaces that text, it means that certain texts can't be sent anymore.
> 
> We generally aim to display messages as they were written by the sender, since the sender knows best what they want to say. If the receiving client just modifies the sender's text, that might break the original meaning. If the sender wants the receiving client to do something with the message, it should annotate the message stanza in a standardized way. We would consider supporting this if it was properly standardized.

Would it not be possible to make this a toggle in the settings? That way the option is available but it isn't the default behavior

dilinger:
> > We do not think that it would be a good idea to just modify a message like this.
> > Someone might actually want to send `xmpp:user@example.com` as a URI. If the receiving client just replaces that text, it means that certain texts can't be sent anymore.
> > We generally aim to display messages as they were written by the sender, since the sender knows best what they want to say. If the receiving client just modifies the sender's text, that might break the original meaning. If the sender wants the receiving client to do something with the message, it should annotate the message stanza in a standardized way. We would consider supporting this if it was properly standardized.
> 
> Would it not be possible to make this a toggle in the settings? That way the option is available but it isn't the default behavior

I can certainly understand the concern, as @fiaxh 's comment is absolutely right. A settings toggle would need to be per-chat rather than global, because you might have a cheogram SMS group text chat where you want the feature enabled, and another regular jabber chat where someone might be sending you xmpp links without wanting them translated.

I was brainstorming with someone else about using various extensions to annotate the message in such a way that tells the recipient that they may/should translate xmpp: links. However, we didn't come up with anything workable; hence my request for a hint. It sounds like the jmp.chat folks ARE willing to modify the code to support this, but we need to figure out what standardized method would work for dino (and other xmpp clients).

PR #1213, opened 2022-03-14
set a generic MIME type for encrypted file

### LINKED ISSUE #1116: "omemo" not a valid MIME type
https://github.com/dino/dino/blob/7ce49d7d68f19fb49a33e907485d83da2c9131a3/plugins/omemo/src/file_transfer/file_encryptor.vala#L41

That's not a [valid MIME type](https://datatracker.ietf.org/doc/html/rfc2045#section-5.1).

If the goal is to hide the real content type, why not set it to `application/octet-stream` or something overly generic like that?

sunny775:
This sets a valid but generic MIME type for OMEMO encrypted files. 
*application/octet-stream* though valid MIME type for the encrypted stream, doesn't tell the exact MIME type of the encrypted file thus serves the purpose.

Closes #1116


sunny775:
Hi @mar-v-in 👋
kindly review this PR

fiaxh:
Thanks!

Issue #1220, opened 2022-03-29
Feature request: XEP-0352: Client State Indication


Zash:
[XEP-0352: Client State Indication](https://xmpp.org/extensions/xep-0352.html) defines a way for the client to indicate its active/inactive state, which is commonly used by servers to enable more aggressive optimizations meant to reduce network traffic, and thus radio and battery usage on for mobile devices.

Since Dino now sees some usage on actual mobile devices (especially with the libhandy branch) and since laptops are also mobile devices, it would be good to support this.

Implementation would presumably involve watching application focus events and sending the corresponding commands per the XEP, and of course feature detection so that it is only enabled when supported by the server.

PR #1223, opened 2022-04-13
Send idle presence on screen lock event

### LINKED ISSUE #715: Feature request: Set account to away while screen locked
It would be nice to have the ability to set the presence status to away when the screen is locked to have users of traditional clients not be confused by you being constantly available.

Patiga:
Fixes #715 

Connects to the login1 session to access the LockedHint property
Tested with two dino clients on separate devices
I mostly improvised where to add a new class/member/signal, so please give feedback whether I should relocate code to somewhere else :)


dos1:
This could also set Client State Indication (XEP-0352, #1220), making it very useful on mobile devices.

selurvedu:
What if I don't want this? I prefer to now disclose my availability. I want to be either online or offline.

Issue #1195, opened 2022-02-13
Incorrect calculation of best camera framerate


foucault:
The best framerate for a v4l2 webcam is not properly calculated in `get_best_caps` (`plugins/rtp/src/device.vala`).

When the algorithm iterates over all the available v4l2 capabilities it tries to determine the best framerate for each one of YUYV (video/x-raw) modes presented by the hardware (`best_fraction`, line 357 in `device.vala`). Regardless of what's determined to be the "best" YUYV mode from within the conditional right after (line 385) the best fractional framerate will always point to the last iterated framerate, which might be an extremely low one, like 7 or 5 FPS. When the framerate is then stored in the `Gst.Structure` (line 394) the fractional framerate will always be that last value which might be different than the correct one as calculated by `best_fps` (line 386). This workaround solves this issue by only updating `best_fraction` when the conditional in line 385 is satisfied.

```diff
diff --git a/plugins/rtp/src/device.vala b/plugins/rtp/src/device.vala
index d4eca09a..1db8c996 100644
--- a/plugins/rtp/src/device.vala
+++ b/plugins/rtp/src/device.vala
@@ -354,6 +354,7 @@ public class Dino.Plugins.Rtp.Device : MediaDevice, Object {
             int best_height = 0;
             for (int i = 0; i < device.caps.get_size(); i++) {
                 unowned Gst.Structure? that = device.caps.get_structure(i);
+                Value? best_fraction_now = null;
                 if (!that.has_name("video/x-raw")) continue;
                 int num = 0, den = 0, width = 0, height = 0;
                 if (!that.has_field("framerate")) continue;
@@ -369,7 +370,7 @@ public class Dino.Plugins.Rtp.Device : MediaDevice, Object {
                         int fps = den > 0 ? (num/den) : 0;
                         int in_fps = in_den > 0 ? (in_num/in_den) : 0;
                         if (in_fps > fps) {
-                            best_fraction = fraction;
+                            best_fraction_now = fraction;
                             num = in_num;
                             den = in_den;
                         }
@@ -386,6 +387,7 @@ public class Dino.Plugins.Rtp.Device : MediaDevice, Object {
                     best_width = width;
                     best_height = height;
                     best_index = i;
+                    best_fraction = best_fraction_now;
                 }
             }
             Gst.Caps res = caps_copy_nth(device.caps, best_index);
```

fiaxh:
[...]

PR #1224, opened 2022-04-19
Support devices with multiple framerate options in get_max_fps

wentam:
Some video devices will report framerate as a single value:
image

And others will report framerate as a list of options:
image

At the time of writing, dino only works with the former. When attempting to video call using a device that has multiple framerates, get_max_fps will return zero and as a result dino won't find any video devices to use.

This PR should support both cases properly. 

Note that I also had to apply the patch in #1195 to make this work as well, as the correct framerate was not obtained for use in the video call without it.


maweil:
I can confirm that my Logitech C270 does work now after applying this PR and the mentioned patch. It was not recognized before.

fiaxh:
Thanks for investigating and fixing this issue!

PR #1233, opened 2022-05-06
Inhibit suspend during calls

XaviDCR92:
## Background
During a call, the device was still able to suspend during the call, causing inconvenience to users, specially on mobile devices, which are suspended more often to save battery.

Suspend inhibitors are therefore used to avoid an operation such as an ongoing call from being suspended. More specifically, the `Gtk.Application.inhibit/unhibit` function pair has been used.

## Known issues

This PR is working well for my PinePhone (`systemd-inhibit --list` would list the suspend inhibitor triggered by Dino), but does not seem to cause any effect on Kubuntu 20.04 LTS. I have not tested thoroughly, though.

## Additional notes

Maybe inhibiting `idle` (so the session is not locked and/or the screen turned off) could be reasonable as well. Please share your thoughts.

### DIFF
--- a/main/src/ui/call_window/call_window_controller.vala
+++ b/main/src/ui/call_window/call_window_controller.vala
@@ -22,6 +22,7 @@ public class Dino.Ui.CallWindowController : Object {
     private bool window_size_changed = false;
     private ulong[] call_window_handler_ids = new ulong[0];
     private ulong[] bottom_bar_handler_ids = new ulong[0];
+    private uint inhibit_cookie;
 
     public CallWindowController(CallWindow call_window, CallState call_state, StreamInteractor stream_interactor) {
         this.call_window = call_window;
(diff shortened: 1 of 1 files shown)


XaviDCR92:
> - It's not ideal to put inhibiting and uninhibiting into different places and to duplicate the code. You could tie the inhibiting to the call window instead of the call itself. You can inhibit on [window creation](https://github.com/dino/dino/blob/master/main/src/ui/call_window/call_window_controller.vala#L26) and uninhibit on the Window.destroy signal.

Thanks for the tip! I have rebased the branch accordingly.

> - Since you want to prevent the screen from locking, you should inhibit idle, not suspend.

It is true `idle` prevents the screen from locking, although a mobile user might want to have a call (typically, an audio-only call) with screen off so as to save battery and reduce heat (I actually do this all the time). In such case, `idle` would not prevent automatic suspend, so the latter must be explicitly inhibited as well.

Therefore, I have updated the branch so now both `idle` and `suspend` are inhibited during calls.

XaviDCR92:
Latest push was to remove unneeded changes to `main/src/ui/application.vala`.

fiaxh:
> does not seem to cause any effect on Kubuntu 20.04 LTS. I have not tested thoroughly, though.

Did you only check with `systemd-inhibit --list` or did you actually see that your screen locks during a call? As far as I understand, the Gtk.inhibit function supports a number of ways to inhibit suspend, and might have used a non-systemd one on your system. I also tested it on my system - it doesn't show up in that list but works nevertheless.

XaviDCR92:
I can confirm the inhibitor locks are enabled during the call, as shown below:
image

PR #1234, opened 2022-05-14
Show warning dialog on failed GTK video sink creation

XaviDCR92:
## Background

Dino depends on `gstreamer1.0-gtk3` (or equivalent package provided by the distribution) be available on the host system in order to allow [`video_widget.vala`](https://github.com/dino/dino/blob/99c076254abc6e2b03b784732a76a389e5a4f801/plugins/rtp/src/video_widget.vala#L32) to create GTK video sinks, which are used for video calls. Unfortunately, such requirement was not specified anywhere and, since this is a runtime (rather than build time) requirement, Dino would still build without issues.

Moreover, Dino would only return a lousy warning message over `stderr` instead of showing a warning dialog, so it would be visible to users running Dino from the command line only.

## Implementation

In order to assist users with this issue, this PR suggests to display a more informative `Gtk.MessageDialog` to those running Dino without `gstreamer1.0-gtk3`, as shown below:

image

### DIFF
--- a/plugins/rtp/src/video_widget.vala
+++ b/plugins/rtp/src/video_widget.vala
@@ -41,6 +41,14 @@ public class Dino.Plugins.Rtp.VideoWidget : Gtk.Bin, Dino.Plugins.VideoCallWidge
             add(widget);
             widget.visible = true;
         } else {
+            var dialog = new Gtk.MessageDialog(null, MODAL, WARNING, OK, null);
+
+            dialog.text = "Could not create GTK video sink";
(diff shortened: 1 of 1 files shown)


hololeap:
I package for Gentoo and it hasn't been obvious which dependencies are needed to make video/audio calls work correctly. This could be helpful for others who are building from source outside of a package manager.

mar-v-in:
We're now using GTK4 and the video widget does not require any special gstreamer sink.

PR #1235, opened 2022-05-14
Add send button on chat view

XaviDCR92:
## Background

On some specific devices e.g.: PinePhone, using the current key combination to introduce a `\n` character to a chat (i.e., `Shift+Enter`) is not possible from a virtual keyboard, such as [`squeekboard`](https://gitlab.gnome.org/World/Phosh/squeekboard). On the other hand, mobile-friendly instant messaging applications such as [Conversations](https://conversations.im) feature a send button, instead of using the `Enter` key for such purpose.

## Overview

As the title suggests, this PR adds a send button (using the `document-send` icon) to the right side of the chat view. The button is only sensitive if the chat message box is not empty:

image
image

When the send button is enabled, the `Enter` key can be used to introduce a `\n` character:

image

In order not to break existing behaviour, the button is optional and disabled by default, and it can be enabled from the settings menu, as shown below. Likewise to other available settings, send button settings are persistently stored into the database.

image

### DIFF
--- a/libdino/src/entity/settings.vala
+++ b/libdino/src/entity/settings.vala
@@ -12,6 +12,7 @@ public class Settings : Object {
         notifications_ = col_to_bool_or_default("notifications", true);
         convert_utf8_smileys_ = col_to_bool_or_default("convert_utf8_smileys", true);
         check_spelling = col_to_bool_or_default("check_spelling", true);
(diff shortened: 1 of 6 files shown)


XaviDCR92:
Superseded by https://github.com/dino/dino/pull/1418.

PR #1236, opened 2022-05-15
User-selectable default encryption settings

XaviDCR92:
## Background

Despite Dino claiming _"no unencrypted chats leaving your computer"_ on [its website](https://dino.im/), truth is this only applies to *transport*-level encryption, whereas **end-to-end encryption is disabled by default**, which is a huge privacy risk for users. So far, OMEMO or OpenPGP must be explicitly enabled by the user from the encryption widget below, for each new conversation:

image

Several attempts at solving this issue have already been proposed:
- https://github.com/dino/dino/pull/845
- https://github.com/dino/dino/pull/632
- https://github.com/dino/dino/pull/1035

Unfortunately, all of them were unsuccessful since they preferred OMEMO as the default encryption method, a decision some contributors did not particularly agree with.

## Overview

Therefore, this PR instead prompts the user for a default encryption method, and stores this choice persistently for any new conversations. The first time a user starts a chat, the following modal dialog shows up:

image

### Features

- These settings are persistently stored into the database, and can be changed anytime from the settings menu, as shown below.

image

- The modal dialog above is not shown for public chat rooms, following the same logic as the [encryption button](https://github.com/dino/dino/blob/99c076254abc6e2b03b784732a76a389e5a4f801/main/src/ui/chat_input/encryption_button.vala#L93-L102), or when a default encryption method has been already selected.
- Previous encryption settings for each conversation are respected.
     - In order not to break existing behaviour, this means unencrypted chats are left as such, and the dialog above will only appear for new conversations.

## Additional notes

- Neutrality over the decision to pick OpenPGP or OMEMO as the default encryption method has been respected as much as possible for this PR. Unfortunately, GTK automatically activates one of the `RadioButton`s that are shown on the modal dialog, so it inevitably had to be one of them (of course, unencrypted cannot be the default option).
    - As for the settings menu, "Ask" is selected by default, so the modal dialog above appears when the user starts a new conversation.
- As opposed to the settings menu, XEP numbers for OMEMO and OpenPGP have been referenced on the modal dialog so users can search them in order to get a more informed decision.
- `Encryption.parse` was not in fact required (plain `int` could have been stored into the database instead of `string`), but makes it more readable when inspecting the database.

I hope this PR finally solves this issue, so users can enjoy end-to-end encrypted chats by default!

### DIFF
--- a/libdino/src/entity/conversation.vala
+++ b/libdino/src/entity/conversation.vala
@@ -32,7 +32,7 @@ public class Conversation : Object {
             }
         }
     }
-    public Encryption encryption { get; set; default = Encryption.NONE; }
+    public Encryption encryption { get; set; default = Encryption.UNKNOWN; }
     public Message? read_up_to { get; set; }
     public int read_up_to_item { get; set; default=-1; }
 
--- a/libdino/src/entity/encryption.vala
+++ b/libdino/src/entity/encryption.vala
@@ -6,7 +6,28 @@ public enum Encryption {
         OMEMO,
         DTLS_SRTP,
         SRTP,
-        UNKNOWN,
+        UNKNOWN;
+
+        public static Encryption parse(string str) {
+                switch (str) {
+                        case "DINO_ENTITIES_ENCRYPTION_NONE":
+                                return NONE;
+                        case "DINO_ENTITIES_ENCRYPTION_PGP":
+                                return PGP;
+                        case "DINO_ENTITIES_ENCRYPTION_OMEMO":
+                                return OMEMO;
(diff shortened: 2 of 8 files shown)


dvn0:
@fiaxh @mar-v-in

Please review this well-crafted PR.

Alternatively please disable the Pull Requests feature for this project if you don't care for outside contributions.

SuperSandro2000:
Can you rebase?

XaviDCR92:
> Can you rebase?

I have rebased against latest `master` on https://github.com/dino/dino/pull/1414.

PR #1237, opened 2022-05-21
update appdata

Aman9das:


fiaxh:
Thanks for your PR.

You edited the im.dino.Dino.appdata.xml file directly, however this file is generated from the irespective .in file. Also, the content_rating was missing some entires, and the release version was incorrect.

The appdata.xml.in file is updated accordingly in e833a924b5a66048506b4e0b0885ce2e35cac6fc. The appdata.xml will be updated from it the next time it is regenerated - probably the next time we fetch translations.

Aman9das:
Thanks! 

Also now that flatpak dino has achieved feature parity, would you like to get access to the flathub repo? There is popular demand for publishing to main flathub, which I think the developers should have access to..

Issue #1227, opened 2022-04-25
Segfault when calling contact from Conversations


stevenroose:
When I'm logged into Dino and Conversations and I'm calling a contact through Conversations, my Dino crashes with this message:
```
(dino:3800000): libdino-CRITICAL **: 11:50:14.152: dino_module_manager_get_module: assertion 'identity != NULL' failed
[1]    3800000 segmentation fault (core dumped)  env G_MESSAGES_DEBUG=libdino,OMEMO ~/bin/dino
```

Running latest master as of now (baa4a6a1eb4e98558acfef6e2a24c2c13d42db5c).

This is the coredump

```
coredumpctl debug 3800000
           PID: 3800000 (dino)
           UID: 1000 (steven)
           GID: 1000 (steven)
        Signal: 11 (SEGV)
     Timestamp: Mon 2022-04-25 11:50:14 WEST (2min 25s ago)
  Command Line: /home/steven/bin/dino
    Executable: /home/steven/bin/dino
 Control Group: /user.slice/user-1000.slice/session-2.scope
          Unit: session-2.scope
         Slice: user-1000.slice
       Session: 2
     Owner UID: 1000 (steven)
       Boot ID: d6cbf6539ac94d4298704dca5518e264
    Machine ID: 4558a59b797340a79d642c16a50fa29e
      Hostname: steven-thinkpad
       Storage: /var/lib/systemd/coredump/core.dino.1000.d6cbf6539ac94d4298704dca5518e264.3800000.1650883814000000.zst (present)
     Disk Size: 14.1M
       Message: Process 3800000 (dino) of user 1000 dumped core.
                
                Module /home/steven/bin/dino with build-id 86d77cd34561ef666f721a3e2e189ae8a8b00d26
                Module /home/steven/software/dino/build/libqlite.so.0.1 with build-id 79ed3c97e00cb0fbe31e17df2332d0e929f9b583
                Module /home/steven/software/dino/build/libxmpp-vala.so.0.1 with build-id 837bff72d09bd95ff0143896da33530824f7a7f7
                Module /home/steven/software/dino/build/libdino.so.0.0 with build-id 8534e4945c2f7edc6469b7687f657f673fe8202e
                Module linux-vdso.so.1 with build-id ae8518a0710c2679844504efb751b01abf13cef5
                Module p11-kit-trust.so with build-id 8b539aa1d5ee1ffc32e4c6f71c132e40e09ffbb7
[...]

PR #1259, opened 2022-08-15
Fix crash when calling contact from Conversations (#1227)

rozhuk-im:
Add null check for get_module() result


fiaxh:
Thanks.

This happens when the rtp plugin is disabled. While that setup should of course run without crashing, be aware that you won't be able to make calls with that build.

Issue #178, opened 2017-10-12
Mobile UI


zuglufttier:
Now that the Librem 5 got funded (https://puri.sm/shop/librem-5/), there might be a demand in the near future for a xmpp client that works on a small screen.

tophazus:
Not only Librem 5. There are people using Gajim on PocketCHIP (https://getchip.com/pages/pocketchip) with 480x272 pixels and the Pyra (https://pyra-handheld.com/) will also be ready soon™. There is a lot of demand for "small screen".

tigre-bleu:
Messages app for Gnome has some great mockups for messaging application on phones. See here: https://gitlab.gnome.org/Community/Design/app-mockups/blob/master/messages/messages.png

dos1:
In order to implement a responsive UI, Dino would just have to use appropriate widgets from the libhandy library: https://source.puri.sm/Librem5/libhandy

zuglufttier:
There are test image for the Librem 5 available: https://developer.puri.sm/Librem5/Development_Environment/Boards/qemu.html

So one could try to get things running already ;) First test made dino start but I couldn't even start one chat. Don't know what exactly happened as the whole VM became unresponsive.

dos1:
Running is not an issue. The UI is ;)

![Screenshot_20190717_163857](https://user-images.githubusercontent.com/126204/61384727-74935080-a8b1-11e9-8f0d-1a7b39a5d50e.png)

zuglufttier:
Yes... And I don't understand why Purism insisted to develop their own XMPP client... It will takes months or maybe years to implement all the features. Dino is now in a pretty good working state, I think.

ericschdt:
I would love to see a mobile Dino version for the PinePhone, which is now shipping and slowly creates a real demand for a good XMPP client on Linux phones. ;)

mar-v-in:
I started initial work on using libhandy in Dino for a better mobile UI. Feel free to try it out from the [feature/handy branch](https://github.com/dino/dino/tree/feature/handy).

thibaultamartin:
For anyone interested, the main window has been made adaptive, and [here's a recording of the app being resized and used at mobile size](https://mamot.fr/@thibaultamartin/103829274414213063).

Thanks a lot for tackling that issue, it's going to be a bliss on mobile!

masipcat:
Any ETA of this on master? Looks awesome!

singpolyma:
[...]

PR #1276, opened 2022-09-14
Make the main window adaptive with libadwaita

### LINKED ISSUE #1182: Consider move to libadwaita
GNOME 42 will move all core apps to use libadwaita, as it is the successor to libhandy. Moving Dino to libadwaita would help it to fit in with other GNOME applications

tpikonen:
Changes the Gtk.Paned widget in the main window to Adw.Leaflet from libadwaita. Fixes #1182.

Uses Adw.Application and a HeaderBarless Adw.Window for the main window. Changes the layout to two columns / vertical boxes with a single leaflet wigdet and two header bars (or title bars in the NonCSD version).

Adds a back button to the ConversationTitlebar interface and implements it in the concrete classes.

Also makes the about dialog into Adw.AboutWindow.

This is a step towards solving #178.

### DIFF
--- a/cmake/FindAdwaita.cmake
+++ b/cmake/FindAdwaita.cmake
@@ -1,0 +1,11 @@
+include(PkgConfigWithFallback)
+find_pkg_config_with_fallback(Adwaita
+    PKG_CONFIG_NAME libadwaita-1
+    LIB_NAMES libadwaita-1
+    INCLUDE_NAMES adwaita.h
+    )
+
+include(FindPackageHandleStandardArgs)
+find_package_handle_standard_args(Adwaita
+    REQUIRED_VARS Adwaita_LIBRARY
+    VERSION_VAR Adwaita_VERSION)
--- a/main/CMakeLists.txt
+++ b/main/CMakeLists.txt
@@ -11,6 +11,7 @@ find_packages(MAIN_PACKAGES REQUIRED
     GObject
     GTK4
     ICU
+    Adwaita
 )
 
 set(RESOURCE_LIST
--- a/main/data/conversation_list_titlebar_csd.ui
+++ b/main/data/conversation_list_titlebar_csd.ui
@@ -1,7 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <interface>
     <requires lib="gtk" version="4.0"/>
-    <object class="GtkHeaderBar" id="header_bar">
+    <requires lib="adw" version="1.0"/>
+    <object class="AdwHeaderBar" id="header_bar">
         <property name="hexpand">False</property>
         <style>
             <class name="dino-left"/>
--- a/main/data/theme.css
+++ b/main/data/theme.css
@@ -4,7 +4,9 @@
  */
 
 window.dino-main .dino-header-right {
-    background: @theme_base_color;
+    background: @insensitive_bg_color;
+    margin-top: 6px;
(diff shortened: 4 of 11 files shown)


selurvedu:
`Adw.AboutWindow` was introduced in libadwaita 1.2: <https://valadoc.org/libadwaita-1/Adw.AboutWindow.html>
```vala
[ Version ( since = "1.2" ) ]
```
Hence, you have to add this:
```patch
--- a/main/CMakeLists.txt
+++ b/main/CMakeLists.txt
@@ -11,7 +11,7 @@ find_packages(MAIN_PACKAGES REQUIRED
     GObject
     GTK4
     ICU
-    Adwaita
+    Adwaita>=1.2
 )
 
 set(RESOURCE_LIST
```

Otherwise it fails to build if libadwaita is older than 1.2.

selurvedu:
It may be a good idea to wrap the about window part with something like `#if ADWAITA_1_20` in the same manner as it's done with `#if VALA_0_50`, `#if GLIB_2_64`, `#if GST_1_16`, `#if GTK_4_6` etc.
That would let Dino build on e.g. older Fedora versions, but the Dino devs should know better, maybe they're planning to use libadwaita >= 1.2 features anyway, so this is not needed.

tpikonen:
> It may be a good idea to wrap the about window part with something like `#if ADWAITA_1_20` 

Thanks for the suggestion. As you said, that's for the maintainers to decide. Talking of which, it would be nice to get some kind of feedback on this PR. I've been using this without problems for close to 2 months and it's definitely an improvement over the `feature/handy` branch on mobile, mostly because swiping now works in the main window.

mar-v-in:
@tpikonen Sorry for not replying here yet. We're well aware if your work here and I have no doubts that it improves UX on mobile significantly, but we also have to make sure it doesn't hurt UX on desktop, especially on non-GNOME and non-Linux systems, which are not really in scope of libadwaita. For example we're typically using server-side decorations instead of client-side decorations on non-GNOME systems, and this seems to be no longer supported in libadwaita (at least not officially).

So the review process takes longer not because of your code, but because of the effects libadwaita has on all kinds of (non-GNOME) stuff that need to be reviewed first.

pravi:
@tpikonen thanks for your work on this, I just built it and started using it on my Librem 5 and loving it (I had to use a scaled down version of gajim till now which was really painful) ! Can you make adding multiple accounts adaptive too?

pravi:
Another thing I noticed is contact request is not yet adaptive. Should I keep reporting more issues I find here?

tpikonen:
> Another thing I noticed is contact request is not yet adaptive. Should I keep reporting more issues I find here?

I'm aware of (most of) the issues. I don't think it's necessary to report them here yet, since it's not known if this PR will be merged. Maybe you can collect them, and report if and when this is merged.

mossymossmoss:
Just here to say that this is some great work! Looking forward to running the Libadwaita version on my PinePhone!

pravi:
Just one request, if possible, I get a large number of spam, but I can't close any chats. On desktop I get a close button on hover but on mobile that close button never comes, may be on long press that close button can be highlighted? Or on the chat menu if we can close a chat (like in android conversations) that would be nice. Thanks again for this work.

mar-v-in:
I think most of this has been merged now manually, plus some stuff on top to fix a few issues we noticed. Thanks for your contribution @tpikonen.

PR #1282, opened 2022-09-29
Replace xmpp uris with the name of the person if they're in your roster

earboxer:
This is especially useful for users of Cheogram/JMP.chat for group MMSs, so you can know who is talking to you without having to memorize their cell numbers.

before

after

### DIFF
--- a/libdino/src/entity/settings.vala
+++ b/libdino/src/entity/settings.vala
@@ -12,6 +12,7 @@ public class Settings : Object {
         notifications_ = col_to_bool_or_default("notifications", true);
         convert_utf8_smileys_ = col_to_bool_or_default("convert_utf8_smileys", true);
         check_spelling = col_to_bool_or_default("check_spelling", true);
+        replace_xmpp_uris = col_to_bool_or_default("replace_xmpp_uris", true);
     }
 
     private bool col_to_bool_or_default(string key, bool def) {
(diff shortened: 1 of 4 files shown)


speedygonzales9:
Duplicate of PR #1212?

earboxer:
Yeah, looks like it is. My version has fewer features: it just replaces the links in the message view, not in the list view, but it *does* have a settings toggle to enable/disable this feature.

So... looking forward to either of these being merged (but not both ;)

speedygonzales9:
Are you planning to change how `xmpp:` links are displayed in a push notification if they are known? This would be nice to see.

dilinger:
Could probably work together to merge the two PRs, but I'd like to hear from the Dino maintainers that they'd actually be interested in merging either of them first. (I'm the author of #1212 )

fiaxh:
We do not think that it would be a good idea to just modify a message like this.

Someone might actually want to send `xmpp:user@example.com` as a URI. If the receiving client just replaces that text, it means that certain texts can't be sent anymore.

We generally aim to display messages as they were written by the sender, since the sender knows best what they want to say. If the receiving client just modifies the sender's text, that might break the original meaning. If the sender wants the receiving client to do something with the message, it should annotate the message stanza in a standardized way. We would consider supporting this if it was properly standardized.

Issue #1028, opened 2021-03-30
Can't start a chat with bare domain JID


SamWhited:
If you type "me@example.net" in the "Start Conversation" drop down when this JID is not in your roster the user shows up and you can begin chatting with them. However, if you just type "example.net" it does not show up and you can't start a conversation. However, domains are valid JIDs and administration bots may be running at them (for example, https://jmp.chat runs their account management bot at `cheogram.com` which you can't message in Dino without adding it to your roster first).

PR #1305, opened 2022-11-02
Allow Starting Conversation with JID Without Localpart

psycotica0:
Even though it's rare, it is possible for a JID to not have a localpart, especially components.

We should still be able to start conversations with this JID, though, without having to resort to another client.

That having been said, there may be a reason this was specifically excluded, so let me know if there's more stuff you'd like me to add to make this acceptable!

## Before
image

## After
image


singpolyma:
Fixes #1028

Issue #1293, opened 2022-10-18
Add an option to rejoin a disconnected muc


pravi:
An MUC appears greyed out when it was not online, but there was no option to reconnect to it from Dino. I have to rejoin it from Gajim to send messages.

pravi:
This is more serious than I initially thought as I was not getting any messages in a group I never left or wanted to leave. I had to go to Gajim and rejoin it manually to see the history.

pravi:
Can we please bookmark all the rooms we join by default or at least prompt to bookmark it or ignore it? Losing messages in groups we want to follow is a serious failing in a multi device world. Does Dino auto join all bookmarked rooms? Or it joins only in rooms where there was a message when Dino was online?

licaon-kter:
It joins all the rooms that have "auto join" flag setup.

How did you add the bookmark?

pravi:
@licaon-kter how do I add these flags in the first place? I probably joined by typing the full address or accepting invite or joined via another client. Not having an option to add bookmark itself is a bug I think.

licaon-kter:
Eg. Conversations add the flag when you add the bookmark, and afaik Dino does too.

pravi:
I have not seen an option to add bookmark in Dino. Is this "Persistent" toggle in an muc details?

licaon-kter:
Add bookmark is the Join action, press lower left plus, right?

pravi:
Oh, I never realized this is for adding bookmark. We should make it explicit I think. I used to just type the address and connect.

pravi:
And there should be an option to add bookmark after you join a channel too like conversations allows.

fiaxh:
Dino already does (or should do) the things you request:

- Dino automatically tries to rejoin a MUC if it detects that it isn't connected anymore via [MUC Self-Ping](https://xmpp.org/extensions/xep-0410.html)
- When you join a MUC in Dino, it automatically creates a bookmark for it with the autojoin flag set to `true`. This happens even if you don't go through the plus button that licaon-kter described. Just typing the address and clicking connect already creates this bookmark.

If you experience reconnection issues, it's not due to a lack of the features you mentioned but rather due to a potential bug or due to another issue.

pravi:
I had to rejoin every muc hosted on chat.yax.im (probably it was down for a longer time than dino retries).

PR #1310, opened 2022-11-10
Parse conference.password from XEP-0402 bookmarks

tpikonen:
The conference password was not parsed from XEP-0402 responses. This caused autojoin to password protected group chats to fail, and when joining manually, the password had to be typed in every time.

Probably helps with #1293.


fiaxh:
Thanks.

Issue #1326, opened 2022-12-15
Feature: Unread message indicator


143981:
Like this:
![mQluRgQeGYAeujgBX72ckNE6u4PDKd3p](https://user-images.githubusercontent.com/106692737/207905988-15f6c00e-6c98-4870-b251-0383603444d9.png)

This is really need without tray icon with indicator...

fiaxh:
This looks like a duplicate of #144.

Issue #144, opened 2017-09-05
prefix window title with number of unread messages so you can see it in the taskbar of the window manager e.g.


darix:
similar to what github and so do.

zinid:
Any chance this issue will even be considered? This is a stopper issue for me, because I hate popups. I also think this will not over-complicate the application's design or make it geeky.

RalfJung:
I agree this would be very useful. I often miss Gnome's notifications (or have them muted entirely since they interrupt my concentration). Other chat applications put an unread message counter in the window title or some indication into the icon so a quick glance at my taskbar and browser tab bar immediately tells me "I have some unread messages in Signal, Zulip, Matrix, and Mattermost; let's check those out". But dino doesn't do anything like that. I regularly completely miss messages because of that and then my contacts ask why I don't reply.

Maybe if Gnome's notification system worked better (e.g., grouping by application, like on Android), just sending a notification would be enough, but they way it works I usually have around 100 notifications at the end of a day and then just discard them all since it's not possible to go through all of them.

I hope https://github.com/dino/dino/pull/1336 can be merged :)

PR #1336, opened 2023-01-08
Unread message indicator #1326 #144

143981:


licaon-kter:
Pics?

143981:
> Pics?

fiaxh:
Regarding code structure:

- The ConversationSelectorRow is responsible for displaying a single row in the conversation list. Computing the total unread message count and calling an update routine on the application is very much outside of its responsibility.
- You put the function for updating the window title into the application class, however it would be more fitting to put it into the window logic.
- A suitable place for both the above would be the main window controller.
- Don't use the Conversation object as a storage space for the unread value. It is a representation of persisted data.

Also, there is no need to make a setting for this feature.

PR #1341, opened 2023-01-19
adding_tooltips

iampkumar02:
I have added tool-tips in chat.

image


fiaxh:
Thanks for your PR!

- I looked a bit into what wording other chat applications are using. Whatsapp, Signal and Telegram use "Delivered" and "Read" in their documentation, and I would suggest that we also follow that route.
- Please make the strings translatable by writing _("Delivered") instead of "Delivered"
- The tooltip for the encryption icon is a bit more complicated. For example, if OpenPGP is used, that icon would appear even though the user is verified. And we might want to name the encryption. I would suggest to leave it out of this PR since there is a dedicated PR for that topic.

main/src/ui/conversation_content_view/conversation_item_skeleton.vala
```diff
@@ -186,8 +187,10 @@ public class ConversationItemSkeleton : Plugins.ConversationItemWidgetInterface,
 
     private void update_received_mark() {
         switch (content_meta_item.mark) {
-            case Message.Marked.RECEIVED: received_image.icon_name = "dino-tick-symbolic"; break;
-            case Message.Marked.READ: received_image.icon_name = "dino-double-tick-symbolic"; break;
+            case Message.Marked.RECEIVED: received_image.icon_name = "dino-tick-symbolic";
+                received_image.set_tooltip_text ("Received"); break;
```
fiaxh, line 191: We usually format multi-line case-statements like this:

```suggestion
            case Message.Marked.RECEIVED:
                received_image.icon_name = "dino-tick-symbolic";
                received_image.set_tooltip_text("Received");
                break;
```

fiaxh:
Thanks.

In my last message I forgot to also note that we have to wrap tooltip strings in Util.string_if_tooltips_active(). This is a work-around for a bug in some GTK versions which sometimes leads to the application crashing. I just added that to this PR.

PR #1342, opened 2023-01-20
Pin on click

m-shinder:
I've added ability to pin chats directly from conversation list.
This involves several changes in application logic:
- Change signal used to select conversation.
`GtkListBox` have separate signals to select and activate rows, 
so, from semantics's perspective,  `row_selected` is more correct way to choose chat.
- Add pin logic to conversations themselves. 
First `pinned` is no more assumed as boolean value now, pin order matter. 
This allows to add option to rearrange pin in future.
I agree that it done in kinda messy way, but i did not found way to do it more correctly. If you can suggest, I'm  ready to fix it
- Change all external pin logic to call `pin` and `unpin` functions. This is simply point of this PR :)

### DIFF
--- a/libdino/src/entity/conversation.vala
+++ b/libdino/src/entity/conversation.vala
@@ -3,6 +3,11 @@ using Xmpp;
 namespace Dino.Entities {
 
 public class Conversation : Object {
+    private static Conversation[] list = new Conversation[0];
+    public delegate void Applicable(Conversation c);
+    public static void apply(Applicable func) {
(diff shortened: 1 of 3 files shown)


fiaxh:
With your change, a conversation is pinned when the user double-clicks the entry in the conversation list.
I don't think that would be a very helpful feature:

- Pinning is not something that is done very frequently, thus I wouldn't give it such prominent shortcut.
- The feature is not discoverable by the user.

PR #1344, opened 2023-01-21
QR code can be scanned properly

iampkumar02:
QR code is scaled up and can be scanned properly.


fiaxh:
Thanks for your PR! I just tried it out and it indeed scaled the QR code up.

But then I started wondering why we need to increase the pixel size in order for this to work. The QR code should™ already be much larger than it currently is. It seems that we were using the wrong Widget for our purpose. According to the GTK dokumentation

> GtkImage displays its image as an icon, with a size that is determined by the application. See [GtkPicture](https://docs.gtk.org/gtk4/class.Picture.html) if you want to show an image at is actual size.

Indeed, our issue was that the QR code was icon-sized and we wanted to display something at its actual size :slightly_smiling_face: I swapped out the Image for a Picture in c813a6d2405980c71450cefa10abdf11fab8e995 and now it works.

PR #1361, opened 2023-02-09
Use modern GNOME patterns for preferences

BrainBlasted:


licaon-kter:
Pics?

fiaxh:
Thanks for your PR, that definitively looks better than before :)

There was one issue, because you used the Switch.activate signal to listen for changes of the switch state. This does not always work. To quote the [documentation](https://valadoc.org/gtk+-3.0/Gtk.Switch.activate.html):

> The activate signal on GtkSwitch is an action signal and emitting it causes the switch to animate.
> Applications should never connect to this signal, but use the notify:active signal.

I changed that to use notify:active in e896f94.

PR #1363, opened 2023-02-10
Add Flatpak manifest

sonnyp:
Inspired by https://github.com/flathub/im.dino.Dino/blob/f68d2cf457fdce72ff86effe4df369d52083c07d/im.dino.Dino.yml

Credits shared with @Aman9das and @linkmauve

Closes https://github.com/dino/dino/pull/343

Uses JSON instead of yaml so it can be used with GNOME Builder (it's a long story... :smile: )

### DIFF
--- a/im.dino.Dino.json
+++ b/im.dino.Dino.json
@@ -1,0 +1,75 @@
+{
+    "id": "im.dino.Dino",
+    "runtime": "org.gnome.Platform",
+    "runtime-version": "44",
+    "sdk": "org.gnome.Sdk",
+    "command": "dino",
+    "finish-args": [
+        "--share=ipc",
+        "--socket=fallback-x11",
+        "--socket=wayland",
+        "--socket=pulseaudio",
+        "--socket=gpg-agent",
+        "--share=network",
+        "--device=all",
+        "--talk-name=org.freedesktop.Notifications"
+    ],
+    "modules": [
+        {
+            "name": "libsignal-protocol-c",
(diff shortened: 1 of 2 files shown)


im.dino.Dino.json
```diff
@@ -0,0 +1,76 @@
...
+    "finish-args": [
+        "--share=ipc",
+        "--socket=fallback-x11",
+        "--socket=wayland",
+        "--socket=pulseaudio",
+        "--socket=gpg-agent",
+        "--share=network",
+        "--device=all",
+        "--talk-name=org.freedesktop.Notifications",
+        "--device=dri"
```
AsciiWolf, line 16: This is a duplicate. There is already a `--device=all` specified.

rugk: Actually, why does it even need `--device=all`?

rugk (changes requested):

im.dino.Dino.json
sonnyp: Camera

sonnyp:
Dino is now on Flathub https://flathub.org/apps/details/im.dino.Dino

Thanks again to @Aman9das and @linkmauve 

I updated the manifest here to be the exact one used on Flathub. https://github.com/flathub/im.dino.Dino/blob/13fc3bf7ca4555e9b2c22b3317a5b4794f51ff7d/im.dino.Dino.yml

Except for `org.gnome.Platform` is using `master` here.

PR #1364, opened 2023-02-11
General: Refresh UI of contact/conversation details

BrainBlasted:


BrainBlasted:
Before:

image

After:

image

000exploit:
Looks perfect

licaon-kter:
Ok, now pics for a channel/group that you are the owner of? The moderator of? The admin of?

000exploit:
> Ok, now pics for a channel/group that you are the owner of? The moderator of? The admin of?

For example, admin of group on jabber.ru:

<details> 
  <summary>Admin of conference</summary>


image


</details>

BrainBlasted:
Ah, I need to add spacing to `main_box`

000exploit:
> Ah, I need to add spacing to `main_box`

And move from `Gtk.Button` with `flat` css styling to... `Adw.EntryRow`? `Gtk.Entry`?? `Gtk.EditableLabel`???

image

BrainBlasted:
I'm leaving that for a separate MR - the way rows are built needs to change fundamentally.

000exploit:
> the way rows are built needs to change fundamentally.

I suspected that they were not made easy by looking at the structure of the widgets in the inspector... Anyway, you're doing a great job! Thank You!

BrainBlasted:
Spacing added :)

000exploit:
It's better now
<details>
<summary>Still admin of conference on jabber.ru</summary>

image
</details>

licaon-kter:
Thanks

BrainBlasted:
Quick push because I noticed that I didn't mark the Toast text as translatable.

000exploit:
Just to demonstrate the preferences window:
<details>
<summary>Preferences</summary>

image
</details>

000exploit:
One little one. Here the address of the room is duplicated (title bar and under room name). Consider removing one of them.

tpikonen:
Looks great, but here's some small nitpicks:

 * I would use the 'view-more' icon (three vertical dots) instead of 'info' in the header bar.
 * When viewing details on a conference, the number in 'Message history' list item is ellipsized when it has 3 digits (the same probably applies to the 'Default number of history mesages returned by room' item):
message_hist
 * These dialogs are rather annoying to close on Phosh (swipe to open the window switcher, swipe up to close the dialog, tap to switch to the dino window again). Maybe there should be a close button at the end of the list?

earboxer:
I tried it out and 
1. There was no  icon for the button in the header bar (see @tpikonen's comment)
2. When I clicked on OMEMO for a contact, it spawned way too many windows (I had to kill both dino and my wm to recover)

BrainBlasted:
Late to responding:

> I would use the 'view-more' icon (three vertical dots) instead of 'info' in the header bar.

I used `info` here because that button isn't a menu. `view-more-symbolic` is reserved for menus generally and could be confusing for a user. CC: @bertob for input.

> These dialogs are rather annoying to close on Phosh (swipe to open the window switcher, swipe up to close the dialog, tap to switch to the dino window again). Maybe there should be a close button at the end of the list?

That is a phosh issue. Upstream we're working on a pattern to resolve this, and the general rule for apps is to not design around this issue until we get everything implemented upstream. Dino wouldn't be alone with this papercut.

> When viewing details on a conference, the number in 'Message history' list item is ellipsized when it has 3 digits (the same probably applies to the 'Default number of history mesages returned by room' item):

Is this a moderation item? I'm not seeing this in any of the rooms I'm in.

BrainBlasted:
Not really. We could use AdwEntryRow, but that requires more restructuring.

fiaxh:
Thanks for your work!

After looking at your UI refresh of the dialog, I also played a bit around with the dialog and ended up changing most of the previous dialog code. In addition to your improvements of working on smaller screens and using PreferenceRows, it also makes use of EntryRows and other more dedicated PreferenceRows, so the above mentioned ellipsizing-issue should be resolved. There are also dedicated buttons for some settings, and a new row for the channel topic. (e2c34bf2235c9f85fc91de9c0f1b74858f4ef89e)

PR #1388, opened 2023-03-14
FreeDesktopNotifier: Set notification categories

dos1:
This provides notifications servers some context on how to handle the notification.


main/src/ui/notifier_freedesktop.vala
```diff
@@ -120,6 +121,7 @@ public class Dino.Ui.FreeDesktopNotifier : NotificationProvider, Object {
         hash_table["sound-name"] = new Variant.string("phone-incoming-call");
         hash_table["urgency"] = new Variant.byte(2);
         hash_table["desktop-entry"] = new Variant.string(Dino.Application.get_default().get_application_id());
+        hash_table["category"] = new Variant.string("im");
```
mar-v-in, line 124: Isn't there anything more suitable for calls?

dos1: Not really: https://specifications.freedesktop.org/notification-spec/latest/ar01s06.html

Issue #396, opened 2018-07-03
Use meson build system


tintou:
It's fast, has great Vala support and is now the default build system for GNOME

dos1:
While I fully understand and support migrating from autotools to meson, Dino already uses CMake, which is just fine.

BaptisteGelez:
Hello! I would like to know if you would accept a PR that adds meson as a build system (and that removes CMake if you don't want to maintain both)? I've heard from some other contributors that using CMake was not that easy for them, and it probably discourages some people from contributing.

selurvedu:
JFYI, both https://github.com/benwaffle/vala-language-server and https://gitlab.gnome.org/esodan/gvls do not work without Meson build files. As a result, the code editor of GNOME Builder is unusable because it [relies on GVls](https://blogs.gnome.org/despinosa/2020/03/02/gvls-0-14-0-released/) for finding symbols.
Also, the run / debug feature [does not support CMake](https://gitlab.gnome.org/GNOME/gnome-builder/-/issues/498#note_631144), so using Builder as a frontend for gdb is not possible as well:

> We do have minimal CMake support, but it lacks run support (unless you're using Flatpak) due to deficiencies in extracting runnable executables from CMake. It's decidedly low priority for us because GNOME doesn't use CMake.

I'm not saying that CMake should be dropped in favour of Meson. I'm just sharing my findings to help anyone who bumps into similar issues with Vala language servers and/or GNOME Builder.

Prince781:
As the developer of VLS, I think you guys should check out https://github.com/Prince781/vala-language-server/issues/73 and give your input. It's possible that if you change your CMake-Vala addon code to use `add_custom_target()` instead of `add_custom_command()`, the language server would be able to work with your project. I don't know much about CMake though so your input is appreciated.

sojuz151:
My personal solution for using Vala language server is generating simple meson.build, good enough to get autosuggestion working. You need to copy some files so I created a python script for this. It works good enough 
```
#!/usr/bin/env python3
from pathlib import Path
subprojects = ['crypto-vala','libdino','main','plugins','qlite','xmpp-vala']
sources = []
customVapis = []
for sp in subprojects:
    for path in Path(sp+"/src").rglob('*.vala'):
        sources +=  ["'%s'"%(str(path))]
    for path in Path(sp+"/src").rglob('*.vapi'):
        customVapis +=  ["'%s'"%(str(path))]       
buildFiles = open("meson.build","w")
[...]

PR #1394, opened 2023-03-20
First steps of meson support

hrxi:
Basic configuration of qlite, xmpp-vala, the Dino library and the Dino application are supported. There's no support for the plugins.

This e.g. enables using the Vala language server.


hrxi:
CC #396.

fiaxh:
Thanks! :)

Issue #802, opened 2020-04-10
Option to remove avatar


Shungy:
Dino does not give an option to remove your avatar, only change it with another avatar. As a basic principle, any change to an account should be reversible. Please consider adding this option.

fiaxh:
Support for Avatar removal was added in https://github.com/dino/dino/commit/21ae42762d8a57da5cb1ec40b46e7510fc3121ad.

Issue #1039, opened 2021-04-19
Dino does not clear contact avatar on removal


Zash:
1. Open chat with a contact (someone who has an avatar published)
2. Have the contact remove their own avatar
3. Observe their avatar in Dino

Expect: Dino should revert to the default (ie colored square with a letter).

Actual: No change observed.

Dino version: 0.2.0~git86.20210412.328c3cf

Per [XEP-0084 3.5 Publisher Disables Avatar Publishing](https://xmpp.org/extensions/xep-0084.html#pub-disable) receiving the following indicates that a contact has removed their avatar:

```xml
<message id='bAQ4v0lrHFlpSZFB' type='headline' from='user@example.com'>
  <event xmlns='http://jabber.org/protocol/pubsub#event'>
    <items node='urn:xmpp:avatar:metadata'>
      <item id='5daf833b-b81b-4896-9e5c-689dd73627b6'>
        <metadata xmlns='urn:xmpp:avatar:metadata'/>
      </item>
    </items>
  </event>
</message>
```

With servers supporting more of XEP-0060 on users PEP services, it may also be good to handle [purge](https://xmpp.org/extensions/xep-0060.html#owner-purge) if it isn't too much trouble:

```xml
<message id='w46wOynPy-IFRdfm' type='headline' from='user@example.com'>
  <event xmlns='http://jabber.org/protocol/pubsub#event'>
    <purge node='urn:xmpp:avatar:metadata'/>
  </event>
</message>
```

Or even the [nuclear option](https://xmpp.org/extensions/xep-0060.html#owner-delete)...

fiaxh:
Incoming support for avatar deletion was added in fe45ab575c687febc1f342b0882a7597bd6ae9dc, and UI for deleting the avatar from Dino was added in 21ae42762d8a57da5cb1ec40b46e7510fc3121ad.

PR #1398, opened 2023-03-22
Remove avatar

iampkumar02:
- #802

- Here I have implemented the feature to remove avatar. Still it only changes for current instances. @fiaxh @mar-v-in  Can you tell how to make the changes permanent so that after creating new instance of application it persists the changes?


fiaxh:
Thanks for your PR.

The issue is that you are only removing the cached avatars. They will be simply loaded from the database and from the network on the next startup.

What you actually want to do is to remove the avatar from the user's server. The protocol used for that is defined in [XEP-0084](https://xmpp.org/extensions/xep-0084.html). There are two Pubsub nodes, one that stores the image data, and one "metadata" node that stores a reference to the other node. We should clear them both in order to remove the user's avatar.
Then, everyone (including us, although that is probably buggy, see #1039) will get the update from the server and clear the respective user avatar.

fiaxh:
Since this PR was opened as a GSoC teaser task and not adjusted since, I'll close it. If you want to continue working on this, we can of course reopen it.

PR #1399, opened 2023-03-22
Fix lack of /me rendering in quotes

### LINKED ISSUE #1390: /me doesn't get displayed when coupled to a reply
Version: 47a0669

## Steps to reproduce

- Send a reply (0461) which includes a /me command
- Observe from another Dino

## What happens
The /me message isn't displayed, only the reply.

## Expected behaviour
Both the reply and the /me message are displayed somehow

Epicalert:
Fixes #1390

adb2b58:

This PR:

### DIFF
--- a/main/src/ui/conversation_content_view/quote_widget.vala
+++ b/main/src/ui/conversation_content_view/quote_widget.vala
@@ -29,6 +29,10 @@ namespace Dino.Ui.Quote {
             if (content_item.type_ == MessageItem.TYPE) {
                 var message = ((MessageItem) content_item).message;
                 this.message = Dino.message_body_without_reply_fallback(message);
+
+		if (message.body.has_prefix("/me ")) {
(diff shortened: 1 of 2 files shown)


dvn0:
LGTM

main/src/ui/conversation_content_view/quote_widget.vala
```diff
@@ -29,6 +29,10 @@ namespace Dino.Ui.Quote {
             if (content_item.type_ == MessageItem.TYPE) {
                 var message = ((MessageItem) content_item).message;
                 this.message = Dino.message_body_without_reply_fallback(message);
+
+		if (message.body.has_prefix("/me ")) {
+			this.message = @"<i><b>$(Markup.escape_text(display_name))</b> " + this.message.substring(4) + "</i>";
```
fiaxh, line 34: Since the label now uses markup, you have to escape every string that you don't want to contain markup. That means, also the message. Otherwise, the message "/me <b>test</b>" would be displayed with a bold "test" instead of the actual message.

PR #1401, opened 2023-03-23
Fix a crash when joining a Slidge Telegram MUC

bodqhrohro:


bodqhrohro:
Eh, it's a duplicate of #1405 and is irrelevant now.

PR #1414, opened 2023-04-01
[GTK4] User-selectable default encryption settings

XaviDCR92:
## Notice

This PR rebases https://github.com/dino/dino/pull/1236 on top of latest `master`, GTK4/`libadwaita`-based branch. The description below has been copied from the original PR, while the screenshots have been updated accordingly.

## Background

Despite Dino claiming _"no unencrypted chats leaving your computer"_ on [its website](https://dino.im/), truth is this only applies to *transport*-level encryption, whereas **end-to-end encryption is disabled by default**, which is a huge privacy risk for users. So far, OMEMO or OpenPGP must be explicitly enabled by the user from the encryption widget below, for each new conversation:

image

Several attempts at solving this issue have already been proposed:
- https://github.com/dino/dino/pull/845
- https://github.com/dino/dino/pull/632
- https://github.com/dino/dino/pull/1035

Unfortunately, all of them were unsuccessful since they preferred OMEMO as the default encryption method, a decision some contributors did not particularly agree with.

## Overview

Therefore, this PR instead prompts the user for a default encryption method, and stores this choice persistently for any new conversations. The first time a user starts a new chat, the following modal dialog shows up:

image

### Features

- These settings are persistently stored into the database, and can be changed anytime from the settings menu, as shown below.

image

- The modal dialog above is not shown for public chat rooms, following the same logic as the [encryption button](https://github.com/dino/dino/blob/99c076254abc6e2b03b784732a76a389e5a4f801/main/src/ui/chat_input/encryption_button.vala#L93-L102), or when a default encryption method has been already selected.
- Previous encryption settings for each conversation are respected.
     - In order not to break existing behaviour, this means **unencrypted chats are left as such, and the dialog above will only appear for new conversations**.

## Additional notes

- Neutrality over the decision to pick OpenPGP or OMEMO as the default encryption method has been respected as much as possible for this PR.
    - As for the settings menu, "Ask" is selected by default, so the modal dialog above appears when the user starts a new conversation.
- As opposed to the settings menu, XEP numbers for OMEMO and OpenPGP have been referenced on the modal dialog so users can search them in order to get a more informed decision.
- `Encryption.parse` was not in fact required (plain `int` could have been stored into the database instead of `string`), but makes it more readable when inspecting the database.

I hope this PR finally solves this issue, so users can enjoy end-to-end encrypted chats by default!

### DIFF
--- a/main/data/default_encryption_dialog.ui
+++ b/main/data/default_encryption_dialog.ui
@@ -1,0 +1,77 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Generated with glade 3.22.2 -->
+<interface>
+  <requires lib="gtk+" version="3.6"/>
+  <object class="GtkDialog" id="dialog">
+    <property name="can_focus">True</property>
+    <property name="modal">True</property>
+    <property name="default_width">320</property>
+    <property name="default_height">260</property>
+    <child type="titlebar">
+      <placeholder/>
+    </child>
+    <child>
+      <object class="GtkBox">
+        <property name="can_focus">False</property>
+        <property name="orientation">vertical</property>
+        <child>
+          <object class="GtkBox">
+            <property name="visible">True</property>
+            <property name="can_focus">False</property>
+            <property name="orientation">vertical</property>
+            <child>
+              <object class="GtkLabel" id="default_encryption_warning_label">
+                <property name="visible">True</property>
+                <property name="can_focus">False</property>
+                <property name="label" translatable="yes">You are opening a new conversation without having set end-to-end encryption by default.
+
+It is strongly recommended to enable it to prevent your messages from being read by third parties.
+
+Please select an option to start this conversation with. Choosing one of the encryption methods will also set it as default in the global settings.
+</property>
+                <property name="wrap">True</property>
+              </object>
+            </child>
+            <child>
+              <object class="GtkCheckButton" id="omemo">
+                <property name="label" translatable="yes">OMEMO (automatic setup)</property>
+                <property name="visible">True</property>
+                <property name="can_focus">True</property>
+                <property name="receives_default">False</property>
+                <property name="active">False</property>
+              </object>
+            </child>
+            <child>
+              <object class="GtkCheckButton" id="openpgp">
+                <property name="label" translatable="yes">OpenPGP (external setup required)</property>
+                <property name="visible">True</property>
+                <property name="can_focus">True</property>
+                <property name="receives_default">False</property>
+                <property name="active">False</property>
+                <property name="group">omemo</property>
+              </object>
+            </child>
+            <child>
+              <object class="GtkCheckButton" id="none">
+                <property name="label" translatable="yes">Unencrypted</property>
+                <property name="visible">True</property>
+                <property name="can_focus">True</property>
+                <property name="receives_default">False</property>
+                <property name="active">False</property>
+                <property name="group">omemo</property>
+              </object>
+            </child>
+            <child>
+              <object class="GtkButton" id="accept_button">
+                <property name="label" translatable="yes">Apply</property>
+                <property name="visible">True</property>
+                <property name="can_focus">True</property>
+                <property name="receives_default">True</property>
+              </object>
+            </child>
+          </object>
+        </child>
+      </object>
+    </child>
+  </object>
+</interface>
--- a/libdino/src/entity/conversation.vala
+++ b/libdino/src/entity/conversation.vala
@@ -33,7 +33,7 @@ public class Conversation : Object {
             }
         }
     }
-    public Encryption encryption { get; set; default = Encryption.NONE; }
+    public Encryption encryption { get; set; default = Encryption.UNKNOWN; }
     public Message? read_up_to { get; set; }
     public int read_up_to_item { get; set; default=-1; }
 
--- a/libdino/src/entity/encryption.vala
+++ b/libdino/src/entity/encryption.vala
@@ -11,6 +11,27 @@ namespace Dino.Entities {
         public bool is_some() {
             return this != NONE;
         }
+
+        public static Encryption parse(string str) {
+            switch (str) {
+                case "DINO_ENTITIES_ENCRYPTION_NONE":
+                    return NONE;
+                case "DINO_ENTITIES_ENCRYPTION_PGP":
+                    return PGP;
+                case "DINO_ENTITIES_ENCRYPTION_OMEMO":
+                    return OMEMO;
+                case "DINO_ENTITIES_ENCRYPTION_DTLS_SRTP":
+                    return DTLS_SRTP;
+                case "DINO_ENTITIES_ENCRYPTION_SRTP":
+                    return SRTP;
+                case "DINO_ENTITIES_ENCRYPTION_UNKNOWN":
+                    // Fall through.
+                default:
+                    break;
+            }
+
+            return UNKNOWN;
+        }
     }
 
 }
--- a/libdino/src/entity/settings.vala
+++ b/libdino/src/entity/settings.vala
@@ -12,11 +12,18 @@ public class Settings : Object {
         notifications_ = col_to_bool_or_default("notifications", true);
         convert_utf8_smileys_ = col_to_bool_or_default("convert_utf8_smileys", true);
         check_spelling = col_to_bool_or_default("check_spelling", true);
+        default_encryption = col_to_encryption_or_default("default_encryption", Encryption.UNKNOWN);
     }
 
     private bool col_to_bool_or_default(string key, bool def) {
         string? val = db.settings.select({db.settings.value}).with(db.settings.key, "=", key)[db.settings.value];
         return val != null ? bool.parse(val) : def;
+    }
+
+    private Encryption col_to_encryption_or_default(string key, Encryption def) {
+        var sval = db.settings.value;
+        string? val = db.settings.select({sval}).with(db.settings.key, "=", key)[sval];
+        return val != null ? Encryption.parse(val) : def;
     }
 
     private bool send_typing_;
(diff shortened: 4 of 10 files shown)


main/data/default_encryption_dialog.ui
```diff
@@ -0,0 +1,78 @@
...
+              <object class="GtkLabel" id="default_encryption_warning_label">
+                <property name="visible">True</property>
+                <property name="can_focus">False</property>
+                <property name="label" translatable="yes">No default end-to-end encryption method has been previously selected for this conversation. XMPP defines some XEP for end-to-end encryption that Dino supports.
+
+It is strongly recommended that one of the following end-to-end encryption methods below is selected before sending a message.
+
+TAKE INTO ACCOUNT UNENCRYPTED CONVERSATIONS COULD BE READ BY THIRD PARTIES.
+
+These settings can be later changed from the Settings menu.</property>
```
mbeko, line 32: The text is long, technical, contains redundancy and partly uses all-capital letters that are usually perceived as shouting. It is also slightly misleading by mixing global default encryption and  encryption for the given conversation.

This could potentially confuse or annoy users and doesn't match the simple interface of Dino.

```suggestion
                <property name="label" translatable="yes">You are opening a new conversation without having set end-to-end encryption by default.

It is strongly recommended to enable it to prevent your messages from being read by third parties.

Please select an option to start this conversation with. Choosing one of the encryption methods will also set it as default in the global settings.</property>
```

main/data/default_encryption_dialog.ui
```diff
@@ -0,0 +1,78 @@
...
+                <property name="wrap">True</property>
+              </object>
+            </child>
+            <child>
+              <object class="GtkCheckButton" id="omemo">
+                <property name="label" translatable="yes">OMEMO (XEP-0384)</property>
```
mbeko, line 38: XEP is a technical term and an abbreviation at that. In addition with the number, that can easily confuse users without technical background. If they don't understand it, they can ignore it, choose an encryption method randomly and hope for the best. Or not encrypt and deal with the bad conscience of having chosen the discouraged option. Or search on the internet and read long XEPs that are meant for developers, not end users who just want to chat with their grandchildren.

Dino's simple interface is aimed at such users, so we have to expect a considerable amount of people who don't care much about the technical details and will choose an encryption method if it doesn't get in their way.

Everything those people need to know is that OMEMO will do everything automatically for them while for OpenPGP, they have to set up the keys with an external program.

If someone is really interested in the details, they can just type "OMEMO" in a search engine and will get a lot of information, both the XEP number and simpler descriptions targeted at end users.

```suggestion
                <property name="label" translatable="yes">OMEMO (automatic setup)</property>
```

main/data/default_encryption_dialog.ui
```diff
@@ -0,0 +1,78 @@
...
+                <property name="visible">True</property>
+                <property name="can_focus">True</property>
+                <property name="receives_default">False</property>
+                <property name="active">False</property>
+              </object>
+            </child>
+            <child>
+              <object class="GtkCheckButton" id="openpgp">
+                <property name="label" translatable="yes">OpenPGP (XEP-0027)</property>
```
mbeko, line 47: See my other comments for the reasoning behind the suggestion
```suggestion
                <property name="label" translatable="yes">OpenPGP (manual setup)</property>
```

Alternatively this, but it might be too long:
```suggestion
                <property name="label" translatable="yes">OpenPGP (external setup required)</property>
```

main/data/default_encryption_dialog.ui
```diff
@@ -0,0 +1,78 @@
...
+                <property name="visible">True</property>
+                <property name="can_focus">True</property>
+                <property name="receives_default">False</property>
+                <property name="active">False</property>
+                <property name="group">omemo</property>
+              </object>
+            </child>
+            <child>
+              <object class="GtkCheckButton" id="none">
+                <property name="label" translatable="yes">Unencrypted (not recommended)</property>
```
mbeko, line 57: The description text already mentions that it's strongly discouraged, we don't have to rub it in. Repeating isn't likely to convince anybody, it just feels more aggressive or even patronising
```suggestion
                <property name="label" translatable="yes">Unencrypted</property>
```

main/data/default_encryption_dialog.ui
```diff
@@ -0,0 +1,78 @@
...
+                <property name="visible">True</property>
+                <property name="can_focus">True</property>
+                <property name="receives_default">False</property>
+                <property name="active">False</property>
+                <property name="group">omemo</property>
+              </object>
+            </child>
+            <child>
+              <object class="GtkButton" id="accept_button">
+                <property name="label" translatable="yes">Accept</property>
```
mbeko, line 67: This verb feels a bit out of place here. Normally, "accept" is used in dialogs where you just acknowledge an information or you choose between accepting and rejecting/cancelling.

You rather make a configuration and tell the app to apply it.

```suggestion
                <property name="label" translatable="yes">Apply</property>
```

main/data/default_encryption_dialog.ui
XaviDCR92: I agree your suggestion sounds less aggressive, while still making the point clear. Thanks!

main/data/default_encryption_dialog.ui
XaviDCR92: > XEP is a technical term and an abbreviation at that. In addition with the number, that can easily confuse users without technical background.

My intention was to make clear Dino [currently](https://github.com/dino/dino/wiki/Supported-XEPs) only supports [XEP-0027](https://xmpp.org/extensions/xep-0027.html), which is considered obsolete, supposedly in favor of more modern XEPs such as [XEP-0373](https://xmpp.org/extensions/xep-0373.html) or [XEP-0374](https://xmpp.org/extensions/xep-0374.html).

But I agree it sounds _too_ technical for non-tech-savvy users, thus I find your suggestion better. I just hope it does not fail to maintain the neutrality between choosing either OMEMO or OpenPGP, as this has been a reportedly controversial topic within the Dino community.

main/data/default_encryption_dialog.ui
XaviDCR92: As opposed to `manual setup`, I think `external setup required`, while longer, makes clear the fact external tools must be used to set up OpenPGP encryption.

main/data/default_encryption_dialog.ui
XaviDCR92: I agree with your suggestion. Thanks!

main/data/default_encryption_dialog.ui
XaviDCR92: I prefer the original suggestion, but I agree to change it if it sounds too aggressive. Thanks!

main/data/default_encryption_dialog.ui
mbeko: I see. I think that users with sufficient knowledge and interested in the supported XEPs will be able to [look it up on the website](https://dino.im/#features). Therefore, personally, I wouldn't add that information in the UI. But if you would like to do so, here are two ideas:

* Make the option texts links to the related XEPs, though I'm not sure if this is possible with GTK
  * [OMEMO](https://xmpp.org/extensions/xep-0384.html) (automatic setup)
  * [OpenPGP](https://xmpp.org/extensions/xep-0027.html) (external setup required)
*  Add a question mark icon in the corner of the dialogue that provides a short technical explanation when hovered or clicked on

I wrote my suggestions intentionally in a neutral way, i.e. the texts in parentheses describe only facts. Those are chosen in a way to be relevant for less technical users and easy to understand. So I think it's fine regarding neutrality, but of course it's just suggestions from my side and in the end you decide :slightly_smiling_face:

XaviDCR92: >- Make the option texts links to the related XEPs, though I'm not sure if this is possible with GTK 

I am not sure, either. However, I agree tech-savvy users would look up that information somewhere else.

> - Add a question mark icon in the corner of the dialogue that provides a short technical explanation when hovered or clicked on

Sounds like a good suggestion. I will have a look at it ASAP.

> I wrote my suggestions intentionally in a neutral way, i.e. the texts in parentheses describe only facts.

And I also think these decisions are neutral. Considering how many PRs related to this topic have already been either rejected or ignored because of neutrality reasons between OMEMO and OpenPGP, I just wanted to make sure this PR is not affected by this.

rms-joe:
can't wait to see this merged. just happened again, while me usin dino, the other end tells me i'm sending unecrypted.

@XaviDCR92 is there no option to set "none" as default?
so non encrypters needs to answer request every time?

or misinterpretation from my side?

XaviDCR92:
> is there no option to set "none" as default?
> so non encrypters needs to answer request every time?

You are somewhat right - users who want all of their conversations unencrypted will be asked every time a _new_ conversation is opened. In other words, this means a user would not be notified again if it was already decided to keep a  conversation unencrypted.

While I would not disagree with changing the existing behaviour to accommodate this new use case, I would first prefer the main devs from this project to show some actual interest in this PR. It has been ready for review for over 6 months now, and its [GTK3 equivalent](https://github.com/dino/dino/pull/1236) was also eventually ignored back in the day.

Jieiku:
This would resolve one of a few issues that stops me from moving from pidgin to dino.

Jieiku:
If anyone that likes this pull request decides to fork and rename the Dino project, please let me know. I am interested in contributing to Dino, but not when pull requests that I and many others are interested in get ignored for over a year.

I can understand not merging a pull request if there is an issue with it, but there is no feedback or comments from a maintainer at all...

XaviDCR92:
A few months ago, I already started a (still unbranded) [fork of Dino](https://gitea.privatedns.org/xavi/dino). As opposed to upstream, it is still based on GTK3 and provides some of the pull requests that did not make it into upstream, such as the [GTK3 equivalent to this PR](https://github.com/dino/dino/pull/1236), as well as some experimental features.

I would find it interesting if more people would join the effort into improving this application, even if that means departing from the original project.

Jieiku:
Excellent, I am going to check it out. I will review the code and try to figure out what is necessary to add systray support unless you already did that.

mxlgv:
@XaviDCR92 hello! Noticed strange behavior. The window seems to be detached from the main window, it seems you forgot to use the set_transient_for() method for your Dialog. In addition, maximizing and minimizing dialogue seems completely inappropriate.

josephmturner:
> @XaviDCR92 is there no option to set "none" as default?
> so non encrypters needs to answer request every time?

Thanks for this PR!  My two cents as a user of JMP (where encryption with the SMS bridge is pointless) is that an option to select "none" as default is an important addition.

fiaxh:
OMEMO by default was added as an option as part of https://github.com/dino/dino/commit/680f0dd0a768b201b62899905c38210e834dde91.

PR #1428, opened 2023-04-22
Focus ChatInput textbox after selecting emoji

### LINKED ISSUE #671: There's no cursor if the input field is clicked while the emoji dialog is opened
1. Open the emoji dialog
2. Click into the input field

The emoji window closes but no cursor is displayed in the input field. However, you can actually start typing in the input field and the cursor comes back.

kmq:
After selecting an emoji, the emoji is inserted
into the textbox, but focus remains on the emoji_button.

This causes the EmojiChooser to be opened again if a user hits the Enter key directly, but text is inserted into the textbox if they continue to type.

This commit just explicitely focuses on the textbox after an emoji has been selected.

This also fixes #671


fiaxh:
Thanks!

PR #1431, opened 2023-04-24
Force resync button

bodqhrohro:


fiaxh:
As far as I see, you added an option for the user to explicitly sync all history of a contact that is available on the server.

Since you wrote that code, I assume that you're having problems with MAM sync? If there is a sync issue, this code doesn't solve the problem. It would only add a work around that would require the user to first realize that messages are missing, and then take manual action. That is not the user experience we want to have in Dino. Instead, we should fix the actual problem, and make MAM sync work reliably.

After we implemented MUC MAM sync there were a couple of issues, and we already fixed the ones that we found. Personally, I'm not aware of sync issues still existing, but that doesn't mean they don't exist. If you still experience problems, it would be great if you could look into when they happen. We have quite some [debug logs](https://github.com/dino/dino/wiki/Debugging#debug-output) on MAM sync which might help to identify the issue.

PR #1432, opened 2023-04-24
Add a keyboard shortcut to show keyboard shortcuts

eerielili:
Hello,

I like having a keyboard shortcut to show keyboard shortcuts, so here's that!
The shortcut would be `Ctrl+?`.

Kind regards,
eerielili


fiaxh:
Thanks.

PR #1445, opened 2023-05-23
A simple fix for I2P and Lokinet.

PoneyClairDeLune:
A simple fix for I2P and Lokinet.

The job of maintaining TLDs for mixnet TLDs can become overwhelming, just saying. Better to just offer a togglable option to allow untrusted certificates.

### DIFF
--- a/libdino/src/service/connection_manager.vala
+++ b/libdino/src/service/connection_manager.vala
@@ -382,9 +382,11 @@ public class ConnectionManager : Object {
     }
 
     public static bool on_invalid_certificate(string domain, TlsCertificate peer_cert, TlsCertificateFlags errors) {
-        if (domain.has_suffix(".onion") && errors == TlsCertificateFlags.UNKNOWN_CA) {
+        if ((domain.has_suffix(".onion") || domain.has_suffix(".i2p") || domain.has_suffix(".loki")) && errors == TlsCertificateFlags.UNKNOWN_CA) {
             // It's barely possible for .onion servers to provide a non-self-signed cert.
-            // But that's fine because encryption is provided independently though TOR.
+            // But that's fine because encryption is provided independently though Tor.
+            // Same for .i2p servers on I2P, and .loki servers on Lokinet.
+            // Although maintaining exceptions for mixnet TLDs could become overwhelming soon.
             warning("Accepting TLS certificate from unknown CA from .onion address %s", domain);
             return true;
         }


PoneyClairDeLune:
> Is there a way to request the domain's suffix e.g. `domain.get_suffix()` and check it if it is in an array/list of suffixes instead? This would simplify the code a bit. Then other proxified suffixes can be easily added to the array/list.

I know nothing about Vala, I just treated Vala as C when I was writing that code and was afraid of breaking anything. Maybe some others could get it into an array instead.

XutaxKamay:
Would be great to add snode aswell if that's possible

libdino/src/service/connection_manager.vala
```diff
@@ -382,9 +382,11 @@ public class ConnectionManager : Object {
     }
 
     public static bool on_invalid_certificate(string domain, TlsCertificate peer_cert, TlsCertificateFlags errors) {
-        if (domain.has_suffix(".onion") && errors == TlsCertificateFlags.UNKNOWN_CA) {
+        if ((domain.has_suffix(".onion") || domain.has_suffix(".i2p") || domain.has_suffix(".loki")) && errors == TlsCertificateFlags.UNKNOWN_CA) {
```
eerielili, line 385: as @Quix0r pointed, you could something like
```vala
string [] domain_exceptions = {".onion", ".i2p", ".loki", ".etc"} 

foreach (var domain_exception in domain_exceptions) {
    if ( domain.has_suffix(domain_exception) && errors == TlsCertificateFlags.UNKNOWN_CA) {
      //do stuff
      break
    }
}
```

PoneyClairDeLune: I'll do that shortly.

eerielili: @PoneyClairDeLune Nice, and I wanted to add too that the `domain_exceptions` variable should have a more apt name like `tld_exceptions`, I think it's more readable

PoneyClairDeLune: @eerielili Done in https://github.com/dino/dino/pull/1445/commits/e4da09f90512e20c87c848b64eaa5844df7cec4a

eerielili: @eerielili Nice. Yeah, you should place inside the function, it looks cleaner and it'll work the same since they'll be in the same scope.
I'm not too big on OOP but I suggest declaring it as `private const string[]` at the top of the class _if_ needed in other functions, then you could call it with `this.tld_exceptions`.

PoneyClairDeLune:
> Would be great to add snode aswell if that's possible

@XutaxKamay What's an "snode"?

XutaxKamay:
> > Would be great to add snode aswell if that's possible
> 
> @XutaxKamay What's an "snode"?
> 
> Edit: Is it something related to the "Snet" in Cuba?

https://git.lokinet.io/lokinet/lokinet/src/branch/dev/docs/dns-overview.md

libdino/src/service/connection_manager.vala
sergiuprt: late to the party, but, can you add the `.local` to the list?

PoneyClairDeLune: @sergiuprt Just did it in [`859ead5d`](https://github.com/dino/dino/pull/1445/commits/859ead5db77cb019161bf94f198eb8cfb31cf6d9), but at this rate I doubt the Dino team would merge it in the foreseeable future. I've long been away from the XMPP ecosystem.

PR #1453, opened 2023-06-21
Make everything compilable with meson

hrxi:
Still missing for feature parity with the CMake build:
- Disabling plugins.
- Detecting versions of libsoup.
- Detecting versions of gstreamer.
- Selecting media types for gstreamer.
- Building tests.

I think this could be merged on its own, however.


meson.build
```diff
@@ -1,18 +1,38 @@
...
 dep_icu_uc = dependency('icu-uc')
 dep_libadwaita = dependency('libadwaita-1')
+dep_libcanberra = dependency('libcanberra')
+dep_libgcrypt = dependency('libgcrypt')
+dep_libqrencode = dependency('libqrencode')
+dep_libsrtp2 = dependency('libsrtp2')
+# libsignal-protocol-c has a history of breaking compatibility on the patch level
+# we'll have to check compatibility for every new release
+# distro maintainers may update this dependency after compatibility tests
+dep_libsignal_protocol_c = dependency('libsignal-protocol-c', version: '>=2.3.2,<2.3.4')
```
eli-schwartz, line 30: Probably this should be an array of versions:

```suggestion
dep_libsignal_protocol_c = dependency('libsignal-protocol-c', version: ['>=2.3.2'], ['<2.3.4'])
```

hrxi: Thanks! Done.

PR #1455, opened 2023-07-04
application: Automatically load help overlay

BrainBlasted:
GTK automatically loads certain resources for you when using the right tools. This MR
changes Dino's main window to use AdwApplicationWindow and the shortcuts.ui file
into help-overlay.ui. This is so GTK can automatically load the file, set up the
action, and set the keyboard shortcut for the action.

See https://docs.gtk.org/gtk4/class.Application.html#automatic-resources


fiaxh:
Thanks, I didn't know about automatic resources.

PR #1456, opened 2023-07-05
Ignore non-DTLS data before handshake is complete

singpolyma:
https://datatracker.ietf.org/doc/html/rfc9147#name-processing-guidelines
https://datatracker.ietf.org/doc/html/rfc5764#section-5.1.2

If data is received before handshake is complete, discard it rather than forwarding it blindly to GnuTLS which can get confused.

### DIFF
--- a/plugins/ice/src/dtls_srtp.vala
+++ b/plugins/ice/src/dtls_srtp.vala
@@ -38,7 +38,11 @@ public class Handler {
     }
 
     public uint8[]? process_incoming_data(uint component_id, uint8[] data) throws Crypto.Error {
-        if (srtp_session.has_decrypt) {
+        if (data[0] >= 128) {
+            if (!srtp_session.has_decrypt) {
+                debug("Received data before SRTP session is ready, dropping.");
+                return null;
+            }
             if (component_id == 1) {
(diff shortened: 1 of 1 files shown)


mar-v-in:
Hi @singpolyma. Thanks for looking into this.

I have two questions with this patch in mind:
- Why is JMP sending data that it expects the other end to discard? Wouldn't it be more efficient if you discard it on the JMP side?
- What RTP data is JMP sending before the DTLS-SRTP handshake completed? Without the handshake completed, that data can't be encrypted (or at least not using any sensible keys), which rings alarm bells on my end. If JMP is sending data unencrypted until the handshake is finished, an active attacker can just drop the DTLS datagrams to be able to listen into the audio stream send by JMP.

singpolyma:
If you look at the first link ( https://datatracker.ietf.org/doc/html/rfc9147#name-processing-guidelines ) you see the sketch of what happens:

> it is possible for records that are protected with the new epoch to be received prior to the completion of a handshake. For instance, the server may send its Finished message and then start transmitting data. Implementations MAY either buffer or discard such records

Basically, Asterisk is too-eager and starts sending data as soon as *it* thinks the handshake is done, but possibly before Dino has received the Finish message, and because of UDP some of this data can easily arrive before the finish message itself.

PR #1462, opened 2023-07-11
fix: Add x node to MUC PM stanza

### LINKED ISSUE #1306: Dino does not sent the </x> in a MUC PM's stanza
Relevant issue on Gajim: https://dev.gajim.org/gajim/gajim/-/issues/11268

### Issue

1. Open Dino and join a MUC
2. Join the same MUC with Gajim
3. Send a PM to someone in the MUC from Dino

### Result

Sent PM from Dino looks like the following:
```
XMPP OUT [MyJID@conversations.im stream:0x4aa19d0 thread:0x187f350 2022-11-04T12:35:16+0100]
<message id='80aa7d97-05a3-4913-874b-30703b24718e' to='gajim@conference.gajim.org/dino' type='chat'>
  <body>
    Hello from Dino!
  </body>
  <request xmlns='urn:xmpp:receipts' />
  <active xmlns='http://jabber.org/protocol/chatstates' />
  <markable xmlns='urn:xmpp:chat-markers:0' />
</message>
```

It does not contain `<x xmlns='http://jabber.org/protocol/muc#user' />` and thus Gajim handles the message wrong. Though as @larma pointed out:
> because this requirement was only added in revision 1.28 of this XEP, receiving entities MUST NOT rely on the existence of the <x/> element on private messages for proper processing.

https://xmpp.org/extensions/xep-0045.html#privatemessage

mesonium:
Add <x/> tag in MUC-PMs to support better Carbon delivery in compliance with XEP-0045 v1.28 and above.

Fixes #1306

### DIFF
--- a/libdino/src/service/muc_manager.vala
+++ b/libdino/src/service/muc_manager.vala
@@ -653,7 +653,7 @@ public class MucManager : StreamInteractionModule, Object {
     }
 
     private void on_build_message_stanza(Entities.Message message, Xmpp.MessageStanza message_stanza, Conversation conversation) {
-        if (conversation.type_ == GROUPCHAT_PM) {
+        if (conversation.type_ == Conversation.Type.GROUPCHAT_PM) {
             Xmpp.Xep.Muc.add_muc_pm_message_stanza_x_node(message_stanza);            
         }
     }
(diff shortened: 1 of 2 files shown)


fiaxh:
Thanks! :slightly_smiling_face:

Issue #1029, opened 2021-04-05
Expose MUC roles/affiliations in the UI


VShell:
I cannot work out how to make someone a moderator, admin, owner, or member in Dino. I would like this to be available in the members list in a MUC.

pravi:
In Gajim, this is exposed when right clicking on a member in the members list. In Dino this could be added as right click and/or in the options that come on left click (send private message).

kousu:
In Conversations, the UI for writing membership bits looks like

<img src=https://user-images.githubusercontent.com/987487/212204383-8afb5443-9c62-41eb-86d2-26d39cb98360.jpg width=200>


Indeed, Dino only *reads* membership bits, e.g. it can tell the difference between these cases

![2023-01-12-184039_205x289_scrot](https://user-images.githubusercontent.com/987487/212204287-00eb4445-090b-4618-b17e-d38af83bfd94.png)
![2023-01-12-184054_197x289_scrot](https://user-images.githubusercontent.com/987487/212204290-cd1371b8-192b-4e0a-b8d6-b0b2f66936bd.png)

Clicking on a user's name.should allow promotion/demotion.

PR #1465, opened 2023-07-14
Fix showing the kick option to owners

Zash:
Missing case in the switch defaulted to returning false for Owners, thus preventing they with the most privileges from using those privileges.


fiaxh:
Thanks.
In the future, I think we should remove the kick function and offer banning instead.

Zash:
So mostly the affiliation part of #1029 then.

PR #1034, opened 2021-04-10
Add unread indicator

aereaux:
This should close #192

This is my first time working on dino or writing any Vala, so let me know if there's anything that's not great here.  It seems to work fine with my small amount of testing, but I'm not sure what to use for the unread indicator, right now I just have a Label as a placeholder.

Possible improvements:
- [x] Currently if you have a conversation open and new messages arrive, the unread indicator will not appear.  Not sure if this is desired.  Perhaps just when unfocused, more matching the behavior of the unread number in the convo list
- [ ] Jump to the unread message if it would be off the screen

PR #1473, opened 2023-08-09
Add unread indicator

mid1221213:
This is an edit of #1034 with some updates to the Gtk4 version of Dino, and an addition: the unread indicator is updated whenever the conversation is unread and not only when switching / creating conversations.

Thanks to the author of the original PR (@aereaux) who made most of this contrib!

Disclaimer: I don't know Vala, at all, this is my first contribution to a project in this language, this PR may contain complete non-sense, but at least it does what it says ;-)

### DIFF
--- a/libdino/src/entity/conversation.vala
+++ b/libdino/src/entity/conversation.vala
@@ -36,6 +36,8 @@ public class Conversation : Object {
     public Encryption encryption { get; set; default = Encryption.NONE; }
     public Message? read_up_to { get; set; }
     public int read_up_to_item { get; set; default=-1; }
+    public bool been_focused_out { get; set; default = true; }
+    public bool been_focused_out_and_in { get; set; default = true; }
 
     public enum NotifySetting { DEFAULT, ON, OFF, HIGHLIGHT }
     public NotifySetting notify_setting { get; set; default = NotifySetting.DEFAULT; }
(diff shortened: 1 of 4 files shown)


aereaux:
I don't use dino much any more, thanks for taking this over!

aereaux:
This has been in progress for quite a while.  It seems pretty good to me trying it out, is there anything else that needs to be done for it?

aereaux:
I've been using this now in my Dino build, works well except for the fact that when I'm reading a conversation and a new message gets added later the unread marker message disappears.  Any idea on the best way to fix this?  I might get a chance to take a look at it in a bit

mid1221213:
> I've been using this now in my Dino build, works well except for the fact that when I'm reading a conversation and a new message gets added later the unread marker message disappears. Any idea on the best way to fix this? I might get a chance to take a look at it in a bit

Sorry, I can't see a quick fix to this issue. To fix that, I think the values computed at the beginning of `update_unread_indicator()` should be saved somewhere, and then used, and reset when the conversation is changed.

aereaux:
I've been using the last version you uploaded.  Works great, thanks!

fiaxh:
Thanks for the PR.

This implementation has some issues:
- The conversation object is a representation of the database and shouldn't be used to pass around temporary variables
- The conversation view shouldn't contain code for specific conversation items, they go into their own classes.
- I think it should be possible to simplify the unread-line-moving logic somewhat (see commit)
- There are some runtime criticals due to variables being accessed before they are set
- Style: The line and the text have different colors (because the line is being greyed-out), there is no spacing between the line and text

I changed those things and pushed it in 4cc7e076e65c604e7a2f840ddc394ea70700577d

mid1221213:
Thanks a lot for the review, the comments and the commit :+1:

PR #1474, opened 2023-08-15
Obfuscate password length in account dialog (fixes #797)

### LINKED ISSUE #797: Password field shouldn't indicate password length

eerielili:
Closes issue #797.
By default, it displays a password length of 12 characters regardless of the real password length.

### DIFF
--- a/main/src/ui/manage_accounts/dialog.vala
+++ b/main/src/ui/manage_accounts/dialog.vala
@@ -44,8 +44,10 @@ public class Dialog : Gtk.Dialog {
         image_button.clicked.connect(show_select_avatar);
         alias_hybrid.entry.changed.connect(() => { selected_account.alias = alias_hybrid.text; });
         password_hybrid.entry.changed.connect(() => {
-            if (password_hybrid.text != "************") {
-                selected_account.password = password_hybrid.text;
+            string? pw_buffer;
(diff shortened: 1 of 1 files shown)


fiaxh:
This should be solved as part of 680f0dd0a768b201b62899905c38210e834dde91, which is a rewrite of the accounts and settings dialogs. Previously, it wasn't clear what the password entry field actually does. Now, passwords are only displayed (in dots) when the user enters them. Previously, editing the password field only made sense after the password was changed on another device. This case is now explicitly handeled like this:

eerielili:
@fiaxh Neat!

Issue #1321, opened 2022-11-24
allow closing a chat from the chat screen hamburger menu


pravi:
This will improve the usability along with #1276 on mobile screens and for touch screens. Currently closing a chat is possible only by hovering on a chat with a mouse and not possible on a touch screen.

pravi:
Since we have libadwaita support merged, we need a way to close chats in touch screen devices.

eerielili:
![image](https://github.com/dino/dino/assets/81772782/b3bf1f56-014c-415f-b6b6-0c50fa81cf9a)

Like this would be great?

fiaxh:
Conversations can be closed from the conversation menu now.

PR #1478, opened 2023-09-07
Add a way to close conversation from conversation dialog(#1321)

pksadiq:
It's currently not possible to close a conversation on mobile phones as the close button on conversation row is shown on hover which is not possible on touch only devices.

Also, hovering to close may occasionally result in the wrong item being closed in case a new conversation is open somewhere before the to be closed item.

So add a button to conversation dialog so that users can close the conversation on mobile phones

### DIFF
--- a/main/data/contact_details_dialog.ui
+++ b/main/data/contact_details_dialog.ui
@@ -100,6 +100,18 @@
                                         <property name="margin-start">100</property>
                                     </object>
                                 </child>
+                                <child>
+                                  <object class="GtkButton" id="close_conversation">
+                                    <property name="margin-end">100</property>
+                                    <property name="margin-start">100</property>
+                                    <property name="margin-top">18</property>
+                                    <property name="margin-bottom">18</property>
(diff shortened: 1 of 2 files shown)


pravi:
@pksadiq thanks a lot for this pr. I tested it and it works. Though it'd be better if the closing takes you to the conversations list rather than to the next conversation.

pravi:
I have been using this for more than a week to clear my long spam conversations list and I should say this is a huge improvement to dino. Also looking back, this behaviour of going to next chat is actually better when lots of spam conversations are there.

fiaxh:
Thanks for your PR. Adding a way to close a conversation on touch devices is a good idea.

However, I'd like to put it into a different place: Make the tree-dot-button into an actual menu, and have two entries (for now): Contact/Conversation Details and Close Conversation.

fiaxh:
I implemented this as a conversation menu entry in 2f1b806dfe43a6323d8331a70172f01da88034d7.

image

PR #1480, opened 2023-09-14
Add WeakTimeout Pattern to Prevent Leaks

psycotica0:
While doing testing I noticed that skeletons were being leaked, and eventually tracked it down to the timer that updates the time label closing over "this" and then keeping the reference alive, potentially for 24 hours.

I noticed a few other places in the code doing some version of this, and one of them had the "static and weak pointer" approach, which I pulled out into a util class. Now, we still have to make sure we're passing it a static method instead of a lambda, as that would also close over "this" and render the whole thing useless, but at least most of the annoying parts live in the util class now.

Also the call_widget version was doing a weird thing where it was removing itself, but then returning "true"? I'm not sure what that accomplishes, because returning "false" means to not run this again. So I think my new version is the same in practice, but simpler...

There are other timeouts in the code that I briefly looked over, but all of them seemed to be relatively short hard-coded durations, so I left them alone.

But if any of them are long-lived, it's possible they could also benefit from this class in the future.

### DIFF
--- a/libdino/CMakeLists.txt
+++ b/libdino/CMakeLists.txt
@@ -63,6 +63,7 @@ SOURCES
     src/util/display_name.vala
     src/util/util.vala
     src/util/weak_map.vala
+    src/util/weak_timeout.vala
 CUSTOM_VAPIS
     "${CMAKE_BINARY_DIR}/exports/xmpp-vala.vapi"
     "${CMAKE_BINARY_DIR}/exports/qlite.vapi"
--- a/libdino/src/util/weak_timeout.vala
+++ b/libdino/src/util/weak_timeout.vala
@@ -1,0 +1,40 @@
+public class Dino.WeakTimeout {
+    // XXX: If you get an error saying your function doesn't match the delegate, make sure it's static!
+    //      These are marked as "has_taget=false" so you can't close over "this" and leak it in your lambda.
+    [CCode (has_target = false)]
+    public delegate bool SourceFunc<T> (T object);
+
+    [CCode (has_target = false)]
+    public delegate void SourceOnceFunc<T> (T object);
+
+    public static uint add<T>(uint interval, T object, owned SourceFunc<T> function, int priority = GLib.Priority.DEFAULT) {
+        var weak = WeakRef((Object)object);
+        return GLib.Timeout.add(interval, () => {
+            var strong = weak.get();
+            if (strong == null) return false;
+
+            return function(strong);
+        }, priority);
+    }
+
+    public static uint add_once<T>(uint interval, T object, owned SourceOnceFunc<T> function, int priority = GLib.Priority.DEFAULT) {
+        var weak = WeakRef((Object)object);
+        return GLib.Timeout.add(interval, () => {
+            var strong = weak.get();
+            if (strong == null) return false;
+
+            function(strong);
+            return false;
+        }, priority);
+    }
+
+    public static uint add_seconds<T>(uint interval, T object, owned SourceFunc<T> function, int priority = GLib.Priority.DEFAULT) {
+        return add(interval * 1000, object, (owned) function, priority);
+    }
+
+    // This one doesn't have an upstream equivalent, but it seems pretty obvious to me
+    public static uint add_seconds_once<T>(uint interval, T object, owned SourceOnceFunc<T> function, int priority = GLib.Priority.DEFAULT) {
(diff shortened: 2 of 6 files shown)


mar-v-in:
Thanks. I like the idea of a helper for this!

If you add
```vala
[CCode (has_target = false)]
```
above each delegate declaration, the compiler will error (or at least warn) when trying to pass a closure (lambdas that don't capture anything from the environment, especially not anything referencing this, should be fine).

Also I think we should be able to use type parameters on the delegate and add* functions to make the code a little nicer (no casting needed) and at the same time enforce types at compile time.

psycotica0:
I am not a Vala expert, and I tried to get it to use type parameters on a static method and could never figure it out. So I thought maybe it was just not a feature the language had... I would love to be able to avoid the casting! 

I'll experiment with the has_target declaration as well. Anything that makes mistakes less common sounds good to me!

psycotica0:
Alright, I tried the `has_target = false`, and it means I've got to remove my own lambdas from the `_once` variations in the helper, which is a smidge annoying.

But besides that, when I remove a `static` on purpose it does generate an error! On the downside, the error messages it generates are not particularly great:

```
/home/psycotica0/src/dino/main/src/ui/conversation_content_view/conversation_item_skeleton.vala:184.110-184.127: error: Argument 3: Cannot convert from `void Dino.Ui.ConversationSummary.ConversationItemSkeleton.update_time_lambda (GLib.Object)' to `delegate void Dino.WeakTimeout.SourceOnceFunc (GLib.Object)'
  184 |         time_update_timeout = Dino.WeakTimeout.add_seconds_once((int) get_next_time_change(item.time), this, update_time_lambda);
      |                                                                                                              ^~~~~~~~~~~~~~~~~~

```

It's not _wrong_, it's telling me this function doesn't match the expected type, but there's nothing in the error to highlights for me what the problem actually is. Because the problem isn't in the vala type at all, I guess.

So I guess better than a memory leak, but I may tweak the comment I've got there telling you to give me static stuff to instead say "Hey, are you getting this error? Yeah, it's because of this" 😅

mar-v-in:
```vala
public class Dino.WeakTimeout {
    [CCode (has_target = false)]
    public delegate bool SourceFunc <T> (T object);
    [CCode (has_target = false)]
    public delegate void SourceOnceFunc <T> (T object);

    public static uint add<T>(uint interval, T object, owned SourceFunc<T> function, int priority = GLib.Priority.DEFAULT) {
        var weak = WeakRef((Object)object);
        return GLib.Timeout.add(interval, () => {
            var strong = weak.get();
            if (strong == null) return false;

            return function(strong);
        }, priority);
    }

    public static uint add_once<T>(uint interval, T object, owned SourceOnceFunc<T> function, int priority = GLib.Priority.DEFAULT) {
        var weak = WeakRef((Object)object);
        return GLib.Timeout.add(interval, () => {
            var strong = weak.get();
            if (strong == null) return false;

            function(strong);
            return false;
        }, priority);
    }

    public static uint add_seconds<T>(uint interval, T object, owned SourceFunc<T> function, int priority = GLib.Priority.DEFAULT) {
        return add<T>(interval * 1000, object, (owned) function, priority);
    }

    public static uint add_seconds_once<T>(uint interval, T object, owned SourceOnceFunc<T> function, int priority = GLib.Priority.DEFAULT) {
        return add_once<T>(interval * 1000, object, (owned) function, priority);
    }
}
```
Should be doing the trick. Then when calling, you'll have to specify the type parameter explicitly (this looks like a bug in Vala compiler, because it should be able to infer, but anyways).
```vala
Dino.WeakTimeout.add_seconds_once<Foo>(1, foo, (foo) => { print(foo.bar); });
```

psycotica0:
Ok, I've pushed an update! It turns out the inference does seem to work when using static methods, so the callsites didn't have to change at all, and the static methods just got simpler by taking the casts out. Great!

One thing I noticed in another branch, though, is that it seems like the `connect` methods on notifiers... it seems like over there if I use a lambda things get leaked, but if I instead make an instance method and just reference the method name it doesn't appear to.

Like, doing:

```vala
thing.connect(() => do_something());
```

appears to behave differently than:

```vala
thing.connect(do_something);
```

At least in some testing I did. Maybe I was wrong and it was closing over a local stack variable that was the problem in that case, I don't know. Anyway, that's not related to this PR, but it seemed like a difference that may be nice if it was expected and I didn't have to pass self through at all here.

PR #1481, opened 2023-09-17
Simplify talking to yourself (notes). Ctrl+M shortcut, M for Me

eerielili:


eerielili:
Re,

I did some changes: I made the notes more recognizable so if you have more than one account you shouldn't be too confused:
image
 
As you can see it's now underlined, bold and your JID is mentioned.
I also removed the shortcut (Ctrl+M) from the menu action as the feature will be used not as frequently as joining groupchats or starting discussions. 

Kind regards,
lili

eerielili:
The pushes were for a change in how more than one account is handled.
It will now show something similar to this in case of more than one account enabled:
image

eerielili:
not a notes app

Issue #556, opened 2019-04-27
Chat: Delete inband image locally


Echolon:
Hi,

inspired from Conversations, it would be great to delete the shown files (images) from the local inband chat window (and machine). This is interesting for privacy reasons sometime in case other people are likely to see your screen.

Cheers & thanks for Dino

IeP4nieF:
As a workaround you can delete the images manually from the files directory (normaly ~/.local/share/dino/files).

Echolon:
I ran into this problem today. It would be great to have that option for sensitive information to not appear in the chat anymore.

PR #1483, opened 2023-09-19
Allow Hiding Images

psycotica0:
When we get an image we can display, rather than just putting the image in we now nest it inside an expander, which allows it to be hidden.
    
Now if someone has sent you an image you don't want to see anymore, you can make that choice for yourself. Maybe the image was spam, maybe someone was messing with you, or maybe it made perfect sense in context but you're just tired of seeing it in the backscroll. In any case, you hide it and it stays in the history and can be expanded at any time in the future, but it's hidden until then.
    
This state is preserved across conversation changes and restarts, so leaving a conversation and coming back will still not show the image.
    
To preserve this the state is changed to a "collapsed" state. It's a little weird to use the file transfer state for this, but it doesn't really make sense to have a collapsed image in any other state than completed, so I didn't want it to be a new boolean column. Actually, it's weird to put this one the transfer at all, but it seems like this table is considered the source of this UI, and there isn't really another place to information about the image other than the transfer object.
    
We can change this later if we decide it was a bad idea.

Here's what it looks like:
image

### DIFF
--- a/libdino/src/entity/file_transfer.vala
+++ b/libdino/src/entity/file_transfer.vala
@@ -11,7 +11,8 @@ public class FileTransfer : Object {
         COMPLETE,
         IN_PROGRESS,
         NOT_STARTED,
-        FAILED
+        FAILED,
+        COLLAPSED
     }
 
     public int id { get; set; default=-1; }
--- a/libdino/src/service/content_item_store.vala
+++ b/libdino/src/service/content_item_store.vala
@@ -402,6 +402,8 @@ public class FileItem : ContentItem {
                 return Entities.Message.Marked.UNSENT;
             case FileTransfer.State.COMPLETE:
                 return Entities.Message.Marked.NONE;
+            case FileTransfer.State.COLLAPSED:
+                return Entities.Message.Marked.NONE;
             case FileTransfer.State.NOT_STARTED:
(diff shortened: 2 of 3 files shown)


Echolon:
ref: https://github.com/dino/dino/issues/556

maikelthedev:
I like your train of thought but think the result could be improved to have better UX. 

Instead of putting a meaningless name and a caret, put something along these lines instead, same way in Mastodon sensitive images are hidden AND/OR users can decide to have all images hidden as default. 

image

Why do I think this is a better solution in terms of UX?
1. No meaningless file name. 
2. Familiar interface
3. Clearer interaction intructions
4. More pleasant to the eyes.

psycotica0:
Respectfully disagree, or at least your suggestion has different goals than mine.
I don't just want to obscure the content, I want to not see it, so I definitely want it to collapse.

Otherwise if there's a few images in the chat, and I hide them all, I don't regain the space, it's just that now my whole chat window is taken up with 4 blurs that all imply the image I know is behind them. Not helpful to me.

But, there could be a nicer looking version of mine, for sure. I just don't think the blurring version does what I wanted.

And also there'd still need to be some UX you didn't cover to take a currently showing image and make it hidden after I've seen it.

PR #1485, opened 2023-09-28
Start conversation if closed when receiving an audio or video call

### LINKED ISSUE #1268: GTK4 - user chat should open when they call
Dino b8e84c8 amd64 Debian Sid, calling from Conversations 2.10.9+free

Contact chat is closed, only a notification appears (at least in my dunst I can't action on it).

If I open the chat with that contact the buttons are there and I can answer.

eerielili:
Closes #1268.
Don't jump to the conversation to not disrupt you if you're typing.

### DIFF
--- a/libdino/src/application.vala
+++ b/libdino/src/application.vala
@@ -39,12 +39,12 @@ public interface Application : GLib.Application {
         PresenceManager.start(stream_interactor);
         CounterpartInteractionManager.start(stream_interactor);
         BlockingManager.start(stream_interactor);
+        Calls.start(stream_interactor, db);
         ConversationManager.start(stream_interactor, db);
         MucManager.start(stream_interactor);
         AvatarManager.start(stream_interactor, db);
         RosterManager.start(stream_interactor, db);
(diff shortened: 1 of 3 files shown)


fiaxh:
Thanks for your PR.

You put the code for opening a conversation into the two NotificationProviders. Their purpose is only to create a notification in the desktop environment, thus opening a conversation is outside of their responsibility.

The code for opening a conversation when a new message arrives is in the conversation_manager.vala. You can put the code for opening conversations on calls (and file transfers) there as well. The CallManager and FIleTransferManager have signals for new items which you can listen on.

eerielili:
@fiaxh Thanks you for the pointers, I updated my branch.

fiaxh:
With the new changes the code didn't actually work anymore, because the CallManager didn't exist yet at the point when the ConversationManager tried to connect to it. I "fixed" that, but long-term we'll want a better fix for that.

Also, I added setting `conversation.last_active` to the new function in the ConversationManager and removed the respective code in the CallManager.

PR #1486, opened 2023-09-28
More Image Leaks

psycotica0:
While building animated gifs I had some tracing on to make sure I wasn't introducing any leaks, and found some more that were pre-existing, but unrelated to animated GIFs.

The first was a cycle where hooking up the overlay in the file widget introduced a cycle that kept the file widget around even when the parent view was destroyed. I believe this one actually _kinda_ works as-is, because the parent view in practice calls `dispose()` manually on the children, but that's a bit fragile and still isn't quite the same. But then when I implemented #1483 the parent became a standard Expander widget with no special handling, and then it was really broken. So this makes it so any standard widget can be the parent.

The second was the "FileSendOverlay" which was created as a stack variable and then not referenced anywhere else. So that should have crashed immediately, but the leaking from the two callbacks in scope kept it around forever instead. So what I did here was actually reference it from the conversation view controller and then explicitly clear it after the overlay closes (with a slight delay, since that's in a handler on a signal on that object, and it crashes if I free it from that context). I can be a bit looser here because this `this` in this context is the `ConversationViewController`, which appears to functionally be a singleton.

### DIFF
--- a/main/src/ui/conversation_content_view/file_image_widget.vala
+++ b/main/src/ui/conversation_content_view/file_image_widget.vala
@@ -8,6 +8,10 @@ namespace Dino.Ui {
 
 public class FileImageWidget : Box {
 
+    Gtk.Box image_overlay_toolbar;
+    MenuButton button;
+    GestureClick gesture_click_controller;
+
     public FileImageWidget() {
         this.halign = Align.START;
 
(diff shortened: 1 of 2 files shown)


christf:
I am still seeing the symptom in master. It would be great if this could be rebased and ultimately merged.
The "unlink Overlay After it's closed" commit makes it a bit better but it is not fully gone.

fiaxh:
Thanks for your PR and sorry for the very delayed reaction.

[Fix Cycle Holding FileImageWidget Open](https://github.com/dino/dino/pull/1486/commits/da8b5418043e509275b83bd8e7b16ba6775bdefa) has been done while making larger changes on file displaying.

[Unlink Overlay After It's Closed](https://github.com/dino/dino/pull/1486/commits/6d19e8d00dd6b64d6d0256c184724c7f7424c1d9) does not make a difference regarding memory leaks in my tests - maybe that was different back when you wrote it. However, memory leaks in the file send dialog have been fixed in 68b07ef13c857b07d32c2fd8da08a0660f961a0a. Also, the file send dialog is now a normal Dialog as of d479ba81a66a835913697bacf3a0d912ef68346a.

christf:
Thank you for looking at it closely.

PR #1489, opened 2023-10-04
CMake, configure: Force `libsoup` version

XaviDCR92:
Whereas Dino did check which `libsoup` version was required by `libnice`, it did nothing to stop the user from linking against an invalid version, leading to the following error message on startup:

```
(dino:191348): libsoup-ERROR **: 22:55:21.446: libsoup2 symbols detected. Using libsoup2 and libsoup3 in the same process is not supported.
Trace/breakpoint trap (core dumped)
```

Stricter checks have been added to avoid this situation. On the other hand, `USE_SOUP3` was removed in favour of `SOUP_VERSION`, which is the only CMake variable that was being consistently used for this purpose.

### DIFF
--- a/cmake/SoupVersion.cmake
+++ b/cmake/SoupVersion.cmake
@@ -34,7 +34,7 @@ elseif (NOT SOUP_VERSION)
     find_package(Soup2 QUIET)
     find_package(Soup3 QUIET)
     # Only use libsoup 3 if specifically requested or when libsoup 2 is not available
-    if (Soup3_FOUND AND NOT Soup2_FOUND OR USE_SOUP3)
+    if (Soup3_FOUND AND NOT Soup2_FOUND)
         set(SOUP_VERSION 3)
     else ()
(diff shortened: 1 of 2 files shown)


mar-v-in:
The current behavior is somewhat intended. The user choice to use a specific libsoup version should disable and override any automatic detection. There are two reasons:
- Automatic detection does not work reliably on all platforms (e.g. I think it's still not working with musl) and we need a way to entirely turn off the routine (as it can cause fatal errors). Setting SOUP_VERSION manually is what currently allows to turn off the detection.
- Build system and runtime system might be different. When you see the above error message, Dino is only linking against one version of libsoup. libgupnp-igd (a dependency of libnice) can be linked against either libsoup 2 or 3 and we are not supposed to even know, which version of libgupnp-igd is used by libnice at runtime. The automatic detection really is only a hack (which is why it doesn't work on all platforms).

Issue #1350, opened 2023-01-25
Proposing better notification of file and image transfers


gnusupport:
When I drag and drop image to Dino window, I can see image in the chat, but I can't know if that image was really transferred to other person.

On the side list of contacts I could see "image sent". I have asked my contact however, if person has see the image, and person did not see the image.

That means "image sent" was incorrect notification.

There shall be upload progress, and visual notification if image was sent.

speedygonzales9:
Duplicate of issue #1253 ?

XaviDCR92:
@gnusupport @speedygonzales9 should be fixed with https://github.com/dino/dino/pull/1491.

In case you are interested, I am maintaining a new fork of Dino that already includes this fix: https://gitea.privatedns.org/xavi/dino.

PR #1491, opened 2023-10-06
Show file upload/download progress

XaviDCR92:
Fixes upstream issue #1350.

https://github.com/dino/dino/assets/25252461/2ff0dfd8-ad39-4d3c-9dd4-598ff8c4c746

## Notes

Image uploads were incorrectly handled by Dino, as they were always reported as completed even if they were not, maybe so as to show the image preview from the start. Now, Dino shows the upload progress for all file types, and the image is only shown when completed.

### DIFF
--- a/libdino/src/entity/file_transfer.vala
+++ b/libdino/src/entity/file_transfer.vala
@@ -71,6 +71,7 @@ public class FileTransfer : Object {
     public int provider { get; set; }
     public string info { get; set; }
     public Cancellable cancellable { get; default=new Cancellable(); }
+    public uint64 transferred_bytes { get; set; }
 
     private Database? db;
     private string storage_dir;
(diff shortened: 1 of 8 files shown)


fiaxh (changes requested):
Thanks for your PR!

I added some review comments.

XaviDCR92:
@fiaxh all issues have been already addressed. Could you please review again?

fiaxh:
In order to keep the discussions around this PR in one place:

There was an exchange on this topic in the Dino Channel a while back. mar-v-in suggested that it would be better to 

>do [the counting of transferred bytes] in file manager such that it applies to both [file transfer types] just like you did for receiving files

fiaxh:
Done in d78ec05.

PR #1493, opened 2023-10-09
Allow blocking entire domain from conversation details

### LINKED ISSUE #1429: Allow blocking entire domain
In cheogram android app (possibly on conversations too), when you receive a message from stranger (people not added as a contact), in addition to block user, there is a block domain option too. We should have such an option on Dino too to tackle huge amounts of spam.

eerielili:
Might close #1429.
I tested it on my end, works well. Might look around for less lines of code, still a WIP.
image

When domain is blocked:
image

Don't hesitate to give me feedback.

### DIFF
--- a/main/data/conversation_details.ui
+++ b/main/data/conversation_details.ui
@@ -122,11 +122,12 @@
                           </object>
                         </child>
                         <child>
-                          <object class="GtkButton" id="block_button">
+                          <object class="AdwSplitButton" id="block_button">
+                            <property name="menu_model">block_menu_model</property>
                             <property name="visible" bind-source="model" bind-property="show-blocked" bind-flags="sync-create"/>
                             <child>
                               <object class="AdwButtonContent" id="block_button_content">
-                                <property name="icon-name">action-unavailable-symbolic</property>
+                                <property name="icon-name">dino-block-symbolic</property>
                                 <property name="label">Block</property>
                               </object>
                             </child>
@@ -200,6 +201,22 @@
       <item>
         <attribute name="label">Reset to default</attribute>
         <attribute name="action">notification.default</attribute>
+      </item>
+    </section>
+  </menu>
+  <menu id="block_menu_model">
+    <section>
+      <item>
+        <attribute name="label" translatable="1">Block user</attribute>
+        <attribute name="action">do.block</attribute>
+      </item>
+      <item>
+        <attribute name="label" translatable="1">Block entire domain</attribute>
+        <attribute name="action">do.block_domain</attribute>
+      </item>
+      <item>
+        <attribute name="label" translatable="1">Unblock</attribute>
+        <attribute name="action">do.unblock</attribute>
       </item>
     </section>
   </menu>
--- a/main/data/conversation_details.ui
+++ b/main/data/conversation_details.ui
@@ -122,8 +122,7 @@
                           </object>
                         </child>
                         <child>
-                          <object class="AdwSplitButton" id="block_button">
-                            <property name="menu_model">block_menu_model</property>
+                          <object class="GtkMenuButton" id="block_button">
                             <property name="visible" bind-source="model" bind-property="show-blocked" bind-flags="sync-create"/>
                             <child>
                               <object class="AdwButtonContent" id="block_button_content">
@@ -201,22 +200,6 @@
       <item>
         <attribute name="label">Reset to default</attribute>
         <attribute name="action">notification.default</attribute>
-      </item>
-    </section>
-  </menu>
-  <menu id="block_menu_model">
-    <section>
-      <item>
-        <attribute name="label" translatable="1">Block user</attribute>
-        <attribute name="action">do.block</attribute>
-      </item>
-      <item>
-        <attribute name="label" translatable="1">Block entire domain</attribute>
-        <attribute name="action">do.block_domain</attribute>
-      </item>
-      <item>
-        <attribute name="label" translatable="1">Unblock</attribute>
-        <attribute name="action">do.unblock</attribute>
       </item>
     </section>
   </menu>
--- a/main/src/view_model/conversation_details.vala
+++ b/main/src/view_model/conversation_details.vala
@@ -6,15 +6,14 @@ using Gtk;
 
 public class Dino.Ui.ViewModel.ConversationDetails : Object {
     public signal void pin_changed();
-    public signal void block_changed(BlockActions action);
+    public signal void block_changed(BlockState action);
     public signal void notification_flipped();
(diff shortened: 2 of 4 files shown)


pravi:
I have tested it on Purism Librem 5 (debs here https://people.debian.org/~praveen/mobian/), it is not adaptive, please use libadwaita widgets to make it adaptive. arrange the buttons vertically.

Screenshot from 2023-10-25 17-40-45

pravi:
When I have multiple accounts, it does not show which account I'm using.

eerielili:
> When I have multiple accounts, it does not show which account I'm using.

I partly addressed it in [#1482](https://github.com/dino/dino/pull/1482), I'll do a separate one with a line of text underneath the contact's avatar indicating which account is used.

> I have tested it on Purism Librem 5 (debs here https://people.debian.org/~praveen/mobian/), it is not adaptive, please use libadwaita widgets to make it adaptive. arrange the buttons vertically.
> 
> Screenshot from 2023-10-25 17-40-45

I have force pushed a fix to my branch so that the button goes on a new line. Could have calculated the root window width, stored it, and switch the box orientation to vertical when needed but it's simpler like this.

pravi:
I have added  both patches, the account used I can now see, but the window is still same (cutting on both sides).

eerielili:
> I have added both patches, the account used I can now see, but the window is still same (cutting on both sides).

Nice. I think I know why it's cutting, it must be the margin-end value that I doubled when trying to put the button on another line. I put it back to normal, I think it should be fixed for you.

eerielili:
@pravi i did not tag you, could you test https://github.com/dino/dino/pull/1493/commits/8cb26d599a7e9f43fdf719c6063990ce3d3ced9f ? thx

eerielili:
@pravi I have a much better (imho) interface now:

The block button has a dropdown where you can block the domain:
image

main/data/conversation_details.ui
```diff
@@ -122,7 +122,8 @@
                           </object>
                         </child>
                         <child>
-                          <object class="GtkButton" id="block_button">
+                          <object class="AdwSplitButton" id="block_button">
+                            <property name="menu_model">block_menu_model</property>
                             <property name="visible" bind-source="model" bind-property="show-blocked" bind-flags="sync-create"/>
                             <child>
                               <object class="AdwButtonContent" id="block_button_content">
```
eerielili, line 129: @fiaxh , like in 384ef1d3f15c82af1a4647b0ee704f77919ff2b1, should `action-unavailable-symbolic` also be replaced by `dino-block-symbolic`  here? thanks

libdino/src/service/blocking_manager.vala
```diff
@@ -20,19 +20,22 @@ public class BlockingManager : StreamInteractionModule, Object {
         this.stream_interactor = stream_interactor;
     }
 
-    public bool is_blocked(Account account, Jid jid) {
+    public bool is_blocked(Account account, Jid jid, bool domainblock = false) {
```
fiaxh, line 23: Instead of adding a `domainblock` parameter here, you could just pass the domain JID, when a domain block is desired. The `Jid` object has a property `domain_jid`.

main/src/view_model/conversation_details.vala
```diff
@@ -32,6 +33,7 @@ public class Dino.Ui.ViewModel.ConversationDetails : Object {
 
     public bool show_blocked { get; set; }
     public bool blocked { get; set; }
+    public bool domain_blocked { get; set; }
```
fiaxh, line 36: Similarly to the NotificationSetting, you could create an enum that represents the blocking states: None, Blocked, DomainBlocked.

fiaxh:
Thanks for the PR, I like your new design!

Currently, you have two buttons: One "Block" button with the option to "Block domain", and one "Domain blocked" button. I'd suggest to merge the two and always show the same menu with three options: "Block user", "Block entire domain", "Unblock". The content would change between "Block", "Blocked" and "Domain blocked", while the default action would be a normal blocking.

libdino/src/service/blocking_manager.vala
eerielili: Oh I didn't remember this, thanks for pointing it out.

main/src/view_model/conversation_details.vala
eerielili: Will do!

eerielili:
@fiaxh Thanks you very much for the feedback once again.
Here's what the most recent commit looks like:
image
image
image

I'll force push some more to add some `translatable="1"` to some labels.

main/src/view_model/conversation_details.vala
fiaxh: My thought was to also use this enum for representing the state in the ViewModel and Model. So we'd have a `public BlockState block_state { get; set; }` that replaces `blocked` and `domain_blocked`.

main/src/windows/conversation_details.vala
```diff
@@ -27,16 +27,20 @@ namespace Dino.Ui.ConversationDetails {
...
             install_action("notification.default", null, (widget, action_name) => { ((Dialog) widget).model.notification_changed(ViewModel.ConversationDetails.NotificationSetting.DEFAULT); } );
+            install_action("do.block", null, (widget, action_name) => { ((Dialog) widget).model.block_changed(ViewModel.ConversationDetails.BlockActions.USER); } );
+            install_action("do.block_domain", null, (widget, action_name) => { ((Dialog) widget).model.block_changed(ViewModel.ConversationDetails.BlockActions.DOMAIN); } );
+            install_action("do.unblock", null, (widget, action_name) => { ((Dialog) widget).model.block_changed(ViewModel.ConversationDetails.BlockActions.UNBLOCK); } );
         }
 
         construct {
             pin_button.clicked.connect(() => { model.pin_changed(); });
-            block_button.clicked.connect(() => { model.block_changed(); });
+            block_button.clicked.connect(() => { model.block_changed(ViewModel.ConversationDetails.BlockActions.TOGGLE); });
```
fiaxh, line 37: I'd argue that it's the job of the UI to evaluate the result of the toggling. So I'd remove the TOGGLE state from the enum and return USER/UNBLOCK depending on the previous state.

main/data/conversation_details.ui
fiaxh: Yes, please!

main/src/windows/conversation_details.vala
eerielili: I could make it the default action of the signal and revert back to `model.block_changed();` i.e:
`public signal void block_changed(BlockActions action = TOGGLE);`

PR #1495, opened 2023-10-12
[WIP] Add keyboard shortcut to zoom in or out text (fixes #978)

### LINKED ISSUE #978: [Feature Request] Possibility to Zoom
It would be really wonderful if you could zoom in Dino. That is, That is, you should be able to make the font of the chats larger. In many situations, the font is too small for me and I have a hard time deciphering what is written to me. With other programs this is not a problem, because there it is easy to simply make the text larger. This should be possible via keyboard shortcut without delay.

eerielili:


eerielili:
The setting will persist and text size will change dynamically in the conversation and conversation selection rows as well from here
image

cal0pteryx:
Thanks for your PR! I took some inspiration and added this feature to Gajim as well :+1:

eerielili:
> Thanks for your PR! I took some inspiration and added this feature to Gajim as well 👍

No problem, that's really nice!

PR #1510, opened 2023-11-12
XEP-0215: management of the field `expires`

mid1221213:
Add a field in `Xmpp.Xep.ExternalServiceDiscovery` to keep track of the `expires` TURN service value and use it (divided by 2) to restart periodically the external services discovery.

Without this patch, calls (actually ICE) fail to initiate after 24h of Dino uptime because of the expiry of the TURN credentials. 

This happened on my setup using prosody (using the default credentials' TTL of 86400) and coturn.

### DIFF
--- a/libdino/src/service/registration.vala
+++ b/libdino/src/service/registration.vala
@@ -71,6 +71,12 @@ public class Register : StreamInteractionModule, Object{
         return ret;
     }
 
+    public async string? change_password(Account account, string new_pw){
+        XmppStream stream = stream_interactor.get_stream(account);
+        if (stream == null) return null;
+        return (yield stream.get_module(Xep.InBandRegistration.Module.IDENTITY).change_password(stream, account.full_jid, new_pw)).condition;
+    }
+
     public class ServerAvailabilityReturn {
         public bool available { get; set; }
         public TlsCertificateFlags? error_flags { get; set; }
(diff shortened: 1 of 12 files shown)


fiaxh:
Thanks for looking into this.

There are some issues:
- A new Timeout is added whenever a stream is negotiated and they are not cancelled when the stream is broken. That means, there can exist a number of superflurous Timeouts. This is also problematic because the closure captures the stream, preventing it from getting freed.
- Within the closure, you check whether the account has is connected, but it could be connected via another stream. If the account is connected with a different stream, you would still call the on_stream_negotiated function, but with a broken stream.
- Please use a DateTime for storing the expired value instead of an int64? - we try to avoid using question-primitives, as accessing them wrongly leads to crashes.
- Also with those changes, we should rename `on_stream_negotiated`, since it's not only called on stream negotiation anymore :slightly_smiling_face:

mid1221213:
Thanks @fiaxh for your comments. Here is a new version, hopefully better :wink:

fiaxh:
I merged this in 63439cd206aaf3548228e915bd1744389cd2fc54. Also, I simplified the timeout-removal logic a bit in 3a4ac7c11d93553a485aa7dec9be05d71a7c9cb6.

(Also, I actually intended to click on merge here, but accidentally closed the PR by pushing a wrong state to it)

Issue #1222, opened 2022-04-10
Http upload file send is defective when server (jabberd) is configured with max_size: infinity


emildekeyser:
Sending files either doesnt work (after selecting file, you get an error) or uses jingle instead of http upload when both sides are online. This occurs with ejabberd when the max_size parameter from mod_http_upload is set to 'infinity'. Setting it to a arbitray number fixes the issue.

Yuubi-san:
The immediate bug seems to be in `Dino.Plugins.HttpFiles.HttpFileSender.can_send()` at https://github.com/dino/dino/blob/47a066987dec6b83c3964feb3571aee2d03d7f73/plugins/http-files/src/file_sender.vala#L55 which doesn't check `max_file_sizes[conversation.account]` for negativity. Max file size of -1, in turn, comes from `Xmpp.Xep.HttpFileUpload.Module.extract_max_file_size` at https://github.com/dino/dino/blob/47a066987dec6b83c3964feb3571aee2d03d7f73/xmpp-vala/src/module/xep/0363_http_file_upload.vala#L184 when the size limit attribute is absent in the XML (which is not required according to the XEP.

The Right Way (TM) to fix this, IMO, is to not signal non-numeric information in-band with numeric information, e.g. return `long?` instead of `long` from `extract_max_file_size`.

mar-v-in:
Nullable value types are still rather experimental in Vala (there is a compiler flag `--enable-experimental-non-null` to add proper nullable checking, but that's still considered experimental). If `extract_max_file_size()` returned a `long?` of value `null`, the comparison for the transfer size would be "valid", but segfault if no further checks are added.

We generally prefer "misbehavior" over segfaults, that's why we try to not use nullable value types (like `long?`, `int?`, `bool?`; nullable reference types do not segfault). That's why I'd prefer to just make `extract_max_file_size` return `long.MAX` for unlimited file size.

PR #1512, opened 2023-11-14
Fix for jabberd XMPP server 'infinity' http upload file size announce

eerielili:
should fix https://github.com/dino/dino/issues/1222

### DIFF
--- a/xmpp-vala/src/module/xep/0363_http_file_upload.vala
+++ b/xmpp-vala/src/module/xep/0363_http_file_upload.vala
@@ -181,7 +181,9 @@ public class Module : XmppStreamModule {
             }
         }
         if (max_file_size_str != null) return long.parse(max_file_size_str);
-        return long.MAX; //for ejabberd max_size:infinity
+
+        // If there is no max-file-size node, there is no file size limit
+        return long.MAX;
     }
 }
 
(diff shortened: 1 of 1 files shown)


fiaxh:
Did you test this with an ejabberd server that has the maximum size set to infinity? Or do you have an example stanza? I quickly searched through the source code of Conversations and Gajim and didn't find the string "infinity" in either of them, hence I'm wondering if this is really what is sent on the wire.

eerielili:
@fiaxh No, I naively assumed it sent back the string "infinity" when reading the issue.
I did test things out after your comment by setting up a simple ejabberd instance and putting some `stderr.printf` in the `extract_max_file_size` function like so:
```vala
                if (var_attr == "max-file-size") {    
                    StanzaNode value_node = node.get_subnode("value", "jabber:x:data");    
                    max_file_size_str = value_node.get_string_content();    
                    stderr.printf("\nReported file size limit is %s bytes.\n\n", max_file_size_str);    
                    break;    
                }    
```

When setting 'infinity' in the ejabberd.yml file, it didn't print anything, but when setting a numeric value it did print it. I'll test some more.

fiaxh:
Thanks for looking into this. In that case, this is actually not an ejabberd quirk, but the standard-compliant way of announcing that there is no file size limit. And with this change, Dino is now actually handling it :)

eerielili:
> **if such a limitation exists**.

Ooooh nice catch

Issue #1508, opened 2023-11-03
Segmentation fault when showing contact details while server has no support for XEP-0191


horazont:
# How to reproduce

1. Log into a server without support for XEP-0191 (Blocking Command)
2. Add a contact
3. Open a chat with that contact
4. Click on the details menu

# What happens

<pre>
Thread 1 "dino" received signal SIGSEGV, Segmentation fault.
0x00007ffff7dfac48 in xmpp_xep_blocking_command_module_is_blocked (self=self@entry=0x555555fce940, stream=stream@entry=0x555555f2b900, jid=jid@entry=0x55555d123640 "XXX")
    at /home/horazont/Builds/dino/xmpp-vala/src/module/xep/0191_blocking_command.vala:15
15              return stream.get_flag(Flag.IDENTITY).blocklist.contains(jid);
(gdb) bt
#0  0x00007ffff7dfac48 in xmpp_xep_blocking_command_module_is_blocked (self=self@entry=0x555555fce940, stream=stream@entry=0x555555f2b900, jid=jid@entry=0x55555d123640 "XXX")
    at /home/horazont/Builds/dino/xmpp-vala/src/module/xep/0191_blocking_command.vala:15
#1  0x00007ffff7ef4048 in dino_blocking_manager_is_blocked (self=self@entry=0x55555593a410, account=account@entry=0x555555aaf030, jid=Python Exception <class 'ValueError'>: Variable 'static_fundamental_type_nodes' not found.
) at /home/horazont/Builds/dino/libdino/src/service/blocking_manager.vala:25
#2  0x00005555555dccf5 in dino_ui_conversation_details_populate_dialog (stream_interactor=<optimized out>, conversation=0x555556204f00, model=<optimized out>) at /home/horazont/Builds/dino/main/src/ui/conversation_details.vala:12
#3  dino_ui_conversation_details_populate_dialog (model=<optimized out>, conversation=0x555556204f00, stream_interactor=<optimized out>) at /home/horazont/Builds/dino/build/main/conversation_details.c:564
#4  0x00005555555dcfb5 in dino_ui_conversation_details_setup_dialog (conversation=conversation@entry=0x555556204f00, stream_interactor=stream_interactor@entry=0x55555591c560, parent=0x5555564f9e80)
    at /home/horazont/Builds/dino/main/src/ui/conversation_details.vala:112
[...]

PR #1513, opened 2023-11-16
Fix segfault opening conversation details when no XEP-0191 support 

eerielili:
should close https://github.com/dino/dino/issues/1508

It also removes the block button from conv details when it's not supported.


fiaxh:
Thanks for your PR!

In addition to your changes I think it would be nice if the is_blocked function in 0191_blocking_command.vala wouldn't cause Dino to segfault in first place. We could do that by checking whether `stream.get_flag(Flag.IDENTITY)` exists before accessing fields on the flag.

eerielili:
@fiaxh You're right, it makes more sense to put the logic in the file handling the XEP. Working on it.

eerielili:
@fiaxh Done.

Issue #1422, opened 2023-04-07
Code blocks don't work


Zash:
Steps to reproduce:

In any chat, send
````
```
some fenced code block text
```
````

Expected:

The text is rendered in a monospace font.

Actual:

![image](https://user-images.githubusercontent.com/197474/230588036-e045f179-ddfd-4756-a0ca-182b25c1c985.png)

To me it looks as if the three \` characters are treated as a code _span_ containing one \` character.

Tested in versions:
- `0.4.2-1` on Debian 12/testing
- `0.3.0-2-bpo11+2` on Debian 11/stable

**Comment by eerielili** [comment]
> To me it looks as if the three ` characters are treated as a code _span_ containing one ` character.

Correct, it looks like this when printfing it out:
```html
<span color='#9E9E9E'>`</span><tt>`</tt><span color='#9E9E9E'>`</span>
```

eerielili:
Seems like https://github.com/dino/dino/blob/master/main/src/ui/util/helper.vala#L268 is the culprit
For now, changing the two following lines:
```vala
        string[] markup_string = new string[]{"`", "_", "*", "~"};
        string[] convenience_tag = new string[]{"tt", "i", "b", "s"};
```

to

```vala
        string[] markup_string = new string[]{"```","`", "_", "*", "~"};
        string[] convenience_tag = new string[]{"tt", "tt", "i", "b", "s"};
```

should at least make triple backticks work for inline like single backticks.
short on time rn, will probably come back with PR later

PR #1514, opened 2023-11-21
Implement triple backtics as code blocks

eerielili:
closes https://github.com/dino/dino/issues/1422

### DIFF
--- a/main/src/ui/util/helper.vala
+++ b/main/src/ui/util/helper.vala
@@ -270,12 +270,33 @@ public static string parse_add_markup_theme(string s_, string? highlight_word, b
     }
 
     if (parse_text_markup) {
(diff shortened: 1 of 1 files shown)


eerielili:
hello @Zash could you try this PR and give me feedback?
thanks you very much!

fiaxh:
A while back we already had an implementation of code blocks using regexes. However it turned out that the rather complex regex expression caused crashes due to its large recursion stack https://github.com/dino/dino/pull/626#issuecomment-706693765.

I'd be happy to have this feature in Dino, however we should implement a proper markup parser instead of using regexes.

eerielili:
@fiaxh Understood. Seems there's some stuffs at https://valadoc.org/glib-2.0/GLib.MarkupParser.html, will have a look at it.

Issue #1475, opened 2023-08-19
Dino crashes when trying to remove a MUC from the roster that is not actually a proper member of the roster


jscott0:
This SIGSEGV can be reproduced in the following way:

- Start Dino
- Press Ctrl+G
- Type in 'debian@conference.debian.org' or some other channel you are in into the box
- Click on the entry in the list
- Press the minus button at the bottom

```
#0  __lambda6_ (row=<optimized out>, self=0x5555594201f0 [DinoUiAddConferenceDialog]) at ./main/src/ui/add_conversation/add_conference_dialog.vala:104
        _tmp1_ = 0x55555573f4b0 [DinoStreamInteractor]
        _tmp3_ = <optimized out>
        _tmp5_ = <optimized out>
        conference_row = 0x0
        _tmp0_ = 0x0
        _tmp2_ = <optimized out>
        _tmp4_ = <optimized out>
        _tmp6_ = <optimized out>
        __func__ = "__lambda6_"
#1  ___lambda6__dino_ui_select_jid_fragment_remove_jid (_sender=<optimized out>, row=<optimized out>, self=0x5555594201f0) at ./main/src/ui/add_conversation/add_conference_dialog.vala:102
#6  0x00007ffff76d85ff in <emit signal 'remove-jid' on instance 0x5555594536f0 [DinoUiSelectJidFragment]> (instance=instance@entry=0x5555594536f0, signal_id=<optimized out>, detail=detail@entry=0) at ../../../gobject/gsignal.c:3675
        var_args = {{gp_offset = 32, fp_offset = 48, overflow_arg_area = 0x7fffffffc900, reg_save_area = 0x7fffffffc840}}
    #2  0x00007ffff76be3f8 in g_closure_invoke (closure=0x55555970df20, return_value=0x0, n_param_values=2, param_values=0x7fffffffc5f0, invocation_hint=0x7fffffffc540) at ../../../gobject/gclosure.c:832
                marshal = 0x7ffff76c15e0 <g_cclosure_marshal_VOID__OBJECT>
                marshal_data = 0x0
                in_marshal = 0
                real_closure = 0x55555970df00
                __func__ = "g_closure_invoke"
[...]

PR #1516, opened 2023-11-25
Fix crash on removing conference not in roster

eerielili:
should close https://github.com/dino/dino/issues/1475


fiaxh:
The issue is that `conference_row` in the line
```vala
ConferenceListRow conference_row = row as ConferenceListRow;
```
is null, because `row` is not a `ConferenceListRow`, but an `AddListRow` (which is used for entries that are dynamically added based on input in the search field).

Thus, you can solve this crash much simpler by checking `if (conference_row == null) return;` right after that line :slightly_smiling_face:

eerielili:
@fiaxh Oh, how did I not think of that I should've read slower.
Anyways compiled and tested and the issue is fixed.

I updated my branch accordingly.

Issue #836, opened 2020-05-01
Make it possible to filter offline contacts


Sylvhem:
Currently, when clicking on the “Start Conversation” button, Dino displays a window listing all your contacts. It could be useful to add a toggle button in that window to only list contacts that are currently online.

licaon-kter:
Offline contacts have a grey out avatar in chats. Maybe the Contacts view should do the same.

This was asked in the support room this week iirc.

Sylvhem:
> Offline contacts have a grey out avatar in chats. Maybe the Contacts view should do the same.

That’s what @ThibG suggested in #313, but Fiaxh rejected this PR. Apparently, they plan to add a colored dot to indicate status.
I think it would be great to be able to filter offline contacts out of the contact list, even if there was a status indicator. Sure, it's not a problem when you only have a few contacts, but when you have several dozens of them, it may be easier for some people to just filter offline ones out.

paulmenzel:
Seconded. converse.js allows to filter the contact list with the status, and from 300 contact entries on an XMPP server, only two were online, so the filter helped a lot.

PR #1520, opened 2023-12-08
Show contact status in "Start chat" window

### LINKED ISSUE #139: [FR] Show contact status in "Start Chat" window
The following information can be seen for open conversations, but not in the "Start Chat" window. It would be useful to be able to see it in both places:

- [ ] Online / offline (shown as gray / color)
- [ ] Status (available / away / ...)
- [ ] Resource

Low priority and not shown anywhere at the moment:
- [ ] Status message

eerielili:
Should close #139.

online:
image

away and xa (xa will be displayed as "Extended Away", same color):
image

dnd:
image

offline:
image

**Note**: displays the presence status at the moment of opening the window, not synced or polling every five seconds.


fiaxh:
Thanks for your PR.

I feel that this approach clutters this dialog with a lot of text, color and details that might be too specific for an overview dialog. Especially when a user is online with multiple devices.

My suggestion would be to have a small icon (smaller than text size) behind every name: A filled-in dot for an online user, a outlined dot for an offline user, a circle with clock hands for away and a circle with a bar for do-not-disturb. I'd suggest to merge all available devices into a single information: If one device is on do-not-disturb, the contact is do-not-disturb, no matter the other devices. If one device is away and another one online, the contact is online. This should provide the user with a general overview of the status of a contact, while not cluttering the dialog too much.

eerielili:
@fiaxh Good point, it does clutter after some thought and it can get confusing with 3 devices. Thanks you for the feedback.
I came up with this from your suggestions (there's blank space before the @ because i edited out the test username):

Online:
image


DND:
image

Away and extended away:
image

Offline:
image


There's also the "chat" status with this icon:
image
But i couldn't test it out.

eerielili:
I did expand on the issue https://github.com/dino/dino/issues/836 while at it, but I'll do it in a separate branch.

main/src/ui/add_conversation/list_row.vala
```diff
@@ -8,27 +8,55 @@ namespace Dino.Ui {
...
     }
 
     public ListRow() {}
 
+    private void set_status_dot(StreamInteractor stream_interactor, Jid jid, Account account){
+        Gee.List<Jid>? full_jids = stream_interactor.get_module(PresenceManager.IDENTITY).get_full_jids(jid, account);
+        string presence_str = null;
+        if (full_jids != null){
+            int devices = 0;
+            string newline_if_more_devices;
```
fiaxh, line 41: This variable isn't used anymore

main/src/ui/add_conversation/list_row.vala
```diff
@@ -8,27 +8,55 @@ namespace Dino.Ui {
...
+                Jid full_jid = full_jids[i];
+                presence_str = stream_interactor.get_module(PresenceManager.IDENTITY).get_last_show(full_jid, account);
+                if(presence_str != null) devices += 1;
+                newline_if_more_devices = devices > 1 ? "\r\n" : "";
+                switch(presence_str){
+                        case "dnd": this.status_dot.set_from_icon_name("dino-status-dnd"); i = full_jids.size; break; // dnd detected = marked as dnd for all devices
+                        case "online": this.status_dot.set_from_icon_name("dino-status-online"); i = full_jids.size; break;
+                        case "away": this.status_dot.set_from_icon_name("dino-status-away"); break;
+                        case "xa": this.status_dot.set_from_icon_name("dino-status-away"); break;
+                        case "chat": this.status_dot.set_from_icon_name("dino-status-chat"); break;
```
fiaxh, line 52: The way this is written it doesn't allow prioritization of statuses, as the order of resources in the full_jids list affects the outcome. For example, if the user has one online and one dnd device, we'd want the status to be dnd, however if the list is [online, dnd], this routine would return online.
One solution would be to first collect all statuses in a list and then check one after another:
- If the list contains "dnd", then return dnd
- If it contains "chat", then return chat
- If it contains online, return online

And so on.

main/src/ui/add_conversation/list_row.vala
```diff
@@ -8,27 +8,55 @@ namespace Dino.Ui {
...
+        if (full_jids != null){
+            int devices = 0;
+            string newline_if_more_devices;
+            for (int i = 0; i < full_jids.size; i++) {
+                Jid full_jid = full_jids[i];
+                presence_str = stream_interactor.get_module(PresenceManager.IDENTITY).get_last_show(full_jid, account);
+                if(presence_str != null) devices += 1;
+                newline_if_more_devices = devices > 1 ? "\r\n" : "";
+                switch(presence_str){
+                        case "dnd": this.status_dot.set_from_icon_name("dino-status-dnd"); i = full_jids.size; break; // dnd detected = marked as dnd for all devices
```
fiaxh, line 48: You can use the constants from xmpp-vala/src/module/presence/stanza.vala instead of strings here, that avoids pulling protocol details into higher level code. For example `Presence.Stanza.SHOW_DND` for dnd.

fiaxh (commented):
I played a bit around with the layout and made some changes: The status icon is behind the name instead of in front of it, the icon is slightly smaller, and the offline icon is an outlined cicle.

Apart from that I'm not convinced of the filter for online status and would like to keep it out of the PR.

main/src/ui/add_conversation/list_row.vala
eerielili: @fiaxh Oh yeah thinking about it it's more suitable for presence than an ole' switch statement, will do.

main/src/ui/add_conversation/list_row.vala
eerielili: Yeah I noticed that, oops! I did have an amend that I'll force push today

main/src/ui/add_conversation/list_row.vala
eerielili: Will do, I was lazy for those one and didn't want to type the full name of the constants. That's corrected now.

eerielili:
Thanks for the feedback and help. I saw you pushed some changes on my branch shifting some things around (to put the status icon in front for example) so I integrated them in my commit. Ditched the switch statement for this since it's better for prioritizing status:
https://github.com/dino/dino/blob/cf86892dad49351018036db179e5c43569e0c5d4/main/src/ui/add_conversation/list_row.vala#L36-L72


I moved the search online contacts functionality to another branch and force-pushed my present fork to https://github.com/dino/dino/commit/cf86892dad49351018036db179e5c43569e0c5d4.

Kind regards,
lili

fiaxh:
@eerielili I thought I had followed up on this PR, but it seems I didn't actually sent the message. Sorry about that.

My feedback was:
- With your implementation, the first status in the list "wins" and is displayed. For example, if the contact is online from the first device, and dnd on the second, the result should be dnd, not online.
- The variables `devices` and `newline_if_more_devices` aren't used
- You can use the the constants from Xmpp.Presence.Stanza instead of writing the status strings
- No need to pass the jid and account to set_status_dot, they're properties

I pushed your code in 00188bdf9bd91de34060eed195c9fd7a04c8871f and did the mentioned improvements in 8b15417e0f6e90ef510ee70df7d32dbb7ce79393. I also made some other adjustments

- Moved the status behind the name
- Only shows a status if Dino is online
- I wasn't convinced by the filter feature, so I left that out for now
- Update the status if it changes after the dialog was opened

I haven't tested them for a longer time period yet, so please let me know when you find issues.

For future changes on this dialog, note that we want to rewrite the dialog in the future (as we already did with the contact details, accounts and settings dialogs). While it will still take some time, there is a balance to strike between already getting some changes in right now and not putting too much effort into the old implementation anymore.

eerielili:
@fiaxh No problem, duly noted. I'll keep it in mind when the new UI will come out.
I assume "join conference" and "start chat" will be on their own tab, like what was made with account and settings dialog?

>I wasn't convinced by the filter feature, so I left that out for now
Yeah after your feedback I moved it to another branch.

>Update the status if it changes after the dialog was opened
that is neat, i was confused by the signals and didn't touch them much

I prefer waiting out the new UI, I'll open another PR then (sorry for the wrong push, I added one bajillion participants)

Issue #962, opened 2020-12-10
Show subject and description in conference details dialog


mimi89999:
Hello,
I noticed that sometimes if the MUC subject is too long, it doesn't fit in the titlebar subtitle. I also noticed that it's not possible to see MUC description.
I propose to add both fields to conference details dialog as a solution to that problem.

pravi:
I have also noticed this, and sometimes this contains important links which cannot be opened in dino, I have to use gajim just to see the links.

fiaxh:
Since e2c34bf2235c9f85fc91de9c0f1b74858f4ef89e you can see the full topic in the conference details dialog.

anubister:
Using Dino 0.4.4-1~bpo12+1 (on Debian 12) which should onboard this commit, it still does not work. Does it actually work for someone? 👍/👎
![grafik](https://github.com/user-attachments/assets/9dee1516-785e-4bf4-a23e-c32f3d525e1c)

PR #1523, opened 2023-12-12
Enable hyperlinks in topic text to be clicked

### LINKED ISSUE #1042: Clickable URL in room status
Hi,

Room status often shares a link (to wiki, room archives, etc.) . Dino does not make this URL clickable. Would you mind to implement such feature ?

Regards,
Étienne

eerielili:
should close #1042 and save some clicks

before
image

after
image

### DIFF
--- a/main/src/ui/conversation_view_controller.vala
+++ b/main/src/ui/conversation_view_controller.vala
@@ -176,12 +176,17 @@ public class ConversationViewController : Object {
     }
 
     private void update_conversation_topic(string? subtitle = null) {
+        string topic_markup;
         if (subtitle != null) {
             conversation_topic = Util.summarize_whitespaces_to_space(subtitle);
+            topic_markup = Util.parse_add_markup(conversation_topic, null, true, true);
+            conversation_topic = topic_markup;
         } else if (conversation.type_ == Conversation.Type.GROUPCHAT) {
             string? subject = stream_interactor.get_module(MucManager.IDENTITY).get_groupchat_subject(conversation.counterpart, conversation.account);
             if (subject != null) {
                 conversation_topic = Util.summarize_whitespaces_to_space(subject);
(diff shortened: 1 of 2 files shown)


licaon-kter:
But what if the url is not visible because it's further down the text?

main/src/ui/conversation_view_controller.vala
```diff
@@ -178,10 +178,12 @@ public class ConversationViewController : Object {
     private void update_conversation_topic(string? subtitle = null) {
         if (subtitle != null) {
             conversation_topic = Util.summarize_whitespaces_to_space(subtitle);
```
fiaxh, line 180: Please use a dedicated variable here. `conversation_topic` is synced with the actual label in the UI, and putting raw user input into a label that has `use_markup=true` can lead to issues.

main/src/ui/conversation_view_controller.vala
```diff
@@ -178,10 +178,12 @@ public class ConversationViewController : Object {
...
+            conversation_topic = Util.parse_add_markup(conversation_topic, null, true, false);
```
fiaxh, line 181: You could also set `parse_text_markup = true`

fiaxh (changes requested):

main/src/ui/conversation_view_controller.vala
eerielili: @fiaxh done, added a `string topic_markup` at top of function

main/src/ui/conversation_view_controller.vala
eerielili: @fiaxh done setting `parse_text_markup = true`

eerielili:
@licaon-kter i guess one could click the conversation details, the links are clickable there too. It's a good point, I do not know how to approach it.

licaon-kter:
https://github.com/dino/dino/issues/962 i guess

fiaxh (approved):

PR #949, opened 2020-11-27
Change select contact dialog container to HdyClamp

mimi89999:

PR #1533, opened 2024-01-06
Change select contact dialog container to AdwClamp

tpikonen:
This is essentially a port of #949 to libadwaita.

Changes select contact dialog container to AdwClamp and set the width and height small enough to make it fit mobile screens.

Also remove underscores from property names.

### DIFF
--- a/main/data/add_conversation/select_jid_fragment.ui
+++ b/main/data/add_conversation/select_jid_fragment.ui
@@ -4,13 +4,13 @@
     <template class="DinoUiSelectJidFragment">
         <child>
             <object class="AdwClamp">
-                <property name="margin-start">18</property>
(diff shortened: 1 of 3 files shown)


fiaxh:
This changes the layout on the desktop quite a bit as well, and in my opinion in a negative direction.

It should be possible to make this dialog suitable for mobile phones without shrinking it for larger screens as well.
I pushed some changes, could you check whether it still works for your use case?

tpikonen:
> could you check whether it still works for your use case?

Can't see any downsides, it looks better on desktop and works just as well on mobile. Thanks!

fiaxh:
I made a few more adjustments to reduce margins to the left and right on mobile.

PR #1535, opened 2024-01-13
Set MessageMetaItem.label.selectable to false when run on touchscreen

tpikonen:
Add new function Dino.Ui.on_touchscreen() which returns true when Dino is run on a touchscreen device. Use this function to set the selectable property on MessageMetaItem labels.

This makes it possible to move from the conversation view to the conversation list by swiping.

### DIFF
--- a/main/src/ui/conversation_content_view/message_widget.vala
+++ b/main/src/ui/conversation_content_view/message_widget.vala
@@ -29,7 +29,7 @@ public class MessageMetaItem : ContentMetaItem {
     ulong style_updated_id = -1;
     ulong marked_notify_handler_id = -1;
 
-    public Label label = new Label("") { use_markup=true, xalign=0, selectable=true, wrap=true, wrap_mode=Pango.WrapMode.WORD_CHAR, hexpand=true, vexpand=true };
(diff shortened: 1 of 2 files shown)


fiaxh:
With this change, people that use laptops with touch screens would not be able to select text anymore.

Also, setting `selectable=false` might also not be what you want on mobile - mobile users might also want to select text.

I don't know the solution to those issues or whether there is one. Maybe try to search for other projects that use libadwaita, and make use of selectable labels, and see if and how they solve the issue.

tpikonen:
> I don't know the solution to those issues or whether there is one. Maybe try to search for other projects that use libadwaita, and make use of selectable labels, and see if and how they solve the issue.

The optimal solution would be to fix GtkLabel (or make a new label widget) to have a separate selection mode. AFAIK this does not exist yet, but for now this hack has worked fine for me.

fiaxh:
> but for now this hack has worked fine for me.

As I just mentioned, it breaks text selection on laptops with touch screens. This is a blocker for merging.

eerielili:
@fiaxh  @tpikonen Maybe also checking own ressource stanza to check if on desktop/laptop should do the trick ?

fiaxh:
Closing this due to previously mentioned issues.

PR #1539, opened 2024-01-20
Fix gcc error when compiling openpgp plugin

eerielili:
gcc was complaining about _**error**: non-static declaration of 'gpgme_global_mutex' follows static declaration_ in `plugins/openpgp/src/gpgme_fix.c`

After this I was able to compile the plugin and test it fine.


eerielili:
`git clean -fdx` fixed the issue. Oops! And thanks @licaon-kter for the tip

PR #1540, opened 2024-01-20
OPENPGP PLUGIN: Don't list expired/revoked GPG keys

### LINKED ISSUE #91: Do not allow revoked GPG keys
I have 3 private GPG keys on my system of which 2 are revoked. Those should not be able to select in the settings menu.

eerielili:
@fiaxh closes #91

also see https://valadoc.org/gpgme/GPG.Key.html for reference

### DIFF
--- a/plugins/openpgp/src/account_settings_entry.vala
+++ b/plugins/openpgp/src/account_settings_entry.vala
@@ -69,7 +69,7 @@ public class AccountSettingsEntry : Plugins.AccountSettingsEntry {
             return;
         }
         if (keys.size == 0) {
-            label.set_markup(build_markup_string(_("Key publishing disabled"), _("No keys available. Generate one!")));
+            label.set_markup(build_markup_string(_("Key publishing disabled"), _("No keys available. Generate one or check if your key isn't expired or revoked!")));
             return;
         }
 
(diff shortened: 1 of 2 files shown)


eerielili:
Update: I did find a way. I'll need to modify the encryption_activated method in encryption_list_entry.vala. Working on it.

eerielili:
I did add a warning in the account panel:
image

eerielili:
>It still doesn't trigger if the key expires or revokes while dino is in use

I did manage to do that without needing to click on left pane or switching back and forth conversation.
It involves calling method `on_encryption_activated` in `send_text()` method in `main/src/ui/chat_input/chat_input_controller.vala` if PGP is the encryption method, triggering the checks in the plugin.

Now, if you try to send a message but the key has expired or has been revoked, the input should be blocked with the following depending on your situation:
image
image

With this, I think all bases are covered.

@NicoHood @cjj20 Feel free to test and give feedback anytime

eerielili:
@fiaxh Adapted for new UI:
image

eerielili:
@fiaxh
I adapted the code to the new account settings UI and also simplified my code compared to last time.
Do you think I should split the input blocking code in a separate commit?

PR #1541, opened 2024-01-21
Fix crash when toggling on and off an account very fast (#1505)

### LINKED ISSUE #1505: assertion failed: (conversations.has_key(account))
```
libdino:ERROR:/media/d/temp/git/dino/libdino/src/service/conversation_manager.vala:35:dino_conversation_manager_create_conversation: assertion failed: (conversations.has_key(account))
Bail out! libdino:ERROR:/media/d/temp/git/dino/libdino/src/service/conversation_manager.vala:35:dino_conversation_manager_create_conversation: assertion failed: (conversations.has_key(account))
Aborted
```
IDK how it happened, probably because I switched an account on and off too fast.

eerielili:
Should close #1505.
I tried clicking as soon as the switch becomes sensitive repeatedly and it doesn't crash anymore.
There's still some lines whining about:
`dino_connection_manager_connection_reset: assertion 'self != NULL' failed`
but it doesn't crash.

### DIFF
--- a/main/src/ui/manage_accounts/dialog.vala
+++ b/main/src/ui/manage_accounts/dialog.vala
@@ -183,6 +183,7 @@ public class Dialog : Gtk.Dialog {
     }
 
     private void populate_grid_data(Account account) {
+        active_switch.sensitive = false;
         active_switch.state_set.disconnect(change_account_state);
 
         picture.model = new ViewModel.CompatAvatarPictureModel(stream_interactor).add_participant(new Conversation(account.bare_jid, account, Conversation.Type.CHAT), account.bare_jid);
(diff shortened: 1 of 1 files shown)


fiaxh:
Thanks for your PR.

The reason why Dino crashes is that messages are being received on an old connection, even though the account is disabled. Dino then tries to process those messages, but some required structures no longer exist.

The problem isn't exclusive to the connection process. It is just more likely to happen while connecting since lots if messages are being received. Thus, this change doesn't actually solve the issue. After some toggling, it still crashes for me, as the actual problem still remains.

Also, you mentioned that you see CRITICALS on your console. Those are actually relevant, because the program state afterwards can be unsafe, and issues might arise.

It would be great if we can fix this issue, however the UI isn't the place to fix it.

PR #1543, opened 2024-01-24
Make OMEMO plugin UI adaptive

tpikonen:
Print OMEMO fingerprints with shorter lines and with a slightly larger font, this makes them fit narrow dialogs and easier to read.

Adapt contact details dialog: Allow resizing, add AdwClamp and replace separate ScrolledWindows by a single whole-dialog scroll area.

Fix various alignment, margin and size issues in OMEMO related UI elements.


licaon-kter:
UI edits? Attach some before and after pics, c'mon :)

fiaxh:
The OMEMO keys of one's own account were made adaptive as part of 680f0dd0a768b201b62899905c38210e834dde91. For the OMEMO keys of a contact, this will be done by reusing that code as far as possible.

Issue #261, opened 2018-01-03
Feature Request: Password Change : XEP-0077


philipishungry:
Feature Request: Password Change : XEP-0077

The 'jabber:iq:register' namespace enables a user to change his or her password with a server or service. Once registered, the user can change passwords by sending an XML stanza of the following form:

Example 18. Password Change
```
<iq type='set' to='shakespeare.lit' id='change1'>
  <query xmlns='jabber:iq:register'>
    <username>bill</username>
    <password>newpass</password>
  </query>
</iq>
```
https://xmpp.org/extensions/xep-0077.html#usecases-changepw

AlexandriaOL:
Hi, I'm bumping this and stating that this not being present has single-handedly turned me away from Dino for the foreseeable future. This stops adoption from the get-go, at least in my case, because I was handed an account and advised to change the password. Why would I use a client that does not allow me to do that?

In addition, the fact that this issue is listed under "Feature Requests" and not "Security Issues" honestly concerns me, the fact it's implementation as a feature consists of sending a single XML form, but is put in the same category as much, much more complex features and longer-term goals like "Support Audio/Video Chat", and "Windows support" makes me seriously doubt the maintainers of the project, and makes me wonder why the project does not see the lack of an ability to change your password as a security feature.

Thanks for reading.

CryptGoat:
Another bump here. I would love to roll out Dino as the default client for my organization which exclusively runs Linux systems but not being able to change the account password is sadly a huge blocker. If I setup a new account I want the owner of the account to be able to change the password themselves. It indeed is a security concern because some people might even forget to change their password at all. Pretty much any other common XMPP client allows to change their password from their GUI. Please consider to give this issue more attention.

mwild1:
I'm surprised (or maybe not?) that organizations are handing around passwords in this way, instead of using better methods such as LDAP (where users have the same password across services, managed elsewhere) or account creation invitations (so the user can set their own password immediately when they create the account).

I guess what I'm trying to say is: I think this is a feature worth having in Dino, but there are other (better) ways of managing user accounts in an organization, so I don't think this issue is strictly a security concern (allowing people to hand passwords around is the security concern - e.g. you can't guarantee the user will actually change it).

CryptGoat:
> I'm surprised (or maybe not?) that organizations are handing around passwords in this way, instead of using better methods such as LDAP (where users have the same password across services, managed elsewhere) or account creation invitations (so the user can set their own password immediately when they create the account).
[...]

PR #1560, opened 2024-03-12
Change password

smalishevskiy:
Fix issue #261

Fix problem with create XEP-0077. Implement more user frendly UI for change password

change_password

### DIFF
--- a/main/src/ui/manage_accounts/change_password_dialog.vala
+++ b/main/src/ui/manage_accounts/change_password_dialog.vala
@@ -11,14 +11,13 @@ namespace Dino.Ui{
     public class ChangePasswordDialog : Gtk.Dialog {
     
         [GtkChild] private unowned Button change_password_button;
-        [GtkChild] private unowned Stack change_password_stack;
+        /*[GtkChild] private unowned Stack change_password_stack; */
         [GtkChild] private unowned Button cancel_button;
-        [GtkChild] private unowned Entry current_passwd_entry;
-        [GtkChild] private unowned Entry new_passwd_entry;
-        [GtkChild] private unowned Entry confirm_new_passwd_entry;
+        [GtkChild] private unowned Adw.PasswordEntryRow current_password_entry;
+        [GtkChild] private unowned Adw.PasswordEntryRow new_password_entry;
+        [GtkChild] private unowned Adw.PasswordEntryRow confirm_new_password_entry;
         [GtkChild] private unowned Label change_password_error_label;
 
-        private bool are_forms_empty;
         private Account account;
         private StreamInteractor stream_interactor;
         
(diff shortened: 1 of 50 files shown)


main/src/ui/manage_accounts/change_password_dialog.vala
```diff
@@ -0,0 +1,100 @@
...
+    
+        [GtkChild] private unowned Button change_password_button;
+        [GtkChild] private unowned Stack change_password_stack;
+        [GtkChild] private unowned Button cancel_button;
+        [GtkChild] private unowned Entry current_passwd_entry;
+        [GtkChild] private unowned Entry new_passwd_entry;
+        [GtkChild] private unowned Entry confirm_new_passwd_entry;
+        [GtkChild] private unowned Label change_password_error_label;
+
+        private bool are_forms_empty;
```
fiaxh, line 21: This variable is set but never read.

fiaxh (commented):
Thanks for your PR!

Instead of using a Grid-Layout with Labels and Entries, I'd suggest to use Adw.PasswordEntryRow (within an Adw.PreferenceGroup, whithin an Adw.PreferencePage). It gives us some password field features for free and would look like this (this is from the gnome settings password change dialog):

image

Please remove updates to the .po and .pot files, they make PRs harder to review. We'll update the files independently afterwards.

We try to get rid of references to the StreamInteractor from our UI code, and I'd prefer to not introduce any more in the meanwhile. Could you please create a PasswordChangeDialogModel, that has a function to change the password (basically doing the StreamInteraction calls there instead of in the UI code)? The dialog UI code should contain a reference to the model and call the respective function there.

As a general style remark, you abbreviate variable names in various places (passwd, pw, chpw, btn, s, a, ...). We typically don't do that in our codebase and I'd like to keep the variable naming style at least somewhat consistent.

smalishevskiy:
> We try to get rid of references to the StreamInteractor from our UI code, and I'd prefer to not introduce any more in the meanwhile. Could you please create a PasswordChangeDialogModel, that has a function to change the password (basically doing the StreamInteraction calls there instead of in the UI code)? The dialog UI code should contain a reference to the model and call the respective function there.

Can you point me example of model in current code? That is not clean for me where I should create the model's class.

fiaxh:
@smalishevskiy `Dino.Ui.ViewModel.PreferencesWindow` is an example, however the view model for the change password dialog would be much simpler. It could roughly look like this:
```
public class Dino.Ui.ViewModel.ChangePasswordDialog : Object {
    private StreamInteractor stream_interactor;
    public ChangePasswordDialog(StreamInteractor stream_interactor);
    public change_password(Account account, string password);
}
```
The UI would be given an instance of this class. When a new password was entered, the UI would call `change_password` on this class. `change_password` would make the call to the backend to change the password.

fiaxh:
I implemented the changed that I've suggested and merged this in f8c004630f56914438fa1b114530f639748e41c1.

PR #1570, opened 2024-04-02
Allow creating new bookmark when there are no existing bookmarks

kkonsw:
Currently first-time join (with an invite and without any bookmarks) fails to create a new bookmark for a conversation due to early return in the `set_autojoin` function.

This PR removes early return to allow the call to the `add_conference` function (to create a new bookmark).


fiaxh:
Thanks for your PR.

Did you test this change with legacy bookmarks (xep 0048) or bookmarks2 (xep 402)?

I see one issue with this change in combination with the [`add_conference` function](https://github.com/dino/dino/blob/master/xmpp-vala/src/module/xep/0048_bookmarks.vala#L50) from the legacy bookmarks implementation. It uses the same `get_bookmarks` function and then operates on the return value. Since the return value might be null, we should to create a new Set in that case.

PR #1572, opened 2024-04-16
When first built does not display revoked keys (but when reopened it does)

cjj20:
When built it does not display revoked keys but only displays one of the available gpg keys. However when Dino is closed and then reopened the revoked key is shown along with the other gpg keys. Seems to reset itself that way so to speak.

### DIFF
--- a/plugins/openpgp/src/account_settings_entry.vala
+++ b/plugins/openpgp/src/account_settings_entry.vala
@@ -1,5 +1,8 @@
+
 using Dino.Entities;
 using Gtk;
+using Gee;
+
 
 namespace Dino.Plugins.OpenPgp {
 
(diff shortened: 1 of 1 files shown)


cjj20:
@fiaxh Improvement from the last PR in that when freshly built revoked keys are hidden. However when Dino is closed then reopened, seems to reset and the revoked keys can be once again seen from the dropdown. Will continue to improve/fix this but any comments as to why new issue is happening?

eerielili:
Hello, you should close this PR and push your changes on your [revokedkeys](https://github.com/cjj20/dino/tree/revokedkeys) branch.

fiaxh:
Closing this PR. It was made as part of a gsoc application and feedback was not addressed.

PR #1595, opened 2024-06-23
Allow self-signed .onion file transfer certs (#1149)

### LINKED ISSUE #1149: Allow self-signed certificates for .onion http file upload up+down urls

matthewrfennell:
Most Certificate Authorities don't support issuing X.509 certificates for onion sites. However, it can still be useful to provide a certificate over Tor in some circumstances, for instance to tie your alphanumeric Tor address to your site's main identity.

Therefore, many Tor services provide self-signed certificates. This is OK, since the onion service itself guarantees that you are connecting to the entity you think you are.

Dino already allows self-signed certs when communicating over Tor (see 81a5505). However, the same exception does not exist yet for HTTP uploads and downloads - causing these to fail over Tor.

Therefore, in this commit, we add the same exception for uploads/downloads, by passing the host of the upload/download urls to the already existing invalid certificate connection handler.

Note that this handler only allows certificates with type TlsCertificateFlags.UNKNOWN_CA. This means the certificate of your server must also include the onion http upload and download URLs in its certificate - otherwise, the file transfer will fail with TlsCertificateFlags.BAD_IDENTITY.

Closes #1149

### DIFF
--- a/plugins/http-files/src/file_provider.vala
+++ b/plugins/http-files/src/file_provider.vala
@@ -120,6 +120,11 @@ public class FileProvider : Dino.FileProvider, Object {
         var head_message = new Soup.Message("HEAD", http_receive_data.url);
         head_message.request_headers.append("Accept-Encoding", "identity");
 
+#if SOUP_3_0
+        string transfer_host = Uri.parse(http_receive_data.url, UriFlags.NONE).get_host();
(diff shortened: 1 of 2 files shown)


fiaxh:
Thanks for your PR!

`accept_certificate` is only available in `Soup.Message` since version 3. Please use an ifdef to only compile the code if `SOUP_3_0` is set.

fiaxh:
Thanks!

PR #1613, opened 2024-08-19
Use new call.incoming notification category

Pigpog:
The call.incoming notification category is pending merge to FreeDesktop. See https://gitlab.freedesktop.org/xdg/xdg-specs/-/merge_requests/50 for more info.

~~On the condition that the FreeDesktop MR merges,~~ this change to Dino will allow notification daemons to properly respond to incoming calls with appropriate feedback.


Pigpog:
FreeDesktop has merged the call categories into the XDG notifications standard. Marked as ready for review.

fiaxh:
Thanks

PR #1656, opened 2025-01-14
Fix fetching avatars via PubSub

bodqhrohro:


fiaxh:
Please write something about this PR. What issue are you experiencing? What cause did you identify?

bodqhrohro:
I was already told in the MUC that I just understand XEP-0084 wrong, and the former form is a valid PubSub request too. So my transport needs to support both variants.

Yet I have seen a TODO to support multiple nodes in `request_all`, which this patch implements in flexible manner, so it still might be useful.

But it turns out to be destructive on my configuration somehow, as the former variant of the stanza reaches my transport, and the new one doesn't. For now I'm only sure that the transport is not at guilt, as according to `tcpdump` on port 5347, the stanza is missing from the traffic between Prosody and the transport. Now I need to figure out if Dino lies and does not actually send the stanza despite printing it with `--print-xmpp=all`, or Prosody or some of its modules are at guilt. Which is harder to determine due to an encrypted C2S connection.

fiaxh:
From what I understand, this change doesn't solve an issue in the end, so I'll close the PR. I don't remember why I wrote that TODO, but we can add the code when it's needed and can be tested.

PR #1657, opened 2025-01-15
fix: Query duplication in global message search (#1636)

### LINKED ISSUE #1636: Strange behavior in message search
1. Click 'Search messages' 🔍 
2. Search field says e.g. `in:chat@dino.im`
3. Type e.g. "hello" in the end of the search field so that the input reads `in:chat@dino.im hello`
4. Press Enter ⏎
5. Observe strangeness, e.g. the input might change to `helloin:chat@dino.im` or `in:someother@chat.example`

eerielili:
Closes #1636.
The only behavior that this commit do not fix is when you do the following upon opening global message search:

- BackSpace
- Ctrl+A
- BackSpace to delete
- Press Enter

It'll still set the search field to `in:x@x` or `with:x@x` depending on the current chat type (groupchat or not). In my opinion, it's really not so bad.

### DIFF
--- a/main/src/ui/global_search.vala
+++ b/main/src/ui/global_search.vala
@@ -98,6 +98,8 @@ public class GlobalSearch {
     }
 
     private void on_key_released(uint keyval, uint keycode, Gdk.ModifierType state) {
+        if (!auto_complete_overlay.visible) return;
+
         if (keyval == Gdk.Key.Return) {
             auto_complete_list.get_selected_row().activate();
         }


eerielili:
The other behavior previously mentioned has been fixed in a force push, see [e3b67c5](https://github.com/dino/dino/commit/e3b67c5044b02a48c4185efe011a7038be8a0564). It involves checking for a non-zero text length in the search entry.

mar-v-in (changes requested):
The underlying issue is that the custom handling for `on_key_released` to accept the selection from the auto complete overlay is even invoked when the overlay is hidden.

Instead of adding complexity to this handling to handle the double enter pressing, the solution would rather be to add the same `!auto_complete_overlay.visible` check we also have as a first line of `on_key_pressed`.

```suggestion
    private void on_key_released(uint keyval, uint keycode, Gdk.ModifierType state) {
        if (!auto_complete_overlay.visible) return;

        if (keyval == Gdk.Key.Return) {
            auto_complete_list.get_selected_row().activate();
        }
    }
```

eerielili:
Thanks @mar-v-in for the remarks, you're right it's just smarter and simpler to just check the suggestion overlay state.
I'll compile and test it and push-it once it's OK.

mar-v-in (approved):

PR #1676, opened 2025-02-28
Make resource header not send client name

### LINKED ISSUE #1646: Make the resource header more private
### Description
Currently, Dino sets/sends a resource header that gives identifiers to the server. \
   It consists of the client name "dino" and an 8-character string of lowercase letters (a-z) and digits (0-9):

*Example:* `dino.cfg231e9`

To mitigate the Resource Header privacy issue, I propose setting a random resource header every time the application is launched. \
It would look something like this:

```vala
using GLib;

public static string generate_random_string(int length) {
    const string chars = "abcdefghijklmnopqrstuvwxyz0123456789";
    Random random = new Random();
    string result = "";

    for (int i = 0; i < length; i++) {
        int index = random.next_int_range(0, chars.length());
        result += chars[index];
    }

    return result;
}

public static void main(string[] args) {
    // Assuming you have a function to set the resource header
    string random_resource = generate_random_string(13);
    
    // Set the resource header (this is a placeholder, replace with actual method)
    set_resource_header(random_resource);
}

// Placeholder function for setting the resource header
public static void set_resource_header(string resource) {
    // Implementation to set the resource header in the XMPP client
    print("Setting resource header to: %s\n", resource);
}
```

You could also look at *Gajim*, which I believe uses both upper and lower case letters. \
When you leave the resource field blank on *Gajim*, it sets a random one:

	`const string chars = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789";`

souchikjoardar201:
change resource header to not send client name "dino"

This closes #1646

### DIFF
--- a/libdino/src/entity/account.vala
+++ b/libdino/src/entity/account.vala
@@ -74,7 +74,13 @@ public class Account : Object {
     }
 
     private static string get_random_resource() {
-        return "dino." + Random.next_int().to_string("%x");
+        const string chars = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789";
+        Random random = Random();
+        string resource = "";
+        for (int i = 0; i < 8; i++) {
+            resource += chars[random.next_int_range(0, chars.length)];
+        }
+        return resource;
     }
 
     public bool equals(Account acc) {


fiaxh:
Even if an XMPP client uses a random resource, it would still be identifiable by its capabilities and by the way it behaves. 
Examples for that include:

- The order of XML elements that are exchanged
- The format of message IDs
- The exact wording of fallback strings

Using a random resource would only provide a false sense of privacy.

PR #1677, opened 2025-02-28
Dont send client name stanza to other clients

souchikjoardar201:
## Modify`flag.vala` to avoid adding client name in presence stanzas:

These changes ensure that the client name is not included in the presence stanzas sent to other clients, while still maintaining the functionality to show if you are online.

### DIFF
--- a/xmpp-vala/src/module/presence/flag.vala
+++ b/xmpp-vala/src/module/presence/flag.vala
@@ -35,6 +35,8 @@ public class Flag : XmppStreamFlag {
     }
 
     public void add_presence(Presence.Stanza presence) {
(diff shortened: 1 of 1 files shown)


fiaxh:
The presences you modify are the received presences, not the one we send out. Furthermore, dino never sets a subnode with name "client-name" in the presence, so this wouldn't do anything even for outgoing presences.

On the topic of "hiding" the client identity: Even if you remove the name "dino" from stanzas, the client will still be identifiable through other means (e.g. order of XML elements, wording of fallback strings, ...). This change would only provide a false sense of privacy.

Issue #1092, opened 2021-08-15
No inbound audio on calls


Anomnomnomous:
Whenever placing or receiving calls in Dino, after Dino has been running for a hour or so, I don't hear incoming audio. The person on the other end still hears me fine. If I restart Dino, calls will work fine again for a while. I tested this with the main branch and the libhandy branch. It also has this issue sometime if I switch to my cellular connection from wifi on my Pinephone. Here's the errors I saw:

````
(dino:83567): GStreamer-WARNING **: 14:52:28.178: ../gst/gstpad.c:4621:gst_pad_push_data:<encode-speex-2361928233_rtp_depay:src> Got data flow before segment event

(dino:83567): GStreamer-WARNING **: 14:52:28.178: ../gst/gstpad.c:4368:gst_pad_chain_data_unchecked:<encode-speex-2361928233_speex_decode:sink> Got data flow before segment event

(dino:83567): rtp-WARNING **: 14:52:37.818: plugin.vala:130: Warning in pipeline: Can't record audio fast enough
````
It repeats that last error a few times then:
````
(dino: 83567): ice-WARNING **: 14:52:41.793: dtls_srtp.vala:51: SRTP packet failed the message authentication check (2)
````
It repeats that a couple of times then goes back to repeating the "Can't record audio fast enough" error.

I've reproduced it on systems running Debian 11 with the main branch of Dino (0.2.0\~git115.20210607.ce292d03-1) and Mobian 12 with the libhandy branch of Dino (0.2.0\~git134.20210729.093491d) over wired, wireless, and cellular.

lorenz:
I'm also experiencing this. Most of the time it fixes itself after ~30s though. Really weird.

PR #1679, opened 2025-03-07
p/rtp: do not keep around and reuse GstPipeline

### LINKED ISSUE #1451: After a call starts, more than 10 seconds are necessary to hear the correspondant (but they can hear me!)
In my case it's when calling or being called by someone that is using Conversations.
They can hear me quickly after the call is accepted. But I can't hear them and they are left speaking in the void for between 5 and more than 10 seconds every time.

Dino version: since when audio calls became possible: 0.3, 0.4, dev version 9a04573f

lorenz:
When the first call is made a GstPipeline is created and kept around even once the call has ended. If the system is then suspended for some time the timing of that idle GstPipeline gets messed up. If later another call comes in it can take from a few seconds up to a minute for audio to be played back without hiccups. The pacing of the incoming RTP packets looks good and no different from when it later starts working. This reproduces on multiple different system types.

Fix this by tearing down the GstPipeline once the last stream is gone. I've been running this patch on multiple systems for some time and it has completely resolved the issue.

Likely fixes #1451 #1092 and maybe some others.

### DIFF
--- a/plugins/rtp/src/plugin.vala
+++ b/plugins/rtp/src/plugin.vala
@@ -269,6 +269,9 @@ public class Dino.Plugins.Rtp.Plugin : RootInterface, VideoCallPlugin, Object {
     public void close_stream(Stream stream) {
         streams.remove(stream);
         stream.destroy();
+        if (streams.is_empty) {
+            destroy_call_pipe();
+        }
     }
 
     public void shutdown() {


mar-v-in (approved):

mar-v-in:
This is effectively included in 50eeb2f16dfa4342d50086404da068f5426cbe3c (but that also ensures the pipe is not closed while it is used to display a video widget, which shouldn't happen in regular usage, but could happen when having an outgoing, ringing video call during another call). Thanks.

PR #1687, opened 2025-03-26
Gray out avatars of counterparts that are offline.

### LINKED ISSUE #1612: cannot see offline status from roster anymore
There seems to be a recent regression, it is not possible to see the online/offline status from contacts anymore.
It used to be indicated by the avatars becoming gray/grey scale (since #190 was closed in 2018).

For me this stopped working, I do not know when, but I see the problem with `0.4.1~git110.20240729.b0ff90a`
via Opensuse package `dino-0.4.1~git110.20240729.b0ff90a-lp155.944.1.x86_64`.

expeehaa:
Grayscaling avatars of offline conversation counterparts has existed until a rewrite in commit db3b0d5f233ee3587ae54f8f035222cb098b11dd, where it was removed without any mention. This PR attempts to fix #1612.

I tried to reintroduce the feature, but as I’m new to Vala and the Dino source code, there are probably much better ways to implement it, and I’ld be happy to change the PR accordingly. I wouldn’t be surprised if there also are unintended side effects that need to be addressed.

### DIFF
--- a/main/data/style.css
+++ b/main/data/style.css
@@ -129,6 +129,10 @@ picture.avatar {
     border-radius: 3px;
 }
 
+picture.avatar .unavailable {
+    filter: grayscale(100%);
+}
+
 /* Overlay Toolbar */
 
 .overlay-toolbar {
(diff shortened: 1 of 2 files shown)


fiaxh:
Displaying the offline status via grey-scale had issues:

- It does not work for grey-scale avatars
- It does not work for people who don't see colors
- Generally, with mobile phones being online 24/7, online status often doesn't mean much

We already introduced a new way to see the status of a contact (not just on/offline) in the contact list, and the status can also be seen when hovering a conversation.

Thus, I wouldn't want to re-introduce the behavior as it was.

PR #1688, opened 2025-03-26
Use the same path for the meson plugin install directory and Dino’s plugin search path.

expeehaa:
Before the change, the plugin install directories were determined by the `libdir` option value prepended to the hardcoded path `dino/plugins`, whereas Dino’s search path for plugins was determined by the `prefix` option value prepended to the `plugindir` option value (by default `lib/dino/plugins`).
I encountered this issue when I created an RPM spec file for openSUSE Tumbleweed that builds with meson instead of cmake. The default meson setup macro sets `--libdir=/usr/lib64`, but the `plugindir` value is not automatically updated and thus Dino installed from the generated package could not find the plugins.

As part of the fix, I took the liberty of removing the `plugindir` option.

### DIFF
--- a/libdino/meson.build
+++ b/libdino/meson.build
@@ -78,7 +78,7 @@ sources = files(
 sources += [version_vala]
 c_args = [
     '-DDINO_SYSTEM_LIBDIR_NAME="@0@"'.format(get_option('prefix') / get_option('libdir')),
-    '-DDINO_SYSTEM_PLUGIN_DIR="@0@"'.format(get_option('prefix') / get_option('plugindir')),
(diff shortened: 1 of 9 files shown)


mar-v-in:
Instead of entirely removing `plugindir`, 87dfa8e670a080c3067d20f85b6af35cd8071a87 makes it relative to `libdir`, allowing you to easily configure a custom `libdir` and have the `plugindir` adjusted as well, but also allow for more customized setups.

Debian configures `libdir` to `$lib/dino` to put dino's libraries into `$lib/dino`, but surely doesn't want the plugins in `$lib/dino/dino/plugins`. With 87dfa8e670a080c3067d20f85b6af35cd8071a87 they can set `libdir` as before and configure `plugindir` to just `plugins`.

PR #1695, opened 2025-04-12
Rename `libdino` to `dino`

catap:
Library doesn't need `lib` prefix which is added automatically.


catap:
It never was a dino issue

PR #1701, opened 2025-04-19
Voice processor: add back call to set stream delay

### LINKED ISSUE #1700: Loud static-like noise on calls with dino 0.5 using webrtc-audio-processing-0.3
When making or receiving calls with the new `dino` version, the other client's audio includes a loud static-like noise which drowns out my voice, and the debug output is littered with this:
```
voice_processor_native.cpp ProcessStream -11
```

I'm using the arch linux `dino` package, which seems to have been built with `webrtc-audio-processing-0.3`.

The `webrtc-audio-processing` source shows that the error code `-11` is `kStreamParameterNotSetError` and is returned by `ProcessStream` when the stream delay was not set.

It looks like the call to `set_stream_delay_ms` was (accidentally ?) removed in commit https://github.com/dino/dino/commit/1c365daa14e0e11bb4931b0c7891a781c5fc7084

jacquescomeaux:
This should fix #1700, assuming the one line of code mentioned there was indeed removed by accident. I've tested this and it makes the static noise go away

### DIFF
--- a/plugins/rtp/src/voice_processor_native.cpp
+++ b/plugins/rtp/src/voice_processor_native.cpp
@@ -218,6 +218,7 @@ dino_plugins_rtp_voice_processor_process_stream(void *native_ptr, GstAudioInfo *
     frame.samples_per_channel_ = info->rate / 100;
     memcpy(frame.data_, map.data, frame.samples_per_channel_ * info->bpf);
 
+    apm->set_stream_delay_ms(native->stream_delay);
     err = apm->ProcessStream(&frame);
     if (err >= 0) memcpy(map.data, frame.data_, frame.samples_per_channel_ * info->bpf);
 


mar-v-in:
Hi. The change you did is in the ifdef block for `webrtc-audio-processing-0`. If available, Dino will use the newer `webrtc-audio-processing-1` or `-2`. Which OS and version of webrtc-audio-processing did you test this against?

jacquescomeaux:
Hello. I tested this on artix linux. I just now tested again with and without the change to confirm the behavior. This time I commented out some lines in `meson.build` to make sure that `webrtc-audio-processing-0.3` was chosen for both tests, and I confirmed that the output of `meson setup build` included this line

```
Dependency webrtc-audio-processing found: YES 0.3.1 (cached)
```

Antiz96:
https://github.com/dino/dino/issues/1700 was due to a downstream packaging issue, as the `dino` Arch Linux package was compiled with the legacy webrtc-audio-processing-0.3 dependency. This is now fixed (https://gitlab.archlinux.org/archlinux/packaging/packages/dino/-/issues/5).

PR #1713, opened 2025-05-15
Fix parse incoming stateless file sharing sources.

fabrixxm:
file transfers where saved without any sources


fabrixxm:
looks like there is a similar issue when sharing files from Dino, but only in MUCs. I this case the XEP-0447 stanza does not have `<sources>` tag

fiaxh:
Thanks for your PR. Did you close the PR on purpose or was this accidentally?

I pushed your fist commit to master.

Regarding your second commit, could you explain the issue you are seeing? Stateless file shares do not have to include a source directly in the file-share, it can also be attached in a subsequent message: https://xmpp.org/extensions/xep-0447.html#attach-source

fabrixxm:
Thank you. Yeah, I wanted to send only the first commit but I got sidetracked and forgot about it.
About my comment, I opened  issue #1714  because I'm not sure about the real cause, so I'm not confident to publish a patch

PR #1736, opened 2025-07-20
Only support SystemMemory caps for video

rmader:
An increasing number of video device providers, such as `pipewiresrc` (in the upcoming 1.6 release) and `v4l2src` (since Gst 1.26), support `memory:DMABuf` caps using the special `DMA_DRM` caps. This is not expected and handled properly, thus filter out any non-default features.

---

If accepted this should be backported to stable releases.


rmader:
@mar-v-in gentle ping for this MR - any chance we can get it in? :)

mar-v-in (approved):

rmader:
Gentle ping again - can we land this? :)

Having it in would be nice for downstreams that want to backport the fix.

mar-v-in:
Finally had time to test this, lgtm and thanks for your work.

PR #1737, opened 2025-07-20
Target resolution of 720p on libcamera devices

rmader:
Most cameras needing the fallback path are provided by libcamera, which does not yet provide framerates. At the same time many such cameras support very high resolutions - often in the 4K category - and are paired with rather weak CPUs. Examples for this are popular mobile phones such as the PinePhone Pro or the OnePlus6.

In order to archive a somewhat stable FPS count in such scenarios, target a resolution of 720p as a compromise of quality and performance.

---

With this PR I can have video calls between a PinePhone Pro and a Pixel 3a with reasonable frame rate and quality (and lower energy drain). Both devices are running postmarketOS.

Related:
 - https://github.com/dino/dino/pull/1738
 - https://github.com/dino/dino/pull/1767

### DIFF
--- a/plugins/rtp/src/device.vala
+++ b/plugins/rtp/src/device.vala
@@ -400,16 +400,45 @@ public class Dino.Plugins.Rtp.Device : MediaDevice, Object {
             }
             if (best_index == -1) {
                 // No caps in first round, try without framerate
+                int target_pixel_count = 1280 * 720;
+                double target_pixel_ratio = 16.0 / 9.0;
+                double best_diversion_factor = 0.0;
+
                 for (int i = 0; i < device.caps.get_size(); i++) {
                     unowned Gst.Structure? that = device.caps.get_structure(i);
                     if (!that.has_name("video/x-raw")) continue;
                     int width = 0, height = 0;
-                    if (!that.has_field("width") || !that.get_int("width", out width)) continue;
-                    if (!that.has_field("height") || !that.get_int("height", out height)) continue;
-                    if (best_width < width || best_width == width && best_height < height) {
+                    if (!that.has_field("width") || !that.get_int("width", out width) || width == 0) continue;
+                    if (!that.has_field("height") || !that.get_int("height", out height) || height == 0) continue;
+
+                    int pixel_count = width * height;
+                    double pixel_count_diversion_factor;
+                    if (pixel_count > target_pixel_count)
+                        pixel_count_diversion_factor = (double)pixel_count / target_pixel_count;
+                    else
+                        pixel_count_diversion_factor = (double)target_pixel_count / pixel_count;
+
+                    double pixel_ratio;
+                    if (width > height)
+                        pixel_ratio = (double)width / height;
+                    else
+                        pixel_ratio = (double)height / width;
+
+                    double ratio_diversion_factor;
+                    if (pixel_ratio > target_pixel_ratio)
+                        ratio_diversion_factor = pixel_ratio / target_pixel_ratio;
+                    else
+                        ratio_diversion_factor = target_pixel_ratio / pixel_ratio;
+
+                    // 1.0 is the best possible value. Any diversion from
+                    // target_pixel_count or target_pixel_ratio increases it.
+                    double diversion_factor = pixel_count_diversion_factor * ratio_diversion_factor;
+
+                    if (best_index == -1 || diversion_factor < best_diversion_factor) {
+                        best_index = i;
                         best_width = width;
                         best_height = height;
-                        best_index = i;
+                        best_diversion_factor = diversion_factor;
                     }
                 }
             }


rmader:
@mar-v-in gentle ping for this MR, too - would love to see it getting in :)

mar-v-in:
I think this leads to somewhat unexpected results (if 1280x720 is not supported):
- The ordering of the caps matters: If the first resolution in the list is < 1280x720, it selects the largest that's < 1280x720. If the first resolution in the list is > 1280x720, it selects the smallest that's > 1280x720. So for devices without 1280x720 the selection may feel a little bit random.
- Portrait resolutions that are <1280 in width and >720 in height, or vice-versa ultra-wide resolutions are ignored (except if they are the first in the list). With the list [360x640, 540x960, 720x1280, 1080x1920], 360x640 would be selected, if the list is the other way round, it would be 1080x1920.

Also of course, better solutions in the long run are:
- automatically pick a lower resolution if we notice framerate dropping below a threshold
- allow the user to select the desired resolution
  - Settings related to A/V support are completely missing as of now. Other useful settings would include:
    - device selection before starting the call
    - disabling of echo cancellation (to improve quality if a headset is used that prevents echo by design)
    - enable/disable of hardware acceleration for video encoding/decoding
  - But this is certainly out of scope for now

For now I agree it's fine to just limit to ~720p. Maybe a good solution could be to select the resolution that's closest to 921600 pixels and if there's two equally close (e.g. 1280x720 and 720x1280), select the one where width/height ratio is closest to 16/9.

rmader:
Thanks for the quick feedback!

> The ordering of the caps matters:

Hm, yeah, I agree that it'd be better if the algorithm would be order-independent.

> Portrait resolutions

While this is possible, I'd like to highlight that native portrait resolutions are - AFAIK - pretty uncommon. Usually portrait resolutions are archived by using landscape with a 90 degree rotation - the part that we handle with the `videoflip` element. It could of course well be that this will change in the future, for example by ISPs handling the rotation in hardware, so probably worth to prepare for.

> Maybe a good solution could be to select the resolution that's closest to 921600 pixels and if there's two equally close (e.g. 1280x720 and 720x1280), select the one where width/height ratio is closest to 16/9.

That sounds like a reasonable approach, will have a look.

rmader:
@mar-v-in  Just gave this another go. I found it a bit hard to weight "closeness to target pixel count" vs "closeness to target pixel ratio" and now tried a solution that takes both into consideration. The result is most likely not perfect, but it is deterministic and has a stronger focus on pixel-count than pixel-ratio and is thus hopefully close to what you asked for.

plugins/rtp/src/device.vala
```diff
@@ -400,16 +400,41 @@ public class Dino.Plugins.Rtp.Device : MediaDevice, Object {
...
+                        pixel_ratio = (double)height / width;
+
+                    double ratio_penalty;
+                    if (pixel_ratio > target_pixel_ratio)
+                        ratio_penalty = pixel_ratio / target_pixel_ratio;
+                    else
+                        ratio_penalty = target_pixel_ratio / pixel_ratio;
+
+                    if (best_width == 0 ||
+                         (ratio_penalty * (pixel_count - target_pixel_count).abs()) <
```
mar-v-in, line 430: This gives a resolution a value of 0 if it has 1280*720 pixels, no matter the ratio. While I agree that resolutions like 960x960, 1024x900, 1152x800, 1200x768, 1440x640, 1536x600 or 1600x576 are probably not that common to appear on a device that also provides any other resolution that's 1280*720 pixels. You could do the same that you did for `ratio_penalty` and calculate a `size_penalty` (that turns 1 for 1280x720 pixels instead of 0).

Then you could calculate `cost = ratio_penalty * size_penalty` and only store `best_cost` rather than `best_ratio_penalty`, `best_pixel_count` and `best_pixel_ratio`.

rmader: Very good point, thanks! Did that and it's not just cleaner and more correct - the factors also looks much more intuitive. Just used `diversion_factor` instead of `penalty` or `cost`.

rmader:
Nice, thanks for the review! What's needed to get this (and https://github.com/dino/dino/pull/1736) landed? If it's another review, can we ping anyone?

mar-v-in:
@rmader this and the other PRs look reasonable from code, but haven't been tested yet on any of my actual devices. I will run a test with them on all devices I have before the next release and will make sure they get merged then, so they will be part of the next release.

rmader:
That sounds reasonable, thanks a lot. I'll go ahead then and propose the patches for pmOS/alpine to carry downstream already, as they appear to help quite a lot - and to get more testing already.

Edit: patches landed in https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/90486

PR #1738, opened 2025-07-20
Use gtk4paintablesink if possible

rmader:
As it's used by various Gnome apps such as Snapshot and Showtime, it's shipped by an increasing number of distros. It can serve as a drop in replacement for the internal sink while providing various advantages, such as native YCbCr/YUV, rotation and colorimetry support, resulting in improved performance and quality.

Thus use it if it is present in the system and offers the desired features.

This is partucularly helpful on phones, where the current pipeline usually requires several copies for format conversions and 90 degree camera rotations.

---

Depends on and includes:
 - https://github.com/dino/dino/pull/1825

Related:
 - https://github.com/dino/dino/pull/1737
 - https://github.com/dino/dino/pull/1775


rmader:
This also runs into an issue in `gtk4paintablesink` with buffer sizes that are not a multiple of the subsampling, e.g. odd buffer widths with yuv420 formats. Can be fixed by adopting the solution from `gtkgstsink` (will look into that).

eerielili:
Hi from #1781. On a Fedora 42 system I can accept a call from Conversations and caller sees me but my Dino's call window is empty.
I do have some errors:
```
(dino:33709): GLib-GObject-CRITICAL **: 23:27:22.412: g_object_ref_sink: assertion 'G_IS_OBJECT (object)' failed
(dino:33709): GLib-GObject-CRITICAL **: 23:27:22.413: g_object_get_is_valid_property: object class 'DinoPluginsRtpSink' has no property named 'paintable'
(dino:33709): GLib-GObject-CRITICAL **: 23:27:22.428: g_object_ref_sink: assertion 'G_IS_OBJECT (object)' failed
(dino:33709): GLib-GObject-CRITICAL **: 23:27:22.429: g_object_get_is_valid_property: object class 'DinoPluginsRtpSink' has no property named 'paintable'
(dino:33709): rtp-WARNING **: 23:27:22.434: video_widget.vala:284: **NOT** display_device is_gtk4paintablesink
(dino:35288): rtp-WARNING **: 23:57:28.234: module.vala:75: No suitable encoder or decoder found for h264
```
I've tested without #1781 and have the same problem.
That last line about h264 does appear too in #1781 but I can see myself and the caller after compiling the PR. It seems the fallback isn't working, maybe because of https://gitlab.gnome.org/GNOME/gtk/-/issues/7671 as you said ?

Great stuff otherwise, wish you the best :) !

rmader:
Thanks for testing! So
> **NOT** display_device is_gtk4paintablesink

means your `gtk4paintablesink` does not have the required features - that should change once you update to Fedora 43.

>  It seems the fallback isn't working

Correct - will fix!

> maybe because of https://gitlab.gnome.org/GNOME/gtk/-/issues/7671 as you said

Nope, it's a mistake in this PR.

> No suitable encoder or decoder found for h264

Looks like your setup is missing elements for H264, so https://github.com/dino/dino/pull/1781 should not really change anything on your setup (unless maybe hardware VP8 en-/decoding, but according to your comment that doesn't seem to be the case, otherwise a bigger change is quality and CPU usage should be visible). See https://github.com/dino/dino/pull/1781#issuecomment-3514179255 how you can check - and https://rpmfusion.org/Howto/Multimedia for how to install the elements in question on Fedora ("Switch to full ffmpeg", "Install additional codec" and "Hardware codecs with AMD (mesa)").

With the later, you should see quite different behavior with https://github.com/dino/dino/pull/1781

rmader:
Rebased and fixed the fallback path (https://github.com/dino/dino/pull/1738#issuecomment-3516575072). With https://github.com/dino/dino/commit/1deaceeb1a61e4a2718f2ded375dbb966d39a341 the Flatpak should now have an up-to-date `gtk4paintablesink`.

rmader:
Added some more changes to reduce another copy and fixed the size of the own video on rotated cameras. Unfortunately there's one videoflip left as getting rid of that requires more intrusive changes. Overall this should be in a pretty good shape and a good drop-in replacement, improving performance quite a bit, especially in combination with Gstreamer 1.28. So I hope it can land soon.

Also includes https://github.com/dino/dino/pull/1825, which can hopefully land earlier.

rmader:
Added a couple of changes making this PR further complete:
1. In the `gtk4paintablesink` case we can now avoid all usages of `videoflip` or `videoconvert` (each possibly causing a buffer copy). This now also works for the flipped own video.
2. Fixed some glitches that are introduced or made more likely when using `gtk4paintablesink`

I'm not fully happy with all patches yet, thus turning this into a draft again until I can further clean them up.

FTR.: this still includes https://github.com/dino/dino/pull/1825 - would be happy if we could already land that, given that it's just a cleanup for EOL platforms.

PR #1742, opened 2025-07-23
Add libav x264 decoder, and modern VA API encoder and decoders

catap:


rmader:
Thanks for the PR! I just tested it and found one issue - `vah264dec` not working without `h264parse` - and one suggestion. Pushed that changes to https://github.com/rmader/dino/commits/pr-1742-fixups/, please feel free to squash into your commits.

With those I think this is good to go.

Semi-related: https://github.com/dino/dino/pull/1775

rmader:
@catap I looked into this a bit further and ended up opening https://github.com/dino/dino/pull/1781, mentioning you in the respective commits - hope that works for you?

catap:
@rmader I not sure that replacing "deprecated" decoder/encoder is good idea. As far as I recall modern is supported by only new enough hardware, and old hardware is supported only with old or deprecated driver.

rmader:
> As far as I recall modern is supported by only new enough hardware, and old hardware is supported only with old or deprecated driver.

Fortunately this is not the case. I just double-checked https://github.com/dino/dino/pull/1781 on my oldest test device with H264 en/decoding support - a Thinkpad X220 from 2011 (*1) - and hardware video decoding with `vah264dec` works perfectly fine. This is also in line with the Gstreamer [1.26 release notes](https://gstreamer.freedesktop.org/releases/1.26/):
> gstreamer-vaapi has been deprecated and is no longer actively maintained. Users who rely on gstreamer-vaapi are encouraged to migrate to the va plugin's elements at the earliest opportunity.

Note that on that device hardware *encoding* is blocked by Gstreamer as [we found it to be too buggy](https://gitlab.freedesktop.org/gstreamer/gstreamer/-/merge_requests/7170) when enabling hardware encoding by default in Gnome-Shell (for screen recording).

1. Sandybridge, the first Intel generation with GLES 3.0 support, which is now also the baseline for GL rendering in GTK4. I.e. older hardware will be very slow either way.

catap:
@rmader I fine with your way. Also, I see that you tested it much better than I'm :)

PR #1763, opened 2025-09-01
Set server_id for MAM messages from MUCs for accounts with no MAM sup…

### LINKED ISSUE #1758: Reply button is not active in MUCs for any messages except of those having origin-id
```
    public static string? get_reference_id(Message message) {
        if (message.edit_to != null) return message.edit_to;

        if (message.type_ == Message.Type.CHAT) {
            return message.stanza_id;
        } else {
            return message.server_id;
        }
    }
```
However, server_id is not being stored in database if it maches the message id.

bodqhrohro:
…port

Fixes #1758

### DIFF
--- a/libdino/src/service/message_processor.vala
+++ b/libdino/src/service/message_processor.vala
@@ -159,9 +159,9 @@ public class MessageProcessor : StreamInteractionModule, Object {
         Xmpp.MessageArchiveManagement.MessageFlag? mam_message_flag = Xmpp.MessageArchiveManagement.MessageFlag.get_flag(message);
         EntityInfo entity_info = stream_interactor.get_module(EntityInfo.IDENTITY);
         if (mam_message_flag != null && mam_message_flag.mam_id != null) {
-            bool server_does_mam = entity_info.has_feature_cached(account, account.bare_jid, Xmpp.MessageArchiveManagement.NS_URI);
-            bool muc_does_mam = message.type_ == Xmpp.MessageStanza.TYPE_GROUPCHAT && entity_info.has_feature_cached(account, message.from, Xmpp.MessageArchiveManagement.NS_URI);
-            if (server_does_mam || muc_does_mam) {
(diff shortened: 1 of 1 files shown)


libdino/src/service/message_processor.vala
```diff
@@ -160,7 +160,8 @@ public class MessageProcessor : StreamInteractionModule, Object {
         EntityInfo entity_info = stream_interactor.get_module(EntityInfo.IDENTITY);
         if (mam_message_flag != null && mam_message_flag.mam_id != null) {
             bool server_does_mam = entity_info.has_feature_cached(account, account.bare_jid, Xmpp.MessageArchiveManagement.NS_URI);
```
fiaxh, line 162:
```suggestion
            bool server_mam_acceptable = !message.type_.is_muc_semantic() && entity_info.has_feature_cached(account, account.bare_jid, Xmpp.MessageArchiveManagement.NS_URI);
```

fiaxh:
Thanks for catching that!

With the current PR code, MUC MAM messages would be accepted if the user's server supports MAM. However, we should accept MUC MAM messages if and only if the MUC server supports MAM.
Also, appart from groupchat messages, there are also groupchat PM messages which are not covered by the PR yet (they'll be covered by `message.type_.is_muc_semantic()`)

bodqhrohro:
Eh, only `Entities.Message` have this method, `Xmpp.MessageStanza` don't (and `new_message` doesn't have the type yet, it's assigned later). Ideas? Moving `determine_message_type` earlier? Introducing the method on `Xmpp.MessageStanza` too?

fiaxh:
Ah, true. The `MessageArchiveManagement.MessageFlag` contains the `sender_jid` of the JID the MAM message came from. We can use that one, then we don't have to derive that information with the help of the message type.

fiaxh:
Thanks!

PR #1772, opened 2025-09-25
Affect only active conversations with the default encryption setting

bodqhrohro:
Current behaviour is confusing for users because the whole roster becomes affected by the default encryption setting before they even have a chance to open preferences and disable OMEMO by default.

### DIFF
--- a/libdino/src/service/conversation_manager.vala
+++ b/libdino/src/service/conversation_manager.vala
@@ -48,13 +48,8 @@ public class ConversationManager : StreamInteractionModule, Object {
 
         // Create a new converation
         Conversation conversation = new Conversation(jid, account, type);
-        // Set encryption for conversation
-        if (type == Conversation.Type.CHAT ||
-                (type == Conversation.Type.GROUPCHAT && stream_interactor.get_module(MucManager.IDENTITY).is_private_room(account, jid))) {
-            conversation.encryption = Application.get_default().settings.get_default_encryption(account);
-        } else {
-            conversation.encryption = Encryption.NONE;
-        }
+
+        conversation.encryption = Encryption.NONE;
 
         add_conversation(conversation);
         conversation.persist(db);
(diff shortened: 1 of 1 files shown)


kousu:
I'm having trouble picturing what this is doing. Can you give some screenshots showing the problematic flow?

When Dino starts for the first time doesn't it start with no conversations open? After a user opens one or two and realizes OMEMO is on they can go find the setting in the prefs and turn it off, and then it will be fine for all future conversations. Will it not? Maybe I'm misunderstanding.

bodqhrohro:
I've described it in OP quite evidently. The setting was useless for *existing* contacts, because Dino creates an invisible conversation object for every contact regardless of them being visible in the chat list or not. So if they finally open a chat with some of existing contacts, they realize it still has OMEMO enabled by default, even though they disabled the setting *after* the contact was added and *before* the chat was opened (the conversation became "active").

Maybe it's non-intuitive for you that users can log into their existing accounts with Dino which already have some contacts in the roster?

kousu:
I didn't understand your original explanation. Now that I do I can reproduce it:

1. I created an account 'test3' on my server.
2. I used Gajim to **add it as a contact** before Dino ever saw it; I disabled the account in Gajim immediately.
3. I logged in to Dino

    image
4. I closed and reopened the preferences window in order to get access to the encryption settings
    image
5. I disabled OMEMO
    image

6. I opened a new chat

    image

    image

Here's the problem: OMEMO is enabled, check the bottom right:

image


I'm sorry I didn't understand the original way you phrased the problem, but also I don't appreciate getting snapped at.  We're both trying to make XMPP stronger for everyone, can we please focus on that? I saw this bug report not getting the attention it deserved and thought prompting you for more details would make it easier for the maintainers to help.

bodqhrohro:
> snapped at

Did I? o_O

Merely being a Slavic coconut.

kousu:
> > snapped at
> 
> Did I? o_O
> 
> Merely being a Slavic coconut.

Okau sorry ! I guess i misread! Text is hard to read tone in. 🫡

I hope this bug can get fixed. Maybe it's enough just to not initialize the setting until the conversation is opened?

bodqhrohro:
This is literally what this patch does.

kousu:
Ohhh i forgot this was a PR. I got flustered by missing social cues and lost track of what was going on.

mar-v-in:
This proposed change disables OMEMO if the default setting is set to disabled and I have an open chat with someone OMEMO encrypted, close it and then reopen it. While some people keep chats open all the time, others like to have a clean list of chats. For the latter that behavior is undesirable. It is meant to be that whatever you configure for a chat will be recovered the next time you open the chat after closing (which was before).

The underlying problem is that conversation objects are created for every contact (to be precise, it probably needs subscription) without need. Rather then fixing symptoms of this (which I would say this PR is about), it would be better to solve that underlying issue (and preserve the current behavior for the encryption setting)

PR #1773, opened 2025-10-05
Multiple messages select

bodqhrohro:


bodqhrohro:
Seems like I'm done finally, hehe.
Sorry for not marking as WIP.

kousu:
OMG I have been dying for this for AGES :star_struck: .  Dino is great but not being able to snip chatlogs to my journal except one-by-one has been a drag.

It took me a minute to find the UI for this, it wasn't intuitive at first but I think now I know it's there it's okay:

image

This is very intuitive. The whole message widget is clickable and the background changes colour to indicate :heart_eyes: 

image

When I copy it copies authors and timestamps, I love that:

```
nick@kousu.ca, [2026-03-07 07:05:41]
Cffff

nick@kousu.ca, [2026-03-07 07:07:41]
Ffggg

nick@kousu.ca, [2026-03-07 07:08:08]
Ddfff

nick@kousu.ca, [2026-03-07 07:08:10]
Fhdhduuf

nick@kousu.ca, [2026-03-07 08:16:20]
Fff
```

Files and Calls are rendered to plaintext in a very reasonable way :+1: 

```
nick@kousu.ca, [2026-02-21 00:21:42]
(Incoming call)
```

```
Master C, [2023-07-21 19:30:19]
(File someting_borrowed_932487fs982.mp3)
```




##  Signal

Signal-Desktop has "Select" in the conversation menu:

image

but I'd literally never noticed that before.  That's not, personally, an intuitive place for it. I always use the "Select" in the per-message menu:

image

image

Interestingly, the one thing Signal-Desktop won't do is copy selected messages, it only has "Forward" and "Delete". The Signal mobile app does allow copying, so this is just an oversight.

Also Signal doesn't include timestamps or authors in the copies!

But it does let me highlight multiple messages with just click-drag, no special mode needed, which I think is a bit easier for it since it's in Electron. I love how easy that is. That also doesn't include authors or timestamps though.

## Suggestions

If I may, I have some ideas for ways to make this even better:

- Maybe the plaintext should include "{author} <{JID}>" so that people can't forge identities in transcripts.
- in `master` there's now a dropdown menu on each MessageItem.  Can "Select" be added to that?

    image
- **Remove** the context menu on MessageItems (actually on the Label within each MessageWidget). Cut/Paste/Delete don't make any sense, and Copy and Select All are replaced by this feature, so now "Select All" confusingly means "select this message's text" while the user probably expects it to mean "select all messages"

    Screenshot From 2026-03-07 21-38-59

    Instead of that context menu, right-clicking the MessageWidget/its label could trigger the  per-message dropdown menu.

- The whole message widget is supposed to be clickable but it's got dead zones where the Label is, which is, unfortunately, the bulk of the widget:

    image

    I experimented and found simply setting `selectable=false` both removes the useless context menu and makes the widgets fully clickable, with no dead-zones

    https://github.com/dino/dino/blob/202c7ba8196f2cd51931028b5ed7f7b0b276ae4e/main/src/ui/conversation_content_view/message_widget.vala#L32

    Notice here that the mouse is now a pointer, not a cursor :
    image

- If the context menu is removed, Ctrl-C can (and probably should) be repurposed to immediately copy the current selection of messages.
- And the Delete key should trigger the new "Delete" option that's in `master`, but for all messages, but that can be a follow-up because now I'm asking for things beyond the scope of your initial PR here, heehee 😝
- There's occasionally a flicker where all checkboxes are checked just after I initiate Select mode. It's hard to reproduce. Maybe I'm seeing things 😵‍💫
- Can Shift-Click please select a range of messages? Otherwise it takes forever to export a conversation.
- Can click-drag select a range of messages?

    Right now click-drag during select mode will temporarily deselect a message if I hover over it, but it's a ghost: and once I glide off it it returns to the selected state and once I release the mouse it returns to the selected state. And it dragging will never select unselected messages.

    I would like:

    - clicking on an unselected message and then dragging to select _all touched messages_
    - clicking on an selected message and then dragging to **un**select _all touched messages_

    that way the user's intent is clear and there's no fiddly "aaah I went too far and undid two thirds of my selection FML" moments.
- Is the checkbox necessary? Other messengers are split on whether it's needed: Signal has it , WhatsApp simply indicates selected by changing the background.

    Is there an a11y issue there? Does the explicit checkbox help screenreaders?

- Add a 'Select All' button? Maybe that's a bad idea.


Finally:
- Is there any chance at all that selecting could work simply with the mouse?  That used to work with Pidgin way back in the day. Maybe once you open Select Mode, dragging over messages selects/deselects them? Or, my dream, trigger select mode simply by dragging over messages?

Thank you for tackling this. I also want to knock out the small quality-of-life nuisances that are total blockers preventing people from using XMPP.


Post-finally old man 😤 @ ☁️: 

- I don't love that messages can be selected individually. Personally I would rather only ever copy ranges. If I want to do editing I can do that in my notepad afterwards. But I know that's what other messengers do, and I think conforming to the common UI jargon is the most important thing for making XMPP attractive.

bodqhrohro:
> But it does let me highlight multiple messages with just click-drag

Selection mode can be activated by holding on an empty area of a message too.

I though on selection by dragging over multiple messages, but forgot before I considered the feature finished, and switched to other tasks.

> so that people can't forge identities in transcripts

I don't understand how really.

In PM the nickname is assigned per-JID. If the copying party wants to mess with something in the text, nothing stops them, it's just plaintext after all. This is why some modern IMs have native forwarding feature, which in some cases allows to verify that a certain account posted the messages and those who forwarded them don't mess on them. Much harder in federation, possibly ActivityPub handles it somehow, I'm not an expert on it totally as I don't use social networks on principle.

In MUCs, full JIDs might not even be seen. MUC settings might prohibit nickname takeovers per se.

> Can "Select" be added to that?

This menu looks like it was merely added to hide the dangerous item. Also, message selection is not a one-message action and it would be confusing and redundant this way.

> Cut/Paste/Delete don't make any sense

IDK where do they even come from, hehe. Possibly old upstream issue too?

> simply setting selectable=false

Thanks.

> There's occasionally a flicker 

Yup, because the width of messages changes and they have to be reflown. Dino surprisingly handles it better than I expected with no changes, but still the visual change is annoying, and I can't see an easy way to mitigate it. Maybe checkboxes can cover the message widget so no reflow is needed, I'll think on it.

> Is the checkbox necessary?

I considered it the primary selection way initially, especially given the dead areas, so the user knows for sure how to deselect a message at least. And it makes it clear that what got enabled, especially accidentally, is a selection mode. Background highlight is an additional visual hint, and might theoretically get broken by themes.

> Add a 'Select All' button? Maybe that's a bad idea.

Yeah, theoretically possible, but I see no practical use. I think there should be an explicit chat export feature, possible multiple chats at once.

> That used to work with Pidgin way back in the day

That generation of IMs printed messages just as a rich text log. Which made usual text operations easy on it, but also had a downside as messages were not even separate entities, just text rows, and thus operations on individual messages, even edit/delete, were quite hard to implement. Gajim before 1.4 had grown with lots of kludges for that which tracked positions of individual characters, add encryption/reading marks there and so. They broke often, they caused visual glitches, plugins broke them even more. Also this approach makes extremely hard to implement virtualized scrolling, and thus requires to keep the whole accumulated chat log in the widget, so Pidgin/Gajim inevitably leak while running in background. This is also why a separate per-day log viewer existed there (not that I find it really bad/useless though). Since 1.4, Gajim moved to per-message widgets too which broke selection for a while. Dino had this approach from the very start.

> Personally I would rather only ever copy range

Why, it makes sense to omit certain sensitive or out-of-context messages, for example.

kousu:
> > so that people can't forge identities in transcripts
> 
> I don't understand how really.
> 
> In PM the nickname is assigned per-JID. If the copying party wants to mess with something in the text, nothing stops them, it's just plaintext after all.

I realize I wasn't clear, maybe I shouldn't have used the word "forge" since it sounds more dramatic than I meant it to sound. Of course recipients can forge anything they want before sending them on and trying to stop that is not serving the user. If someone wants to make remixes or art or just lie with their own data they have the freedom to do so.

My concern is that people will troll each **me** by changing their display names. I have friends who set their names to the names of my parents or teachers to prank me (do you have friends who do that?). In the chat I can still sort of work out who is who but in the copied text if all you preserve is the display name then it makes their trolling easier.

I'm just suggesting that you include the JID in the copied text so it's clear who was who when looking back at transcripts long after the fact :sweat_smile: 

In a MUC with hidden JIDs it's of course impossible to do that (and therefore easy to troll) but that's a deeper issue in that XMPP has [other defenses](https://prosody.im/doc/modules/mod_muc#enforcing-reserved-nicknames) for. 

When JIDs are known I would like to preserve them in the transcripts.

> > Can "Select" be added to that?
> 
> This menu looks like it was merely added to hide the dangerous item. Also, message selection is not a one-message action and it would be confusing and redundant this way.

The only way to get the transcript with authors and timestamps included by your formatting is through your Copy button, which **motivates** me to select a single message. Previously, without your work, I could _only_ select messages one at a time, by highlighting the text in the `Label`, and that loses the author and timestamp and it's error-prone and time consuming because I need to make sure to put the cursor on the very start or very end and drag. It's much much nicer to be able to select each message as a unit, even, or especially, when it's just one.

By the way, "Delete Message" is not hiding, it's a new feature (it sends a [retraction](https://xmpp.org/extensions/xep-0424.html)!), and that menu also (in the right context) has "Edit"

image


It's basically the same context menu that Signal and WhatsApp have e.g. here's Signal:

image

I think it would be cool if Dino took a cue from other messengers people are used to and added "Copy"/"Select" (i.e. your work) and "Info" (to show sender and their OMEMO key, like Conversations can do) to this menu.

> > Cut/Paste/Delete don't make any sense
> 
> IDK where do they even come from, hehe. Possibly old upstream issue too?

They're from the same `Label` as before

https://github.com/dino/dino/blob/202c7ba8196f2cd51931028b5ed7f7b0b276ae4e/main/src/ui/conversation_content_view/message_widget.vala#L32

and setting `selectable=False` disables them too so actually that's easy to solve :)

 
> > There's occasionally a flicker
> 
> Yup, because the width of messages changes and they have to be reflown. Dino surprisingly handles it better than I expected with no changes, but still the visual change is annoying, and I can't see an easy way to mitigate it. Maybe checkboxes can cover the message widget so no reflow is needed, I'll think on it.

Maybe there's a way in Vala to say "don't render anything until I tell you it's okay" and then loop over the messages and add all the checkboxes and then unlock rendering? In fact **it is**, by calling [freeze_notify()](https://valadoc.org/gobject-2.0/GLib.Object.freeze_notify.html). I'm not 100% sure what you would freeze, the Conversation widget as a whole maybe?

On the other hand, I just checked what Signal-Desktop does with their checkboxes. It looks like they have the checkbox element permanently in the UI but sitting at `opacity: 0; z-index: -1` underneath the messages. When shown there's a css transition that reveals the checkbox and slides the texts over, so instead of a flicker there's an animation. That should be directly portable to Dino since GTK uses CSS too, if you like the idea.

https://github.com/user-attachments/assets/5fe49828-9f6c-489a-9246-52af62a5ebc7

> > Is the checkbox necessary?
> 
> I considered it the primary selection way initially, especially given the dead areas, so the user knows for sure how to deselect a message at least. And it makes it clear that what got enabled, especially accidentally, is a selection mode. Background highlight is an additional visual hint, and might theoretically get broken by themes.

Themes are a very good point. The other messengers I'm comparing with fix their themes in place and don't let the OS touch them but Dino is more open and needs to be more tolerant.

I think if you make a slide in animation to avoid the flicker then the checkbox is good :+1: 

> > That used to work with Pidgin way back in the day
> 
> [...] Which made usual text operations easy on it, but also had a downside as messages were not even separate entities, just text rows, and thus operations on individual messages, even edit/delete, were quite hard to implement.  [...] They broke often, they caused visual glitches, plugins broke them even more. [...]

Yes Dino is much nicer and more stable than any of that!  It's just this one specific use case of saving chat transcripts is now much worse :cry: 

> > Personally I would rather only ever copy range
> 
> Why, it makes sense to omit certain sensitive or out-of-context messages, for example.

> > Add a 'Select All' button? Maybe that's a bad idea.
> 
> Yeah, theoretically possible, but I see no practical use. I think there should be an explicit chat export feature, possible multiple chats at once.

Usually I don't want to export an entire transcript, I want to export either a _single_ message or a _single_ _conversation topic_ within it, i.e. the last message or the last hours/week.  How many messages do you think that is? Personally, if I've decided keep a snippet for my journal or notes for my work project it's a minimum 100 messages, and easily 500 especially in an informal chatroom setting with a lot of slang and reactions and 'lol's. I neither want the complete conversation log back nor do I want to click hundreds of checkboxes.  Meanwhile, in the Single message case, I want the copy to be one-click so I would rather click zero checkboxes.

I realized I contradicted myself. What I meant by "I don't like selecting individual messages" was "I don't like checkboxes being the only way to select" because it's slow. I would rather optimize for selecting ranges and have selecting a single message be the base case of that.

I do light editing on my transcripts, but I much prefer to make those edits in a text editor.  Without being able to select/deselect subranges in Dino I have to make those edits by choosing message by message. As a user that feels stressful because it's one of those situations where I am creating a lot of state that could be wiped out at a moment's notice (e.g. if my computer or Dino crashes or I slip and press the Esc key.

Editing in a text editor lets me easily highlight and delete subranges, undo mistakes, and save my work. Being able to save alleviates the stress of making complicated edits.

A full chat export that I edited in a text editor would also work (especially if it shared the same copying code as you) but I would just find myself always deleting the majority of its content, everything up to the start of my most recent conversation.

I hope that makes sense and that you can see where I'm coming from :turtle: 

---

Anyway don't get me wrong just because I have suggestions. I very much like that this feature exists. I really hope it goes in in some form. Even as-is it would be a great improvement to my Dino experience :purple_heart:

bodqhrohro:
> (do you have friends who do that?)

Normies who use XMPP? lol.

I did troll like that a lot back then myself though.

> By the way, "Delete Message" is not hiding

Nah, I mean, what even is the point to hide it under a dropdown when there is a lot of room for buttons still? They should rather auto-collapse if they don't fit, I think.

> to show sender

I rather think sometimes about a feature to display the sending resource right at the messages in the UI. This information is public and evident, and quite useful (especially for me to recall what client did I even send something from lol). Same for Matrix clients, as events have the certain sending device in metadata, but this information is displayed nowhere but the source JSON.

> "don't render anything until I tell you it's okay"

I don't think it makes sense as the way event loop in GTK works eliminates possible redraws/reflows while an event handler is executed until it yields control back. Same as in browser JS which is single-threaded to the core. The problem is the reflow itself, and keeping track of relative message positions.

> slides the texts over

Still not clear how this is done from your video, I think it's implemented by shifting the whole message widget but the checkbox rightwards in a relative way, so reflow is avoided completely. This might be achieved by a gap at the right side, which is not clear as your video is cut there, which leaves room for that. Quite irrelevant here, cause current Dino UI lacks such huge gaps, thus shifting like that would make message text partially cut. Hovering the checkboxes over avatars would be rather indestructive for that matter though. I also could put them in message corners.

> it's a minimum 100 messages, and easily 500 especially 

I think it makes sense to add a "select last N messages" widget to the selection bar. The bar is huge and empty anyway. Also, I recall no IMs which have it, so it would be revolutionary.

kousu:
I just used this for the first time for real :) I built your branch and was able to archive an interesting debate I was part of to my journal.

PR #1780, opened 2025-10-24
File link copy button

bodqhrohro:


fiaxh:
With Dino we try to build a client that is easily understandable and I feel that this functionality would not align with that goal.

- To understand this function, a user first has to understand that files in XMPP are sometimes (but not always!) transferred via http. That's a technical detail on how the transfer works and should not be exposed to the user.
- Files are not always transferred via http, they can also be transferred via Jingle, then a file transfer does not have a http link. Thus, the UI can't be consistent for all file transfers.
- Most direct chats are OMEMO encrypted (it's the default). For OMEMO encrypted file transfers, there isn't even a http url, that could be opened in a browser. The user would get an "aesgcm" link and then somehow decrypt the file on their own.

PR #1775, opened 2025-10-12
rtp: device: Ensure even heights for rescaled streams

rmader:
Odd heights, while generally handled by Gstreamer, can cause issues in
various places - most importantly hardware video en-/decoders and
GL/VK drivers - as in most cases 2x0 or 2x2 subsampled intermediate
pixel formats are used (such as I420, YUYV, NV12 etc.).

---

cc @mar-v-in - this may help with issues you've been observing around hardware en-/decoders.

Related:
 - https://github.com/dino/dino/pull/1738

PR #1781, opened 2025-10-24
Various improvement and cleanups for video codecs/calls

rmader:
Inspired by https://github.com/dino/dino/pull/1742 this is a collection of changes that together improve video call quality quite a bit for me - see individual commits.

---

Tested on:
 - ThinkPad P14s Gen 5 AMD, Fedora 43, VA-API encoding+decoding
 - ThinkPad X220 (Intel Sandybridge, from 2011, *1), Fedora 43, VA-API decoding
 - Desktop AMD system, Fedora 43, VA-API encoding+decoding
 - PinePhone Pro, PineBook Pro, Librem5, postmarketOS, V4L2 stateless decoding
 - Fairphone 5, Pixel 3a, OnePlus6, postmarketOS, software encoding and decoding (V4L2 stateful decoding unstable) 

cc @mar-v-in 

Includes:
 - https://github.com/dino/dino/pull/1775

Related:
 - https://github.com/dino/dino/pull/1742
 - https://github.com/dino/dino/pull/1738

---

Notes:
1. On old Intel systems using the derpecated `intel-vaapi-driver` [video encoding is blocklisted](https://gitlab.freedesktop.org/gstreamer/gstreamer/-/merge_requests/7170) by Gstreamer for stability reasons. In can be forced-enabled by running with `GST_VA_ALL_DRIVERS=1`, however a quick test on a ThinkPad X220 showed clearly visible glitches/artifacts appearing. *Decoding* with VA-API works just fine though (and is enabled by default by this PR).

### DIFF
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -11,7 +11,7 @@ option('plugin-notification-sound', type: 'feature', value: 'disabled', descript
 
 option('plugin-rtp-h264', type: 'feature', value: 'disabled', description: 'H264 codec')
 option('plugin-rtp-msdk', type: 'feature', value: 'disabled', description: 'Intel MediaSDK')
-option('plugin-rtp-vaapi', type: 'feature', value: 'disabled', description: 'Video Acceleration API')
+option('plugin-rtp-vaapi', type: 'feature', value: 'auto', description: 'Video Acceleration API')
 option('plugin-rtp-vp9', type: 'feature', value: 'disabled', description: 'VP9 codec')
 option('plugin-rtp-v4l2', type: 'feature', value: 'disabled', description: 'V4L2 API (stateful)')
 option('plugin-rtp-v4l2-sl', type: 'feature', value: 'disabled', description: 'V4L2 API (stateless)')
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -13,4 +13,6 @@ option('plugin-rtp-h264', type: 'feature', value: 'disabled', description: 'H264
 option('plugin-rtp-msdk', type: 'feature', value: 'disabled', description: 'Intel MediaSDK')
 option('plugin-rtp-vaapi', type: 'feature', value: 'disabled', description: 'Video Acceleration API')
 option('plugin-rtp-vp9', type: 'feature', value: 'disabled', description: 'VP9 codec')
+option('plugin-rtp-v4l2', type: 'feature', value: 'disabled', description: 'V4L2 API (stateful)')
+option('plugin-rtp-v4l2-sl', type: 'feature', value: 'disabled', description: 'V4L2 API (stateless)')
 option('plugin-rtp-webrtc-audio-processing', value: 'auto', type: 'feature', description: 'Voice preprocessing')
--- a/plugins/rtp/meson.build
+++ b/plugins/rtp/meson.build
@@ -79,6 +79,16 @@ if get_option('plugin-rtp-vp9').allowed()
     vala_args += ['-D', 'ENABLE_VP9']
 endif
 
+option_v4l2_enabled = get_option('plugin-rtp-v4l2').enabled()
+if option_v4l2_enabled
+    vala_args += ['-D', 'ENABLE_V4L2']
+endif
+
+option_v4l2_sl_enabled = get_option('plugin-rtp-v4l2-sl').enabled()
+if option_v4l2_sl_enabled
+    vala_args += ['-D', 'ENABLE_V4L2SL']
+endif
+
 lib_rtp = shared_library('rtp', sources, name_prefix: '', c_args: c_args, vala_args: vala_args, include_directories: include_directories('src'), dependencies: dependencies, install: true, install_dir: get_option('libdir') / get_option('plugindir'), install_rpath: default_install_rpath)
 dep_rtp = declare_dependency(link_with: lib_rtp, include_directories: include_directories('.'))
 summary('Voice/video calls (rtp)', dep_rtp.found(), bool_yn: true, section: 'Plugins')
@@ -88,5 +98,7 @@ if dep_rtp.found()
     summary('VP9 codec', get_option('plugin-rtp-vp9').allowed(), section: 'RTP configuration')
     summary('Intel MediaSDK', get_option('plugin-rtp-msdk').allowed(), section: 'RTP configuration')
     summary('Video Acceleration API', get_option('plugin-rtp-vaapi').allowed(), section: 'RTP configuration')
+    summary('V4L2 API (stateful)', option_v4l2_enabled, section: 'RTP configuration')
+    summary('V4L2 API (stateless)', option_v4l2_sl_enabled, section: 'RTP configuration')
     summary('Voice preprocessing', dep_webrtc_audio_processing.found(), section: 'RTP configuration')
 endif
--- a/plugins/rtp/meson.build
+++ b/plugins/rtp/meson.build
@@ -72,7 +72,8 @@ endif
 if get_option('plugin-rtp-msdk').allowed()
     vala_args += ['-D', 'ENABLE_MSDK']
 endif
-if get_option('plugin-rtp-vaapi').allowed()
+option_vaapi_enabled = get_option('plugin-rtp-vaapi').enabled() or (get_option('plugin-rtp-vaapi').allowed() and dep_gstreamer_rtp.version().version_compare('>=1.28'))
+if option_vaapi_enabled
     vala_args += ['-D', 'ENABLE_VAAPI']
 endif
 if get_option('plugin-rtp-vp9').allowed()
@@ -97,7 +98,7 @@ if dep_rtp.found()
     summary('H264 codec', get_option('plugin-rtp-h264').allowed(), section: 'RTP configuration')
     summary('VP9 codec', get_option('plugin-rtp-vp9').allowed(), section: 'RTP configuration')
     summary('Intel MediaSDK', get_option('plugin-rtp-msdk').allowed(), section: 'RTP configuration')
-    summary('Video Acceleration API', get_option('plugin-rtp-vaapi').allowed(), section: 'RTP configuration')
+    summary('Video Acceleration API', option_vaapi_enabled, section: 'RTP configuration')
     summary('V4L2 API (stateful)', option_v4l2_enabled, section: 'RTP configuration')
     summary('V4L2 API (stateless)', option_v4l2_sl_enabled, section: 'RTP configuration')
     summary('Voice preprocessing', dep_webrtc_audio_processing.found(), section: 'RTP configuration')
--- a/plugins/rtp/src/codec_util.vala
+++ b/plugins/rtp/src/codec_util.vala
@@ -102,7 +102,8 @@ public class Dino.Plugins.Rtp.CodecUtil {
                         "msdkh264enc",
 #endif
 #if ENABLE_VAAPI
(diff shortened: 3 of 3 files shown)


rmader:
Latest push added V4L2 decoding support, both stateful and stateless, with the later being enabled by default. Paired with https://github.com/dino/dino/pull/1738 this allowed me to have pretty smooth video on low-end devices like the Librem5(*1) and the PinePhone Pro (only receiving video - the encoding part is still too heavy for them).

@mar-v-in I'm pretty happy with the series now - would be super happy if you find time to review/test

1.: With latest Mesa and GLES renderer forced-enabled with `MESA_GLES_VERSION_OVERRIDE=3.0` - should be the default in the near future.

eerielili:
> Latest push added V4L2 decoding support, both stateful and stateless, with the later being enabled by default. Paired with https://github.com/dino/dino/pull/1738 this allowed me to have pretty smooth video on low-end devices like the Librem5(*1) and the PinePhone Pro (only receiving video - the encoding part is still too heavy for them).

@rmader It might interest me to test it on a feeble AMD CPU laptop, might be able to this weekend :)

rmader:
@eerielili that would be great, thanks!

FWIW: the patches here were also successfully live-demoed with some linux-mobile devices on a recent conference, see https://gitlab.postmarketos.org/postmarketOS/pmaports/-/merge_requests/7272#note_511390 / https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/92360

eerielili:
@rmader Great! I don't have any Linux phone on hand unfortunately.
I have some good news though, your PR did compile OK on the laptop ([AMD E1-6010 APU with AMD Radeon R2 Graphics](https://www.amd.com/en/support/downloads/drivers.html/processors/e-series/e1-series-apu-for-laptops/e1-6010-with-radeon-r2-graphics.html#amd_support_product_spec)) and works.

Did some tests with my laptop<=>phone and it seems your branch is a little easier on this CPU (we're talking 2-3% checking htop) with VP8, quality stayed the same. I feel the received video is a bit less choppier on the laptop.

**EDIT**: Compiling with #1738 on top of it rn, will come back with feedback

rmader:
@eerielili thanks for testing! A few notes:
1. Just in case you haven't done that already: When you do CPU comparisons you'll want to ensure that the "Performance Mode" or "Power Mode" of your device is set to "Performance" so the CPU does not get clocked down. Otherwise the cpu usage reported by `htop` can be misleading - i.e. a higher CPU usage with lowest clock rate is actually better than a lower CPU usage with highest clock rate. You can double-check that e.g. with `powertop` -> "Frequency stats".
1. You can check whether your GPU supports HW en-/decoding either with `vainfo` (for H264, you'd need `VAProfileH264ConstrainedBaseline:	VAEntrypointVLD` for decoding and `VAProfileH264ConstrainedBaseline:	VAEntrypointEncSlice` for encoding IIRC - and for VP8 similar entries starting with `VAProfileVP8`) or by checking if the Gstreamer elements are present (`gst-inspect-1.0 vah264enc`, `gst-inspect-1.0 vah264dec`, `gst-inspect-1.0 vavp8enc`, `gst-inspect-1.0 vavp8dec`) With those available the expectation would be the quality should increase and CPU usage should decrease noticable. If those are *not* available, especially the h264 ones, then most likely your setup is missing some packages.
3. You can also check for h264 software en-/decoders with `gst-inspect-1.0 x264enc` and `gst-inspect-1.0 avdec_h264` - if those (or the `vah264*` elements) are available, H264 should be selected by default - and should result in better quality at similar CPU and bandwidth usage.
4. For https://github.com/dino/dino/pull/1738 you'll need to ensure to have a very recent `gtk4paintablesink` (>= 0.14.3) with GTK 4.20 features enabled. You can check with `gst-inspect-1.0 gtk4paintablesink` and check if the "Pad Templates" nemtion YUV formats like `YUY2`. With that you should see reduced CPU and GPU usage - without, nothing should change.

> it seems your branch is a little easier on this CPU (we're talking 2-3% checking htop) with VP8

That sounds to me like the VP8 CPU en-/decoders are still being used - in which case we wouldn't really expect any change with this branch. The difference should be that H264 is used when possible (better quality) - and hw en-/decoding, in which case the CPU usage should change considerably.

eerielili:
@rmader Oh my, thanks for all these directions !!
1. I was on `schedutil`, switched to `performance` for the tests
2. Seems my integrated GPU doesn't want to play with h264, vainfo output below:
```
vainfo: VA-API version: 1.22 (libva 2.22.0)
vainfo: Driver version: Mesa Gallium driver 25.2.6 for KABINI (radeonsi, , ACO, DRM 2.50, 6.14.0-63.fc42.x86_64)
vainfo: Supported profile and entrypoints
      VAProfileMPEG2Simple            :	VAEntrypointVLD
      VAProfileMPEG2Main              :	VAEntrypointVLD
      VAProfileNone                   :	VAEntrypointVideoProc
```
Probably missing packages on my end, I'll double check https://rpmfusion.org/Howto/Multimedia tomorrow or this weekend.
3. I do have some result for `x264enc` but not `avdec_h264`. I'll see if there's other missing packages.
4. I have good news for #1738! Update to Fedora 43 and in particular installation of package `gstreamer1-plugin-gtk4` solved the issue about not seeing anything in call window ! The `Pad templates` section of `gst-inspect` does indeed list YUY2.
And as you said, same performance with VP8 as normal Dino, nothing broken so far. I haven't tested audio still though.

**EDIT FOR POINT 2**: 
```bash
sudo dnf swap mesa-va-drivers mesa-va-drivers-freeworld
sudo dnf swap mesa-vdpau-drivers mesa-vdpau-drivers-freeworld
```
did the trick, `vainfo` now does show the 
```
VAProfileH264ConstrainedBaseline:	VAEntrypointVLD
VAProfileH264ConstrainedBaseline:	VAEntrypointEncSlice
```

rmader:
~~TODO: double-check that we set bitrates as intended, see https://gitlab.freedesktop.org/gstreamer/gst-plugins-rs/-/blob/5ccafb1405009ec98fc6951a04efb7ad9eb919c4/net/webrtc/src/webrtcsink/imp.rs#L1160-1174~~

Done, looks all correct.

mar-v-in:
With #1775 merged (which will be included in 0.5.1), I don't think anything else of this would still land in 0.5. Especially since we don't want to change compile-time defaults or introduce new compile-time options with patch releases.

Beside that, here's a few comments on this:
- I don't see a reason to fully drop support VP9. I also don't support the notion of VP9 not being used by most other apps in the ecosystem. Some popular video conferencing solutions in fact do use VP9.
- We currently use variable bitrate encoding, but we should definitely change to constant bitrate, as libwebrtc's REMB implementation can't handle vbr and will potentially instruct us to send lower bitrates than actually supported.
- Regarding h264 vs VP8: In my experience, h264 software implementations are usually less efficient (higher CPU usage, larger latency) than VP8 software implementations, at least on my test devices. I don't know if universal data for this is available, but quick searches suggest that many videoconferencing systems prefer. That's why we had removed the h264 software codecs.

For 0.6 version, we'd want to
- change to cbr
- drop msdk
- make all remaining options included at compile time (= get rid of the compile time options)
- allow users to disable/enable hardware acceleration (where we would enable it by default based on a GPU allowlist / blocklist and gstreamer version and users can override)
- allow the use of VP9 by default if preferred by the remote end, but never prefer it ourselves
- rank h264 hardware > vp8, but vp8 software > h264 software

rmader:
Just a simple rebase.

@mar-v-in sounds all pretty good to me. I slightly disagree on the takes on VP9 and H264 but would need to do further testing to make a stronger case - specifically I'd need to also check
1. how `openh264` compares to `avdec_h264`+`x264enc` as the former is installed by default on distros like Fedora in order to work around licensing issues.
2. whether VP9 currently works at all (because of the missing `vp9parse`) and how it compares to VP8/h264 for live-streams. AFAIK it always suffered from very slow encoding speeds.

I definitely like the idea of having a user setting for hardware en-/decoding - that's what we did in Snapshot for 49 and I currently plan to change it from a simple on/off, off being the default, to auto/on/off where auto is the default and enables "known good" cases.

mar-v-in:
From what I gathered, most commercial grade video conferencing solutions (Google Meet, Zoom, Jitsi Meet, Microsoft Teams) prefer AV1, VP9 or H265 if available, and fallback to VP8 or H264 if needed. Fallback might be due to limitations in hardware acceleration, available codecs (in browsers) or limitations in compute ressources.

mar-v-in:
Regarding VP9 / `vp9parse`: I can run `vp9dec` and `vaapivp9dec` without `vp9parse` in a simple pipeline without issues, however `vavp9dec` seems to require a `vp9parse`. So in other words, before adding `vavp9dec` (this PR) it worked without `vp9parse`, now it would be needed.

My tests:
| | | |
|-|-|-|
| Encoding (keep running while testing the others) | | `gst-launch-1.0 autovideosrc ! videoconvert ! vp9enc end-usage=cbr error-resilient=1 cpu-used=16 deadline=1 lag-in-frames=0 ! rtpvp9pay ! udpsink host=127.0.0.1 port=13371` |
| Decoding with `vp9enc` and no `vp9parse` | works | `gst-launch-1.0 udpsrc address=127.0.0.1 port=13371 caps="application/x-rtp" ! rtpvp9depay ! vp9dec ! autovideosink` |
| Decoding with `vaapivp9enc` and no `vp9parse` | works | `gst-launch-1.0 udpsrc address=127.0.0.1 port=13371 caps="application/x-rtp" ! rtpvp9depay ! vaapivp9dec ! autovideosink` |
| Decoding with `vavp9enc` and no `vp9parse` | fails | `gst-launch-1.0 udpsrc address=127.0.0.1 port=13371 caps="application/x-rtp" ! rtpvp9depay ! vavp9enc ! autovideosink` |
| Decoding with `vavp9enc` and `vp9parse` | works | `gst-launch-1.0 udpsrc address=127.0.0.1 port=13371 caps="application/x-rtp" ! rtpvp9depay ! vp9parse ! vavp9enc ! autovideosink` |

mar-v-in:
I also ran some tests with openh264 and found it indeed much faster than `x264enc` and `avdec_h264`. `openh264enc` is (on my system) about the same cpu use as `vp8enc` (and about half of `x264enc`), `openh264dec` takes about twice as much cpu usage as `vp8dec` (which is still 30% less than `avdec_h264`).

Still though, `vp8enc`+`vp8dec` seems to outperform `openh264enc`+`openh264dec` (although video quality is obviously not directly comparable). Notably though, `vah264enc`+`vah264dec` actually performs better on my system than `vavp8enc`+`vavp8dec` (measuring only the CPU use, not GPU use). So this would reconfirm that on systems where we can make use of hardware acceleration, H264 should be preferred over VP8. And of course if VP9, H265 or AV1 are supported for hardware acceleration, that would likely be even better.

catap:
Just fiendly ping here.

rmader:
Thanks for the reminder. I rebased and slimmed the PR down to only modernizing the hardware en-/decoding setup, keeping defaults as is apart from enabling VA-API by default. I dropped all changes related to software en-/decoding and changes to the used codecs - those can be discussed in follow-ups.

@mar-v-in hope this looks more acceptable?

catap:
FWIW I do have working video calls between Dino and Monal, and I run Dino at OpenBSD.

But I run a branch with a few my additional bugfixes.

PR #1786, opened 2025-11-01
Unlink Overlay After It's Closed

christf:
This is a rebase of one of the commits in #1486

Previously the Overlay was a stack variable, and the only reason it showed up was because it was leaked in the lambdas.

But making the references weak just exposes that this currently doesn't have an obvious owner. So I made the view controller the owner, and then in close I just clear that reference out so it can be reaped.

But, I can't free it out while in a handler on its own signal or it crashes, so I have to wait for the next tick first.

### DIFF
--- a/main/src/ui/conversation_view_controller.vala
+++ b/main/src/ui/conversation_view_controller.vala
@@ -12,6 +12,7 @@ public class ConversationViewController : Object {
     private MainWindow main_window;
     private ConversationView view;
     private Widget? overlay_dialog;
+    private FileSendOverlay overlay;
     public SearchMenuEntry search_menu_entry = new SearchMenuEntry();
     public ListView list_view = new ListView(null, null);
     private DropTarget drop_event_controller = new DropTarget(typeof(File), DragAction.COPY );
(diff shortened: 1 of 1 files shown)


mar-v-in:
The commit description seems to not make a lot of sense. 

The overlay is displayed because its widget is added to the widget tree. A widget in the widget tree has a reference from its parent widget. Vala uses reference-counting, there is not a single owner of an object and there is no need to have one. Instead, objects exist as long as someone else holds a reference, and that can well be through the widget tree (that's actually how most widgets in the widget tree are kept alive).

The fact that you can't remove the reference in the close handler points to another issue: whoever invokes the event handler should hold a local reference to the object, so removing an unrelated reference from there shouldn't cause issues.

christf:
ok... at the same time, memory is leaking and it would be great if we could stop it.
I am running dino with the patch and it seems to be better.
if it s the wrong patch, that is fine. I just rebased this because the original pr has not received discussion.

fiaxh:
Adding to mar-v-in comment: In my tests the patch from this PR did not fix the memory leak when opening the file send dialog.

mar-v-in wrote a patch that fixes the issue for me, and I just pushed it to master in 68b07ef13c857b07d32c2fd8da08a0660f961a0a. If you want to give it a try, you could check whether it also solves the issue in your tests.

We'll also include it in the next patch release.

PR #1788, opened 2025-11-04
Make conversation selector sidebar resizable

bodqhrohro:


kousu:
Interesting! 

image

This breaks the auto-hide assumptions of the conversation list so I wonder if some other parts of the code need to be adjusted to compensate.

It's a little weird that it's possible now to resize the window to *only* show the conversation list, but then if the window is shrunk just a little bit it suddenly jumps to showing only the conversation.

image

image

.

I'm not really sure what should happen in that case :thinking: 

## Signal-Desktop

Signal-Desktop imposes a minimum and maximum on the width:

image


image

Below the min it jumps to only showing icons:

image

For what it's worth, Signal-Desktop also allows this degenerate case:

image

## Account Prefs bug

For some reason the back button in the account preferences dialog isn't working on this branch

image

I don't see why this branch would have done anything to that button. I tried to merge with master in case it's an unrelated bug but there's conflicts in main/data/main_window.ui that I couldn't figure out how to resolve.

bodqhrohro:
> auto-hide assumptions of the conversation list 

WDYM?

> if the window is shrunk just a little bit it suddenly jumps to showing only the conversation

Dino is adaptive and was intentionally migrated to an Adwaita widget which enables view switching on mobile screens.

IDK if it's worth to make different behaviour on desktops.

> Below the min it jumps to only showing icons:

Might make sense. There's a minimum width rule set in the UI template, IDK for what.

> I don't see why this branch would have done anything to that button

An upstream bug which probably got fixed after the PR was made.

PR #1789, opened 2025-11-06
Show Conversation when it is selected

earboxer:
In mobile form-factors, [Adw.NavigationSplitView](https://gnome.pages.gitlab.gnome.org/libadwaita/doc/main/class.NavigationSplitView.html) needs to be told to show_content for the content to be shown.

Previously, this was only being done when you clicked an item from the selector in the sidebar. This PR adds that behavior to more locations in the code.

Previously, this was annoying, but I would use my finger to move to the content after adding a conversation. Since a recent commit, (375eddb368212aa63769b4b3fed4c55) it's no longer possible to move to the content using your finger when the sidebar is shown. (Hence me making this PR, the workaround is broken, so I have implemented the actual fix). This should be merged before the next release or it would be a serious regression in functionality for mobile users who have large contact lists.

- **Show content when you search**
	* when you press ctrl f, flip to the content pane so you know you did something
- **navigation_split_view: show content when a conversation is selected**
  * When you start a conversation (clicking a URI, or through the add menu) flip to the content pane so you can converse

### DIFF
--- a/main/src/ui/main_window_controller.vala
+++ b/main/src/ui/main_window_controller.vala
@@ -131,6 +131,7 @@ public class MainWindowController : Object {
         if (do_reset_search) {
             reset_search_entry();
         }
+        window.navigation_split_view.show_content = true;
     }
 
     private void check_unset_conversation() {


fiaxh:
Thanks for your PR.

I would like to avoid putting this very specific logic into the application class. You could put it into the MainWindowController.select_conversation(), which is called in all cases that you covered in the application class, and covers some additional cases.
Your change in main_window.vala would still be needed, though.

earboxer:
@fiaxh Thank you for the review! I made the change, and it still works, as predicted.

PR #1790, opened 2025-11-08
Set message menu box to be not visible by default (fixes #1745)

### LINKED ISSUE #1745: strange box shows up on devuan
Image

when i run dino in the current conversation from the top i see such weird box.

i see it in devuan daedalus.

perhaps it shows up because before it i don't see anything, because it is a new installation and it shows only part of history.

eerielili:
This closes #1745.
When Dino is opened and a conversation selected an empty message menu box can be seen like shown in the video [here](https://github.com/dino/dino/issues/1745#issuecomment-3506148273).
This can also happen when first setting up an account and receiving or sending an invite request without sending a message: a conversation pane will open and the ghost menu box is visible, tucked just behind the input text box:
Image

It's happening because the state of `message_menu_box.visible` is true by default, so it can appear on Dino startup / first convo in a not initialized form (no reply or message correction or emoji icon, etc).
Setting `message_menu_box.visible = false` at the end of the constructor fixes the issue.

### DIFF
--- a/main/data/conversation_content_view/view.ui
+++ b/main/data/conversation_content_view/view.ui
@@ -26,6 +26,7 @@
                                                 <property name="margin-end">10</property>
                                                 <property name="halign">end</property>
                                                 <property name="valign">start</property>
+                                                <property name="visible">False</property>
                                                 <style>
                                                     <class name="card"/>
                                                     <class name="toolbar"/>


fiaxh:
Thanks for your PR!

The ConversationView is created from a .ui file, [this one](https://github.com/dino/dino/blob/master/main/data/conversation_content_view/view.ui#L25), thus it would be good to set default values of properties in there.

eerielili:
@fiaxh I agree, the default makes more sense via properties in .ui file.
I'll work on that in a few minutes.

PR #1791, opened 2025-11-09
Notes to myself

bodqhrohro:


maikelthedev:
This is a nice one and I pretty much understood what it did by just looking at the files changes, the question is, why the need to look at the files changed? Wouldn't it be better to add a simple descriptoin to the PR itself so that the lead maintainer / decision makers have to **invest less time** understanding what this is or users reading pending PRs and upvoting those the find are more of a priority vs those they don't?

fiaxh:
Thanks for your work. I like the idea in general, but I think that this feature would not be easily discoverable if the user has to go to the accounts preferences first. Other messengers put a "Note to Self" entry into the contact list, which would be easier discoverable. I'll close this PR in favour of #1840, which implements "Note to Self" that way.

PR #1795, opened 2025-11-18
Use correct plugin suffix on macOS

hutchisr:
Uses .dylib suffix when building on macOS

### DIFF
--- a/libdino/meson.build
+++ b/libdino/meson.build
@@ -84,9 +84,11 @@ sources = files(
     'src/util/weak_timeout.vala',
 )
 sources += [version_vala]
+plugin_suffix = host_machine.system() == 'darwin' ? '.dylib' : '.so'
 c_args = [
     '-DDINO_SYSTEM_LIBDIR_NAME="@0@"'.format(get_option('prefix') / get_option('libdir')),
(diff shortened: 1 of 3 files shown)


mar-v-in:
GModule expects modules to end in `.so` on macOS, because that's how autotools builds them. While we can change the behavior of Dino on macOS as you propose, it's equally valid to adjust the build system to emit `.so` files on macOS for compatiblity. Given that macOS support has not been official yet, we did not make any such adjustment so far.

hutchisr:
I just thought I'd put this here to show what I did to get it working. It does seem that SUFFIX is deprecated though https://valadoc.org/gmodule-2.0/GLib.Module.SUFFIX.html

mar-v-in:
Indeed, it is deprecated precisely due to this complexity of different file endings being used. `Glib.Module.open()` actually doesn't need a file ending, as it will try out all appropriate file endings (`.so` and `.dylib` on macOS) - we only need this to automatically discover and load all plugins in the respective folder, which we shouldn't do like this anyway (as it can cause compatibility issues).

Thanks anyway for opening this PR, it's good to have this at hand to point people to that have this issue.

PR #1797, opened 2025-11-22
po: complete the Czech translation

balejk:
Most of these should be accurate but for a few (such as "About" or
"Topic") a more relaxed translation was used to better fit the
context(s) or feel more natural. The "Send Typing Notifications"
translation is admittedly a bit convoluted.

I was not sure about the meaning of the underscore in for instance "\_Notifications" but I saw that the other languages keep it so I also kept it, although perhaps arbitrarily placed.

Also, it seems that it is not possible to translate the OMEMO section of Preferences.


fiaxh:
Thanks for working on translations.

We usually manage translations on [Weblate](https://hosted.weblate.org/projects/dino/).

I merged your changes alongside an update from Weblate in 754dfe9ea74567791d242b3cbb16f67e26998d91.

PR #1826, opened 2026-02-15
RFC: Add autostart and background support via libportal

rmader:
A first quick PoC implementing autostart support.

What works:
 - [x] autostart in background using `dino --gapplication-service`
 - [x] autostart with Flatpak
 - [x] show app in Gnome-Shell background apps (apparently only works for flatpaks atm :/ )
 - [x] opening the app from a notification if it hasn't got opened before (runs into an assertion) - https://github.com/dino/dino/pull/1832 included here
 - [ ] Control `window.hide_on_close` via setting

---

Some impressions to show why this is IMO desirable:

1. Easy to enable compared to existing tweaks, especially on phones:

dino-autostart-1-new-2

2. Messages and calls can be received without having the app window open:

dino-autostart-2-new-2

dino-autostart-4-new

3. Works just as well on desktops:

dino-autostart-3

4. Reports as background app (currently only when run as flatpak:
image

---
Related:
 - https://github.com/dino/dino/pull/1828

Closes https://github.com/dino/dino/issues/1476
Closes https://github.com/dino/dino/pull/1826

PR #1828, opened 2026-02-19
Systray Support (Linux)

### LINKED ISSUE #98: Systray support

kousu:
A StatusNotifierItem (SNI) in the system tray that Dino can background itself to.

Features implemented:

- cross-DE support
  - [x] GNOME
  - [x] KDE
  - [x] niri
  - [x] cinnamon
  - [x] xfce
  - [x] sway
  - [x] i3
      - yes, but needs xembed 
- [x] ~~start_minimized option~~
- [x] ~~Remember last minimization state~~
- [ ] ~~-autostart~~ -> to do later; see [discussion](https://github.com/dino/dino/pull/1826)
- [x] unread message icon
- [x] light/dark icons
    - used on KDE
- [x] unread message count tooltip
- [x] Quit option on the tray icon
  - Ctrl-Q still quits dino reliably

This is inspired by and uses code from https://github.com/dino/dino/pull/1209, which fell behind mainline but had some good work.

Fixes https://github.com/dino/dino/issues/98.

## Dev Build

```
git clone -b tray-icon https://github.com/kousu/dino
cd dino
meson setup build
meson compile -C build
G_MESSAGES_DEBUG="dino libdino" ./build/main/dino
```

This needs [the dependencies](https://github.com/dino/dino/wiki/Build#dependencies); if you're running Dino already you might just have them installed but also, especially if you're on an apt system, you probably need to `apt install whatever-dev whatever2-dev ...` until you have everything. Sorry for being ambiguous here, I don't have an apt system close at hand.

## Addressing Past Discussions

@mar-v-in [via](https://github.com/dino/dino/issues/98#issuecomment-519906269):

> So I wonder, what actually is the desired functionality here. Do we need a "systray" icon when closing the window will only minimize and not close it? If closing the window doesn't close it anymore, but minimizes, how do I close it?

Ctrl-Q works, the Quit option on the menu works. If the systray crashes for some reason the main window comes back and then closing the window is quitting.

@1223421 [via](https://github.com/dino/dino/pull/1209#issuecomment-1305211070):
> i continue testing, and another wish: keep the position/sizes of the window when it is hidden/showed if it is possible

I was wondering myself about this. I never thought that deeply about it. After some experimentation I found Under GNOME, KDE, XFCE and Cinnamon the _size_ of a hidden window is maintained but not the position, the window always reappears somewhere the DE wants, usually in the middle but sometimes not but never where you left it, and under i3 and sway the window it always reappears tiled even if it was floating before, so it loses its position, and under GNOME it seems showing the window always maximizes.  Dino shows the same behaviour as Gajim and Signal-Desktop under all these DEs.

@mar-v-in [via](https://github.com/dino/dino/issues/98#issuecomment-519906269):

> Windows 10 [...] should use Toasts ([docs](https://docs.microsoft.com/en-us/windows/uwp/design/shell/tiles-and-notifications/adaptive-interactive-toasts)) for notifications and if running in the background should be mostly invisible.
> macOS [...] Applications can be started without having a window and the fact an app is running is indicated in the Dock. The dock can also display an unread message count and bounce to raise awareness

We can add an entirely separate plugins/tray-icon-windows and plugins/tray-icon-macOS (or dock-icon-macOS) for those, but judging from the comments in #1209 holding up a linux systray because it doesn't also work on two other OSes isn't really a blocker anymore.

# Demos

> [!WARNING]
> These screenshots are slightly out of date: the tooltip now says "Messages waiting" not "{n} unread messages".

## GNOME

image

image

- Doesn't support the tooltip
- Requires [this extension](https://extensions.gnome.org/extension/615/appindicator-support/) until we figure something else out
- Watch out: activation requires double click

## KDE

KDE prefers "symbolic" icons in its tray. The attention icon is subtle, it's the little dot in the top right. Is it too subtle?

image

image

image

or, in Dark mode:

image

image

image

## niri

Screenshot from 2026-02-28 13-25-12
Screenshot from 2026-02-28 13-25-22
.
Screenshot from 2026-02-28 13-35-24

- requires waybar
- doesn't support the tooltip

## Cinnamon

image

image

.

image

image

Dark mode for good measure:

image

Unread messages also trigger Cinnamon to add its own red dot on the task bar!  (But maybe this is just from the notifications.)

image

## XFCE

image

image

.

image

image

## sway

image

image

- Doesn't support the tooltip
- Doesn't support the menu

## i3

image

image

.

image

2026-02-28-134115_1366x768_scrot

- requires `snixembed`

# Related

- https://github.com/dino/dino/pull/1826


Blueshawk:
Thank you for moving the puck on this glacial problem! Also for the heads up to the old threads. I look forward to testing it in peppermint/xfce soon.  :cowboy_hat_face:

kousu:
> Thank you for moving the puck on this glacial problem! Also for the heads up to the old threads. I look forward to testing it in peppermint/xfce soon. 🤠

I'm looking forward to hearing if it works for you @Blueshawk !

kousu:
Wild! Another solution was posted almost simultaneously: https://github.com/dino/dino/pull/1826. It uses libportal and ~~might only~~ works best on GNOME.  I am very down to merge our efforts.
 
---

It also adds an "Autostart" option which I skipped mostly just because it was extra work but also because keeping Dino's UI tight is important to me.  To implement autostart, all it has to do is add a ~/.config/autostart/dino.desktop file with `exec=dino --gapplication-service` (I'm not sure what gapplication-service does but dino already supports it).

On the other hand, we could keep the UI very tight by making autostarting _nonoptional_: if we detect systray support, automatically enable autostarting (and maybe automatically remove it if we detect no systray support?); or what about this concept: install ~/.config/autostart/dino.desktop always, but remove it if the user explicitly quits (either from the tray or from the menu); but quitting due to a signal does not remove it; that you turn on dino and it stays on across reboots, like how chat works on phones, until you choose to turn it off.  I have to say, my instincts react badly to this, but I think that's from years of being trained in being a CLI cavedweller, and it's actually a good idea for most people.

CrudeWax:
Minor question, does the monochrome white/black icon work on other window managers (i.e cinnamon) besides KDE?

kousu:
> Minor question, does the monochrome white/black icon work on other window managers (i.e cinnamon) besides KDE?

Oh I missed this; I answered [here](https://github.com/dino/dino/pull/1828#discussion_r2842391576) sort of.

No it doesn't. The monochrome icon is a choice that KDE makes. Notice e.g. that Transmission uses their colour icon on other environments.

I think I don't want to try to second-guess at people's icon themes in Dino. There's the [icon theming system](https://specifications.freedesktop.org/icon-theme/latest/) which you can use to adjust app icons ([e.g. here's arch's instructions](https://wiki.archlinux.org/title/Icons#Icon_themes)).

I've included white/black copies so icon themers can include Dino in their work, and so you can create an override on your box. Here's how you could use it, @CrudeWax:

```
mkdir -p ~/.local/share/icons/hicolor/scalable/apps/
cp main/data/icons/scalable/apps/im.dino.Dino-symbolic-dark.svg ~/.local/share/icons/hicolor/scalable/apps/im.dino.Dino.svg
```

This is the result in Cinnamon in dark/light modes:

Screenshot From 2026-02-24 01-25-34

image

> [!CAUTION] Watch out for typos! The naming is super picky.

---

Also, I need to document somewhere (here is not ideal but it will have to do) that to _test_ the systray icons you need to futz with your system. Either you copy them manually to /usr/share/icons/hicolor/scalable/apps which is risky, or set up a symlink in `~/.local/share/icons`:
 
```
mkdir -p ~/.local/share/icons/hicolor
ln -sf $(pwd)/main/data/icons/scalable ~/.local/share/icons/hicolor/scalable
```

In either case you also need to restart your system tray (which might mean logging out and back in) to see the change.

main/data/icons/scalable/apps/im.dino.Dino-symbolic.svg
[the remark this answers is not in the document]
kousu: icons in 98444274cdb012c1da9fd81f79233064b62e5eaf  :)

libdino/src/entity/settings.vala
```diff
@@ -80,6 +81,18 @@ public class Settings : Object {
...
+        get { return minimized_; }
+        set {
+            db.settings.upsert()
+                .value(db.settings.key, "minimized", true)
+                .value(db.settings.value, value.to_string())
+                .perform();
+            minimized_ = value;
+        }
+    }
+
```
kousu: This is not a user-visible setting, rather it stores the last state of `main_window.visible` (actually, it's opposite).  Maybe it should not, therefore, be in `settings`. But if not there then where?

I see libdino/src/entity stores things to the database -- including app.settings -- but nothing in there. I don't want to add a new "hidden state" Entity, so for now this will live as a Setting even though it breaks the pattern of other settings.

plugins/tray-icon/src/plugin.vala
```diff
@@ -0,0 +1,360 @@
...
+        main_window = window;
+        debug("main window registered");
+
+        if (minimizable()) {
+          main_window.hide_on_close = true;
+        }
+
+        // When window visibility changes, the SNI tray needs to switch between saying Show/Hide
+        main_window.notify["visible"].connect(() => {
+          debug("notify[visible] fired: visible=%s", main_window.visible.to_string());
```
kousu: Could probably drop the debug()s

plugins/tray-icon/src/plugin.vala
```diff
@@ -0,0 +1,360 @@
...
+      app.stream_interactor.get_module(ChatInteraction.IDENTITY).focused_in.connect((_) => {
+        // on user clicking to a conversation (and hence 'reading' its unread messages).
+        // The update is deferred the Idle loop to ensure the decrement has actually happened.
+        GLib.Idle.add(() => {
+          update_notice();
+          return GLib.Source.REMOVE;
+        });
+      });
+
+      if(yield dbus_service_available("org.freedesktop.Notifications")) {
```
kousu: Checking eagerly here prevents a laggy timeout on non-dbus systems.

kousu:
@mar-v-in, my understanding is that you object to the persistent notification, and [you'd prefer](https://github.com/dino/dino/pull/1826#issuecomment-3947828294) to tie "start minimized"  to "autostart".

I am very interested in not letting this grind to a halt again. In the interest of perfect not being the enemy of good, if I dropped `start_minimized` and the entire GNOME special-case would you be open, in principle at least, to taking this sooner than later?  GNOME users can use [this](https://extensions.gnome.org/extension/615/appindicator-support/) until we figure out a better option.

plugins/tray-icon/src/plugin.vala
kousu: This one was dropped because I'm [not using persistent notifications anymore](https://github.com/dino/dino/pull/1826#issuecomment-3947828294). Though the pattern _is_ used in `has_tray()`:

```vala
      var has_sni = yield dbus_service_available("org.kde.StatusNotifierWatcher");
```

kousu:
I've:

- [x] removed the GNOME notification
- [x] added an attention icon
- [x] made it tolerant of tray crashes
 
You can see how it looks on GNOME now and how the attention icon looks in the updated screenshots above.

The improvements left to my mind are:

- [ ] Debouncing updates; especially when connecting triggering one `update_tray()` per message synced is silly
- [ ] Shaving down status_notifier_item.vala. There's a lot of code in there with heritage [going back to here](https://gist.github.com/mar-v-in/a7642e15dee51f4def03da5aeb4eb5b0) and we aren't using all of it and it should be pared down; especially all the recursive ID chasing stuff in there is messy and I think bug prone.

Also I would want to **squash** the commits before this goes in but I don't think it's quite there yet.

kousu:
> > Minor question, does the monochrome white/black icon work on other window managers (i.e cinnamon) besides KDE?
> 
> Oh I missed this; I answered [here](https://github.com/dino/dino/pull/1828#discussion_r2842391576) sort of.

@CrudeWax I've installed the monochrome icons directly to Breeze and Breeze Dark to make this easier; so now if you install [`breeze-icons` on your system](https://repology.org/project/breeze-icons/versions) and then choose one of them using your Appearance Settings you should get the monochrome dino icon in your system tray. Hope that satisfies!

plugins/tray-icon/src/plugin.vala
```diff
@@ -0,0 +1,255 @@
...
+    public void registered(Dino.Application app) {
+      this.app = app;
+
+      setup_tray();
+
+      app.stream_interactor.get_module(ContentItemStore.IDENTITY).new_item.connect((_x, _y) => {
+        // on new message
+        update_tray();
+      });
+      app.stream_interactor.get_module(ChatInteraction.IDENTITY).focused_in.connect((_) => {
```
dos1: I don't think that's the right signal to listen to, you'll miss the changes made by carbons. Check out #1835.

Since there are multiple potential consumers of this information, it may make sense to put unread message counting to ChatInteraction and make it a property there that can be listened to changes.

mar-v-in: We don't have a total "unread messages" counter at all yet because it also would be very unclear what to count. Count unread messages in groupchats even if you wouldn't get notified on them? Count only those that cause a notification, meaning it's actually not "unread" by the meaning of that word? As you may have noticed, the number that's displayed next to a chat in the sidebar in Dino is just all unread messages, but it changes color if any of those messages are "worth reading" (= had triggered a notification).

I personally would suggest to completely keep out the unread message count from the tooltip: If there's a message unread that would trigger a notification, then there's probably still a notification pending in the notification area - Dino already clears them once the message was read. So in effect the number you would show would either exactly match the number of pending notifications or it is the significantly larger and less useful number of messages.

Instead we could just have an indicator for the mere fact that there is new notifications since the window was last focused, without the exact number. The number really is also not something that has great value for endusers, what difference is it to you whether it says "7 unread messages" or "12 unread messages"?

kousu: >  very unclear what to count.

I agree it's ambiguous and the phrasing of the tooltip is unhelpful. I realized that last night while trying to make the attention icon sync up with my intuitive understanding of whether I had messages that I needed to look at. The tooltip text could become "" vs "Unread messages" corresponding to the plain or the attention icon, without a count.

> Count only those that cause a notification, meaning it's actually not "unread" by the meaning of that word?

There is a clear convention for this: a notification doesn't count as "read", it's just a peek. K9 Mail and Conversations and Signal and WhatsApp and Dino itself all provide "Mark Read" shortcuts on their notifications.


> it changes color if any of those messages are "worth reading" (= had triggered a notification).

I hadn't noticed this no! It's one of those subtle QoL touches that make me love Dino so much :purple_heart: 

By colour you mean the greyscale tone? It's bright for notified and dark grey for non-notified?

image

image

Syncing the tray's attention icon `any(c.colour_scheme is light for c in conversations)` is the most logical to me. 

I see it's calculated here:

https://github.com/dino/dino/blob/6c397f39b8fad1cb903331d1ce444eb0569797ba/main/src/ui/conversation_selector/conversation_selector_row.vala#L247-L253

though not cached anywhere, so @dos1's suggestion of moving that calculation to ChatInteraction rings true to me. I'm game to hook it up, though that seems like feature creep for this PR. For a first pass, I could just copy-paste the calculation as

```
unread != 0 && conversation.get_notification_setting(stream_interactor) == Conversation.NotifySetting.ON
```

but that calculation ignores mentions! In fact I discovered [a bug](https://github.com/dino/dino/issues/1836) while poking at this.

fiaxh:
Thanks for your PR. Some initial thoughts:
- I tested the changes on GNOME and it produced a partially broken main window (no content on the right side, wrong window size), but I assume that's still WIP
- The code can go directly into libdino, there is no need for a plugin
- Counting of unread messages is not implemented efficiently in Dino right now. However, the way this PR counts messages would make things much worse: For every new message, the unread messages of _all_ conversations are recounted, which includes a database query for every conversation. The counting of unread messages can be optimized in the future, but I would like to leave the unread message count out of this PR, in order to keep the scope smaller and finally get tray icons into Dino.
- To get the information whether there are unread messages (not the count, just true/false) you can listen to NotificationEvents.notify_content_item to set the state to "unread messages exist", and to ChatInteraction.focused_in to set the state to "no unread messages". It's not perfect but would serve as a simple urgency hint, and can be improved along with the unread message count in the future.
- When you don't need the arguments when connecting to a signal, you can just leave them out e.g. `tray_item.activate.connect((_x, _y) => ...` can be `tray_item.activate.connect(() => ...`

kousu:
Thank you @fiaxh for your time :pray: 

like I said a bunch of times I _very much_ want this to finally happen. I think it would make a lot of people who have so far passed dino over for years reconsider it. I've had 5 people personally tell me they skipped Dino once they saw it didn't have a tray icon, and I'm looking forward to corralling them back :)

> Thanks for your PR. Some initial thoughts:
> 
>     * I tested the changes on GNOME and it produced a partially broken main window (no content on the right side, wrong window size), but I assume that's still WIP

Oof, any screenshots or other details? /etc/os-release? I'm using my build right now as my daily driver on GNOME, I did `meson install` and it's been very nice for me. The only interaction it has with the main window (I believe??) is via .activate() and .visible(false). It doesn't touch the layout at all. I've also had at least one person from xmpp:chat@dino.im?join use it successfully on their own.

I'm disappointed. I thought it was ready ( when my code is WIP I [mark it draft](https://github.com/dino/dino/pull/1828#event-23126186507) ). But I'm sure we can solve it if we put our heads together.

>     * The code can go directly into libdino, there is no need for a plugin

:+1: 

>     * Counting of unread messages is not implemented efficiently in Dino right now. However, the way this PR counts messages would make things much worse: For every new message, the unread messages of _all_ conversations are recounted, which includes a database query for every conversation. The counting of unread messages can be optimized in the future, but I would like to leave the unread message count out of this PR, in order to keep the scope smaller and finally get tray icons into Dino.
> 
>     * To get the information whether there are unread messages (not the count, just true/false) you can listen to NotificationEvents.notify_content_item to set the state to "unread messages exist", and to ChatInteraction.focused_in to set the state to "no unread messages". It's not perfect but would serve as a simple urgency hint, and can be improved along with the unread message count in the future.

I agree! @mar-v-in already told me to drop the count the tooltip. @dos1's [idea](https://github.com/dino/dino/pull/1828#discussion_r2869233339) of caching unread messages per `ChatInteraction` sounded good to me.  I'll do it your way here and plan to follow up in another PR adding dos1's cache and use it to fix https://github.com/dino/dino/issues/1836 at the same time, so that the tray attention hint matches the attention hints in the conversation list and both respect `NotifySetting.HIGHLIGHT`.

> 
>     * When you don't need the arguments when connecting to a signal, you can just leave them out e.g. `tray_item.activate.connect((_x, _y) => ...` can be `tray_item.activate.connect(() => ...`

Ah cool noted :+1:

fiaxh:
> Oof, any screenshots or other details? /etc/os-release? I'm using my build right now as my daily driver on GNOME, I did meson install and it's been very nice for me. The only interaction it has with the main window (I believe??) is via .activate() and .visible(false). It doesn't touch the layout at all. I've also had at least one person from xmpp:chat@dino.im?join use it successfully on their own.

This is how my window would normally look:
Screenshot From 2026-03-06 21-18-09
And this is with this PR:
Screenshot From 2026-03-06 21-18-31
The conversation content on the right side is missing and the window size is wrongly set.

When I comment out [this line](https://github.com/dino/dino/pull/1828/changes#diff-371ab6acd7e2b46534747f696a287c2d49b483b7dfb758224626dddfa04af783R212), the window loads normally. I checked and it seems that `realize()` is not invoked on the main window anymore.

kousu:
> When I comment out [this line](https://github.com/dino/dino/pull/1828/changes#diff-371ab6acd7e2b46534747f696a287c2d49b483b7dfb758224626dddfa04af783R212), the window loads normally. I checked and it seems that `realize()` is not invoked on the main window anymore.

I do see this, I just didn't clock it as a problem since I always open conversations consciously :purple_circle: 

I have no idea how you found the problem so fast, but I guess it is your app and thank you :). In this case we solved it by you putting your head into it and I just watched :sweat_smile: . That line was a holdover from when I had it auto-hiding and it's not needed anymore, so it can go.

I've done the other things you asked, @fiaxh:

- [x] open chat renders on boot
- [x] no more plugin
  - [x] simplified init since I can better guarantee when the init happens
- [x] no more `get_unread_count()`

kousu:
Sorry, rebased on master and now gitub has buried the previous thread. Please forgiiive 🫠

kousu:
Salutations @mar-v-in, @fiaxh. Do you think you will have any time in say the next two weeks to look over the tray icon again?  It's not perfect but it's good enough.

kousu:
Found [and squashed](https://github.com/dino/dino/commit/584781545ef8bd85a9fedb734860f3cad0f0292d) a rare concurrency bug that led to this happening :P 

Screenshot From 2026-03-25 19-53-04

main/meson.build
```diff
@@ -143,8 +144,15 @@ if meson.get_compiler('vala').version().version_compare('=0.56.12')
...
+install_data('data/icons/scalable/apps/im.dino.Dino-symbolic.svg', install_dir: get_option('datadir') / 'icons/hicolor/symbolic/apps')
+install_data('data/icons/scalable/apps/im.dino.Dino-attention-symbolic.svg', install_dir: get_option('datadir') / 'icons/hicolor/symbolic/apps')
+# these icons are actuallty 128, but there's no apps/128 folder in KDE's themes; apps/48 is the closest.
+install_data('data/icons/scalable/apps/im.dino.Dino-symbolic-light.svg', install_dir: get_option('datadir') / 'icons/breeze/apps/48', rename: 'im.dino.Dino-symbolic.svg')
+install_data('data/icons/scalable/apps/im.dino.Dino-symbolic-dark.svg', install_dir: get_option('datadir') / 'icons/breeze-dark/apps/48', rename: 'im.dino.Dino-symbolic.svg')
+install_data('data/icons/scalable/apps/im.dino.Dino-attention-symbolic-light.svg', install_dir: get_option('datadir') / 'icons/breeze/apps/48', rename: 'im.dino.Dino-attention-symbolic.svg')
+install_data('data/icons/scalable/apps/im.dino.Dino-attention-symbolic-dark.svg', install_dir: get_option('datadir') / 'icons/breeze-dark/apps/48', rename: 'im.dino.Dino-attention-symbolic.svg')
```
mar-v-in: Don't install icons into any themes other than `hicolor`. It's by convention that apps can install icons into the `hicolor` theme, which acts as the fallback/inherit theme for all other themes (including breeze). The other icon theme folders are exclusive to the respective icon themes, apps shouldn't do anything with them.

main/src/ui/tray_freedesktop.vala
```diff
@@ -0,0 +1,229 @@
...
+      };
+
+      // a click on the icon
+      tray_item.activate.connect(() => {
+        if(attention) {
+          // hiding tricks most window managers
+          // into focusing the presented window
+          // but also generally loses the workspace it was on
+          app.window.hide();
+          app.window.present();
```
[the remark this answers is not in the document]
mar-v-in: To do focus stealing on wayland you need an xdg activation token. KDE has updated the StatusNotifierItem interface to include an additional method `ProvideXdgActivationToken(token)` which is invoked before the call to `Activate(x,y)`. If you take the provided token and pass it to `set_startup_id(token)` on the window, the next call to `present()` will be allowed to steal focus.

main/meson.build
kousu: Right. So how do VLC and Transmission have nice light/dark mode icons?

The answer is that someone in KDE packaged and maybe even made them.

<details><summary>Investigation</summary>

```
$ find /usr/share/icons/breeze* | grep -E -i '(transmission|vlc|TeamViewer|wine)-symbolic\.svg' 
/usr/share/icons/breeze/apps/24/vlc-symbolic.svg
/usr/share/icons/breeze/apps/24/TeamViewer-symbolic.svg
/usr/share/icons/breeze/apps/24/wine-symbolic.svg
/usr/share/icons/breeze/apps/22/vlc-symbolic.svg
/usr/share/icons/breeze/apps/22/TeamViewer-symbolic.svg
/usr/share/icons/breeze/apps/22/wine-symbolic.svg
/usr/share/icons/breeze-dark/apps/24/vlc-symbolic.svg
/usr/share/icons/breeze-dark/apps/24/TeamViewer-symbolic.svg
/usr/share/icons/breeze-dark/apps/24/wine-symbolic.svg
/usr/share/icons/breeze-dark/apps/22/vlc-symbolic.svg
/usr/share/icons/breeze-dark/apps/22/TeamViewer-symbolic.svg
/usr/share/icons/breeze-dark/apps/22/wine-symbolic.svg
```

**ALL** of these are part of the icon theme packages, despite being for unrelated, not-even-KDE, apps:

```
$ find /usr/share/icons/breeze* | grep -E -i '(transmission|vlc|TeamViewer|wine)-symbolic\.svg' | pacman -Qqo -
breeze-icons
breeze-icons
breeze-icons
breeze-icons
breeze-icons
breeze-icons
breeze-icons
breeze-icons
breeze-icons
breeze-icons
breeze-icons
breeze-icons
```

</details>

I've removed them from the install script in 57f93546766f316798f99b104dc824ce2a2953e8.

The conventional way to distribute them is to talk to the KDE Breeze maintainers and get them to adopt the icon variants I made => https://invent.kde.org/frameworks/breeze-icons/-/merge_requests/541

Until then, I'm sorry @CrudeWax. The monochrome icons aren't going to be installed by default. If you want them on Cinnamon you're going to have to make your own icon theme or put them in ~/.local/share/icons like [I described](https://github.com/dino/dino/pull/1828#issuecomment-3946563348), that's just the tension icon themes on Linux live with :sob:

main/src/ui/tray_freedesktop.vala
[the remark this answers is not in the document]
kousu: Nice tip! It took me a second to get this sorted but yes, it works under wayland:

## KDE (on wayland)

https://github.com/user-attachments/assets/fe34f842-2fdb-4a81-af51-a90d921e58f7

## KDE (on X11)

X KDE is similar except if the window is currently shown then it blinks the taskbar but doesn't switch to it.

https://github.com/user-attachments/assets/fb7c9e03-2c6c-4a8b-9520-9e9573c6affb


## GNOME (on wayland)

https://github.com/user-attachments/assets/e1cc79c3-4f86-46a1-8080-e62ed40fe505

## sway (on wayland)

sway doesn't listen to focus stealing. It does highlight the activated window/workspace in red, but it did that even without an activation token.

[Screencast_20260412_2140.webm](https://github.com/user-attachments/assets/f5397267-7077-444d-9fa0-2e9bd210a553)

https://github.com/user-attachments/assets/5b807e8a-985a-46bb-8b55-399fefcc456e

## Cinnamon (on X11)

On Cinnamon, focus-stealing sort of half works even though it's X11? If the window is minimized or closed, then present()ing it steals focus:

https://github.com/user-attachments/assets/adc58997-a86e-40c3-8e74-ac665a5faeff

But if it's already shown then present() blinks the taskbar **if** you're on the same workspace, and never jumps focus. If you're on another workspace, **you lose track of the alert** which seems unfortunate :safety_pin: 

https://github.com/user-attachments/assets/27c0dfe5-b2b6-45cf-acec-2ac815be776b


---

(for the record, I'm at d135a2e3d865172e9bea01244836406e42411f73, but I'm going to squash that down)

kousu:
My [responses](https://github.com/dino/dino/pull/1828/changes#r3070501185) got buried for some reason but @mar-v-in I've made the most recent changes you asked for.  How is it looking now?

main/src/ui/tray_freedesktop.vala
```diff
@@ -0,0 +1,244 @@
...
+      // XXX: should these handlers have their handler_ids saved and be properly disconnect()ed in shutdown_tray() ?
+
+      // Sync the attention icon with notifications, and clear it as soon as dino is focused.
+      // This is not great UX; it would be more conventional to sync with
+      // unread messages, but Dino is lacking infrastructure to make that
+      // efficient. This will be improved: https://github.com/dino/dino/pull/1828#issuecomment-4010944395
+      app.stream_interactor.get_module(NotificationEvents.IDENTITY).notify_content_item.connect((content_item, conversation) => {
+        attention = true;
+        attention_target = conversation;
+      });
```
kousu, line 148: I'm saving `attention_target` here so that on Activate the relevant conversation is jumped to.

But looking ahead what @dos1 and I hope to put in with setting the attention flag based on older unread messages from before the current run of Dino, it's not clear what the attention target should be then.

For now I think this works but we'll have to cross that bridge when we come to it.

main/src/ui/tray_freedesktop.vala
```diff
@@ -0,0 +1,244 @@
...
+      this.app = app;
+
+      ulong window_handler = 0;
+      window_handler = app.window_added.connect((window) => {
+        if (!(window is Dino.Ui.MainWindow))return;
+
+        Idle.add(() => {
+          // window_added fires too early; delaying until the idle loop
+          // allows time for app.window = window;
+          setup_tray();
```
kousu, line 233: I'm not 100% sure this comment is still true. It was true when this was a plugin.

nyabinary:
How does it look on the COSMIC DE?

kousu:
> How does it look on the COSMIC DE?

I..don't know! Would you like to try it?

Does this build on your system?

```
git clone -b tray-icon https://github.com/kousu/dino
cd dino
meson setup build
meson compile -C build
G_MESSAGES_DEBUG="dino libdino" ./build/main/dino
```

I'm a bit busy this week so I probably won't get around to figuring out a new DE but it shouldn't be too hard to build so you can see for yourself. Give us some screenshots!

kousu:
@mar-v-in would it be possible to get some of your attention back on this? As far as I know I've addressed all your feedback and it is ready to merge.

kousu:
Hi again @mar-v-in and @fiaxh , is there any interest in this? I'm wondering if it's just fallen off the radar or if there's some deeper concern that is making you hesitant to review or merge.

Hope you're doing well over there either way.

PR #1830, opened 2026-02-21
Fix macOS plugins not loading by building them with .so extension

### LINKED ISSUE #1782: Encryption settings absent on (aarch64) macOS
I built dino with the following:

```sh
git clone git@github.com:dino/dino.git && cd dino

brew install adwaita-icon-theme gpgme icu4c libgpg-error libsignal-protocol-c libgee libsoup@2 cmake vala qrencode libxml2 gspell gst-plugins-base libomemo-c libcanberra gstreamer libadwaita

brew link icu4c --force
export LDFLAGS="-L/opt/homebrew/lib $LDFLAGS"
export CFLAGS="-I/opt/homebrew/opt/libomemo-c/include $CFLAGS"

meson setup build \
     -Dplugin-rtp-h264=enabled \
     -Dplugin-rtp-vaapi=enabled \
     -Dplugin-rtp-vp9=enabled \
     -Dplugin-rtp-msdk=enabled \
     -Dplugin-omemo=enabled

sudo meson install -C build

dino
```

sh4sh:
Fixes #1782, Dino's plugin loader can't find any plugins on MacOS.

It turns out Meson builds them with with `.dylib` file extension, where GLib's `Module.SUFFIX` wants `.so` (even on Mac).

https://github.com/dino/dino/blob/6c397f39b8fad1cb903331d1ce444eb0569797ba/libdino/src/plugin/loader.vala#L41

To get around these upstream issues, we can tell Meson to build our plugins as `.so` to match GLib's expectations.

### DIFF
--- a/plugins/http-files/meson.build
+++ b/plugins/http-files/meson.build
@@ -17,6 +17,6 @@ sources = files(
 vala_args = [
     '-D', 'SOUP_3_0',
 ]
-lib_http_files = shared_library('http-files', sources, name_prefix: '', vala_args: vala_args, dependencies: dependencies, install: true, install_dir: get_option('libdir') / get_option('plugindir'), install_rpath: default_install_rpath)
+lib_http_files = shared_library('http-files', sources, name_prefix: '', name_suffix: 'so', vala_args: vala_args, dependencies: dependencies, install: true, install_dir: get_option('libdir') / get_option('plugindir'), install_rpath: default_install_rpath)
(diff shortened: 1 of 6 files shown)


sh4sh:
..the linked issue seems to have disappeared, however I did find a related PR with a change in the opposite direction (instructing to build plugins as .dylib only): https://github.com/dino/dino/pull/1795

I chose `.so` because my understanding is that GModule insists on that extension. I'm pretty sure they are functionally the same from the MacOS side. Anyway, plugins weren't loading before, but with the change in this PR they now work.

Issue #833, opened 2020-04-30
Notification behaviour in GNOME


schiessle:
I don't know if it is a GNOME issue or a Dino issue. I decided to start here and move it over to the GNOME bug tracker if we find out that it is a GNOME bug.

I'm using Dino with GNOME 3.36. When Dino runs in the background I get a notification for new messages. These notification gets collected in the GNOME notification are which you can open by clicking on the clock. When I click on the notification it just disappears. It would be nice if Dino would start instead and jump to the right conversation.

Is this something Dino could do by providing some kind of "action handler" or is this behavior completely up to GNOME?

mar-v-in:
This is supposed to happen exactly as you describe, however it can only work if you have a dbus service file registered for dino. This happens automatically when doing `make install` (or using any distribution package), but not if you compile from source and run directly from the build folder. You can then copy `./main/data/im.dino.Dino.service` to `~/.local/share/dbus-1/services` and of necessary adjust its `Exec` path.

schiessle:
I use the nighty builds from here https://software.opensuse.org/download.html?project=network:messaging:xmpp:dino&package=dino

The shipped /usr/share/dbus-1/services/im.dino.Dino.service looks ok:

````
[D-BUS Service]
Name=im.dino.Dino
Exec=dino
````

I can also start dino from the terminal by calling "dino". Maybe because I start dino with `--gapplication-service` initially?

mar-v-in:
So Dino is already running when you click the notification, but the main window is not visible? I guess that could also issues (after all the `--gapplication-service` option isn't really tested a lot).

schiessle:
> So Dino is already running when you click the notification, but the main window is not visible? I guess that could also cause issues 

Yes, exactly. Dino runs but the main window is closed when I click on the notification.

esu23:
I have the same problem. 
If I start with `dino`, the main window opens, but once I close dino I stop receiving notifications, obviously. 
If I start with `dino --gapplication-service`, the main window doesn't open but I receive notifications about new messages. Not sure how to make the window visible again.
Tested from Pinephone with Phosh.

rmader:
Should be closed by https://github.com/dino/dino/pull/1832

PR #1832, opened 2026-02-22
FreeDesktopNotifier: Activate app before conversation

rmader:
In order to fix crashes when dino is launched in the background with `--gapplication-service` and gets activated via a message notification before the main window has been opened before. That is because MainWindow is contructed in the app activate callback.

Closes https://github.com/dino/dino/issues/833

---

Preparation for https://github.com/dino/dino/pull/1826, but useful for everyone already using manual autostart launchers with `--gapplication-service`


kousu (approved):
This is an obvious improvement. I hope it goes in 👍

PR #1833, opened 2026-02-22
muc: only do disco on succesfully entering a room

kousu:
Sending a disco#info <iq> on being banned from a room is wasteful, and when _creating_ a room dino asks too soon, receives an error, and gives up on syncing the room metadata giving subtle, hair-tearing inconsistencies because flag.has_room_features() returns false for everything.

It's barely noticeable, because if you're banned from a room you can't query it anyway (though in that case it's still a wasted <iq>), and if you're creating the room all its fields are empty by default.

I noticed this because I'm experimenting with https://modules.prosody.im/mod_muc_defaults; dino incorrectly shows moderation off when it should be on in newly created rooms.

### DIFF
--- a/xmpp-vala/src/module/xep/0045_muc/module.vala
+++ b/xmpp-vala/src/module/xep/0045_muc/module.vala
@@ -113,7 +113,6 @@ public class Module : XmppStreamModule {
 
             stream.get_flag(Flag.IDENTITY).start_muc_enter(bare_jid, presence.id);
 
-            query_room_info.begin(stream, bare_jid);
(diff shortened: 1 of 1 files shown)


kousu:
You can see before this patch in [log.txt](https://github.com/user-attachments/files/25473303/log.txt) how it struggles with several <iq>s that go nowhere; contrast that with with [log-clean.txt](https://github.com/user-attachments/files/25473369/log-clean.txt) which only does one `disco#info` and doesn't provoke errors.

I realize this is a corner of a corner of a case but please consider it; it's a small, easy, correctness fix. :turtle:

fiaxh:
Thanks!

PR #1840, opened 2026-03-08
Add virtual 'Note to Self' roster item

kousu:
Cheogram, Gajim, Signal and WhatsApp all support this, and it is useful.

Screenshot From 2026-03-07 20-01-45

image

I have not updated the translations, I hope that's okay. My understanding is that everything translation, including updating main/po/dino.pot, is handled by Weblate and happens only a few times a year.

## Testing

If you've never built dino before and you want to give this a spin, install the [dependencies](https://github.com/dino/dino/wiki/Build#dependencies) on your system. For Debian that means [these](https://salsa.debian.org/xmpp-team/dino-im/-/blob/debian/master/debian/control?ref_type=heads#L7), Arch are [these](https://gitlab.archlinux.org/archlinux/packaging/packages/dino/-/blob/main/PKGBUILD?ref_type=heads#L11), Fedora is [here](https://src.fedoraproject.org/rpms/dino/blob/rawhide/f/dino.spec#_23).

```
git clone --depth 50 -b note-to-self https://github.com/kousu/dino
cd dino
meson setup build
meson compile -C build
./build/main/dino
```

Let me know if you like it :)

### DIFF
--- a/main/src/ui/add_conversation/roster_list.vala
+++ b/main/src/ui/add_conversation/roster_list.vala
@@ -56,8 +56,22 @@ protected class RosterList {
 
     private void fetch_roster_items(Account account) {
         rows[account] = new HashMap<Jid, ListBoxRow>(Jid.hash_func, Jid.equals_func);
+        bool has_self = false;
         foreach (Roster.Item roster_item in stream_interactor.get_module(RosterManager.IDENTITY).get_roster(account)) {
+            if (roster_item.jid == account.bare_jid) { has_self = true; }
             update_row(account, roster_item.jid);
+        }
+
+        if (!has_self) {
+            // Inject a virtual "Note to Self" contact.
+            // XMPP technically allows people to be on their own contact lists but
+            // in practice it's rarely allowed, in fact there's an embarrassed mention of this in:
+            // https://www.rfc-editor.org/rfc/rfc6121.html#section-2.3.3:
+            // >  Interoperability Note: Some servers return a <not-allowed/> stanza
+            // > error to the client if the value of the <item/> element's 'jid'
+            // > attribute matches the bare JID <localpart@domainpart> of the
+            // > user's account.
+            update_row(account, account.bare_jid);
         }
     }
 
--- a/main/src/ui/conversation_selector/conversation_selector_row.vala
+++ b/main/src/ui/conversation_selector/conversation_selector_row.vala
@@ -39,8 +39,15 @@ public class ConversationSelectorRow : ListBoxRow {
         this.conversation = conversation;
         this.stream_interactor = stream_interactor;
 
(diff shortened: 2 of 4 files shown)


kousu:
Oh whoops, @bodqhrohro has a version of this in https://github.com/dino/dino/pull/1791. Great minds think alike :)

Mine simply does what Signal and Cheogram do, reusing existing UI, so I'm going to keep it open. Sorry, I didn't mean to undermine you... I just dind't know. It was almost quicker to implement than to search the issues :shaking_face:

fiaxh:
Thanks for the PR!

Please also adjust the `get_conversation_display_name` function in `libdino/src/util/display_name.vala`. The idea is to only use the functions in display_name.vala long-term, but not all usages are adjusted yet. That's why with the current PR, the entry in the conversation list correctly shows "Note to Self", but the window title shows the old value.

kousu:
Thank you @fiaxh :) .  I would really like the title to show the same thing. But I just tried that and remembered why I put this in `ui/`: it uses gettext (`_("Note to Self")`) and gettext isn't part of libdino. Should I change the signature to

```
    public static string get_conversation_display_name(StreamInteractor stream_interactor, Conversation conversation, string? muc_pm_format, string? self_format)) {
```

? That is going to cause a bunch of duplication, though I see that is actually what's done with `muc_pm_format` now.


**EDIT** I've implemented it that way in 06516fbb7c43b579ccca2c06d22cec5712e449ea but I marked the fallback string with "[untranslated]" to remind us about this issue

image

kousu:
On the theory that "Note to Self" is a UI concern isolated from libdino, in d9f5bdcc3fbe3656145355c87a666b975be17400 I've simply fixed the window title to match the conversation selector, flowing both through Dino.UI.. That sidesteps the gettext problem.

### Conversation selector

image

### Conversation title

image

## Conversation List

image


---

I've noticed another problem that I missed before ! The Note to Self conversation doesn't jump to the top of the conversations list / the conversations list doesn't scroll to it. I'll try to fix that next :)
**edit:** it was an unrelated UX issue that confused me (I don't use Note to Self often precisely because it's only accessible in Cheogram; I prefer using Signal's Note to Self, but hopefully with this I'll be more encouraged to use my XMPP account ;) ) so it was sorted at the very bottom of my conversations and nothing jumped to it when it was opened; I have a suggestion for helping that in https://github.com/dino/dino/pull/1847.


---

I've rebased on master. I feel good about this patch now. Unless you want to rewrite libdino to understand gettext I think this is the best option.

---

An open quesstion:

* [ ] **Should** the label include the account? Like "Note to Self (user@server.com)" or "Note to Self (account alias)")? If there's only one account that's annoying, but if there's multiple accounts it might be critical.

kousu:
@fiaxh would it be possible to ask for another look at this?  Maybe it's not perfect but it makes Dino more useful for us users right now.

PR #1842, opened 2026-03-12
Fix: Notification handler timeout

christf:
When discussing on the dino muc about high CPU load back in january I was sent this patch to try.
Unfortunately I have no way of finding and attributing the proper author, so if you recognize this, please let me know so we can get this properly attributed.
I have tried the code and it has helped so I am bringing it here for it to not be forgotten...

### DIFF
--- a/libdino/src/dbus/notifications.vala
+++ b/libdino/src/dbus/notifications.vala
@@ -23,10 +23,14 @@ namespace Dino {
         DBusNotifications? ret = null;
         Cancellable cancellable = new Cancellable();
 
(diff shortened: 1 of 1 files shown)


fiaxh:
This patch was written by @mar-v-in and is already on master (3b274a609643632b055630f8e75d3393fbfa46c3). 

However, there seems to be a misunderstanding. To my knowledge, this patch was a fix for a crash and not a CPU optimization. If you observed an improvement in CPU usage it might have been coincidental, I'm afraid.

Issue #85, opened 2017-05-07
Copying multiple messages impossible


NicoHood:
There is no way to select multiple messages and copy+paste them to someone else. It should contain some timestamps and the username/alias.

jubalh:
Sorry but after such a long time I have to: +1

kousu:
I want to +1 this too, especially copying usernames. Right now it is impossible to even copy a single message with its username, which is not really a problem in 1:1 chats, but in groupchats it makes it impossible to follow the log.

selurvedu:
I would be glad to post a small bounty on this issue, but Bountysource seems to be broken. Is there an alternative method?

z-eos:
definitely +1

raingloom:
Can we at least discuss where a new contributor should get started if they wanted to add it? This is such a basic feature it's a bit baffling it hasn't been implemented for almost 3 years.

PR #1872, opened 2026-06-05
Add multi message selection

catap:
Conversation history now supports direct primary button selection across message and file rows; drag selects contiguous spans, Ctrl click toggles individual rows, Shift click extends from the current anchor, and the existing item action menu remains available for non contiguous selection.

Selected content is exported immediately to the primary selection for middle button paste, and Ctrl+C copies the same sender and timestamp annotated payload to the clipboard.

It closes 9 years old https://github.com/dino/dino/issues/85

### DIFF
--- a/main/data/style.css
+++ b/main/data/style.css
@@ -96,12 +96,30 @@ window.dino-main .dino-conversation .message-box.edit-mode:hover {
     background: alpha(@theme_selected_bg_color, 0.12);
 }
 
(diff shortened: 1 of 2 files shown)


catap:
Disclamer: I'm not UI nor Valve programmer, I just use Dino and I really tired from copy and paste messages which requires a lot of manual job.

PR #1873, opened 2026-06-06
Fix chat correction content item lookup

catap:
Direct chat corrections are accepted by bare JID, but the latest correction lookup still constrained the query by the original full resource; a correction sent from another resource therefore produced no matching correction row, returned -1, and wrote that sentinel into the message content item's foreign id.

Align the lookup with the acceptance rule: retain occupant id matching for MUC corrections, retain resource matching for non chat cases where it is still meaningful, and avoid resource filtering for direct chat corrections. Add a database migration which removes existing message content items whose referenced message row no longer exists.

### DIFF
--- a/libdino/src/service/database.vala
+++ b/libdino/src/service/database.vala
@@ -7,7 +7,7 @@ using Dino.Entities;
 namespace Dino {
 
 public class Database : Qlite.Database {
-    private const int VERSION = 33;
+    private const int VERSION = 32;
 
     public class AccountTable : Table {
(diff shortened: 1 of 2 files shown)


fiaxh:
Thanks for catching this!

When a content item can't be loaded, Dino fails to display the whole conversation. I assume that happened in your case as well. I made a small change such that it displays all content items that can be read (cf8e81e26862b463c6d42fc6359557db85eb8d90), while still logging a warning.

catap:
@fiaxh I had catch it by "empty chat" :-)

PR #1874, opened 2026-06-06
Make file downloads recoverable

catap:
Completed HTTP and SFS file transfers may point at a cached file which was only partially fetched, but Dino still treats that cache path as complete. Add a Refetch action to the file, video, and photo menus; it reuses the existing download path, forcing a new GET and replacing the stored cache path when the transfer completes.

Reload image widgets when the file transfer path changes, so photos show the newly fetched cache entry rather than the previous texture. Correct the in progress file menu to call the existing cancel action.

### DIFF
--- a/libdino/src/entity/file_transfer.vala
+++ b/libdino/src/entity/file_transfer.vala
@@ -260,6 +260,12 @@ public class FileTransfer : Object {
         return File.new_for_path(Path.build_filename(Dino.get_storage_dir(), "files", path));
     }
 
+    public File? get_partial_file() {
+        File? file = get_file();
+        if (file == null) return null;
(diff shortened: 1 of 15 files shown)


mar-v-in:
Instead of adding a user interface button for repairing a broken situation, we should just:
1. Make sure to not reach that broken state
2. If it happens nonetheless, automatically detect it and recover from there.

That's probably best done here by downloading the file not to its final destination, but into an in-progress file. That way the file can be easily spotted as not fully downloaded. If the server supports it, the download could then be continued there using a ranged transfer.

catap:
@mar-v-in thanks for review. Here clenaer attempt which doesn't introduce any new UI elemnt.

PR #1876, opened 2026-06-10
Refech history

catap:
Here two commits. The first one fixed anoyed bug, and the second one allows to refetch history without manually edit DB.

1) history-sync: keep incomplete ranges open

Do not treat an exhausted MAM page as proof that an expected range boundary was reached. A completed query with a missing target now leaves the range unmerged; from_end is written only when the query actually starts at the archive beginning

2) history-sync: add manual chat resync

Add a conversation menu action that walks MAM history for the selected chat and stores the scanned messages through the normal receive pipeline. The notification banner reports scanned message counts, restores active resync state when switching back to the chat, and exposes cancellation without live inserting every archived message into the current view.

### DIFF
--- a/libdino/src/service/database.vala
+++ b/libdino/src/service/database.vala
@@ -7,7 +7,7 @@ using Dino.Entities;
 namespace Dino {
 
 public class Database : Qlite.Database {
-    private const int VERSION = 34;
+    private const int VERSION = 35;
 
     public class AccountTable : Table {
         public Column<int> id = new Column.Integer("id") { primary_key = true, auto_increment = true };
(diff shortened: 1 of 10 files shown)


mar-v-in:
Please refrain from doing one pull request with two largely unrelated commits.

On first:
- When would a MAM page be exhausted without the range boundary reached?
- The query may request to fill a hole that is not (fully) available on the server anymore, e.g. because the client was offline for longer than the message retention window. from_end indicates that the current server archive has no older messages, not that no such messages ever existed. Not setting it would result in the client repeatedly trying to fill a hole that can't be filled because the messages aren't stored on the server anymore.

On second:
This shouldn't be a user interface. Bugs should be fixed and the need to repair detected automatically. Requiring the user to detect a broken state and then ask them to press a button (that is not guaranteed to fix it either) is bad UX.

catap:
@mar-v-in I had rebase this branch and reworked the second commit to make recover non interactive. For now I suggest to keep PR with two commits to make review easy because it is about the same things, but I will drop the second commit from this PR and reopen it and new one, if it required.

PR #1877, opened 2026-06-10
rtp: avoid invalid video caps

catap:
Some GStreamer device caps expose ordinary system memory with empty or ANY features, not with an explicit memory:SystemMemory feature; accept those caps while selecting the preferred video format.

If no raw video caps match, return ANY caps after logging a warning instead of copying caps with index -1.


rmader:
Ah nice, thanks, I guess this deserves a
```
Fixes: 6e6d1d16 (Only support SystemMemory caps for video)
```

(https://github.com/dino/dino/pull/1736)

catap:
@rmader pushed with added requested line.

PR #1893, opened 2026-08-11
Muc fixes

mesonium:


mesonium:
I've dropped the last commit as it turned out to be redundant.

moparisthebest:
would be great to merge this and fix these long-standing exploits

